PR #10, opened 2023-09-20
Add slack notification action

bjohnsto:
Create a new github action to notify us on slack every 2 hours about prs that need review, have requested changes, and are approved.


slegendr (commented):
Instead of listing "Number of PRs requesting changes", I think it would be better to say "Number of PRs awaiting changes". That way it is clear that we are waiting on changes to be made, but feedback has already been provided.

slegendr (approved):

PR #12, opened 2023-10-04
Added .gitignore files and removed unused .cvsignore files.

raeburn:
The contents aren't identical because of different semantics, changes in what files are generated, etc.


C2Redhat:
The .gitignore files look good to be but should removing .cvsignore be one commit and adding .gitignore be another?

raeburn:
> The .gitignore files look good to be but should removing .cvsignore be one commit and adding .gitignore be another?

I think it's reasonable to do as one as it's more or less updating the ignore-list functionality from CVS to git. I could make them two separate commits if you think that's clearer.

C2Redhat:
Yes, please.  If it is not too much work.  Thanks!

C2Redhat (approved):
yay

PR #13, opened 2023-10-04
Replace EXTERNAL_STATIC with STATIC.

raeburn:
It still needs defining as a macro or replacing in the text, but using a token of the same length as "static" (which is the version we send upstream) means maintaining indentation (and line breaks) in our internal code consistent with that which we send upstream will be easier.


C2Redhat:
Just one quick nitpick, do we care on lining up the '\' in the Makefile ?

lorelei-sakai:
This change looks reasonable to me.

C2Redhat (approved):
yay

PR #15, opened 2023-10-06
Redefine blkdev_put using VDO_USE_ALTERNATE macro.

C2Redhat:
With Fedora 37 and Fedora 38, 6.5 kernel version and above, blkdev_put definition has changed. We need to use VDO_USE_ALTERNATE to redefine blkdev_put base on the kernel version similar to what we did with blkdev_get_by_path.
This change fixes a bug introduced by perforce change 158874.


bjohnsto (approved):
The change looks correct to me.

lorelei-sakai (approved):

PR #17, opened 2023-10-09
Fix url link

bjohnsto:
Change the space into a %20 for the url link


lorelei-sakai (approved):
Oh good heavenss, I didn't realize thosewere hard-coded links. I assumed you were running it through some kind of URL escape mechanism.

The change seems right, though.

C2Redhat (approved):

PR #20, opened 2023-10-13
don't loop forever trying to set the kernel mem= option

raeburn:
When setting the kernel memory limit, put a bound on the number of
iterations and then give up and clean up.

Tested with LowMemGenData01 on a pfarm.


lorelei-sakai:
Please follow our Github workflow style for the commit message and the PR message.

lorelei-sakai (approved):

PR #21, opened 2023-10-19
[VDO-5591] Make unique name for uploaded logs

bjohnsto:
Logfile package will now be called logs-[repo name]-[pr#]-[last commit sha]


lorelei-sakai:
The content of the change looks reasonable. This is related to VDO-5591, though, right? It would be good to mention the ticket somewhere. Probably in the PR description is more useful than in the base commit message.

PR #25, opened 2023-10-27
[VDO-5599] Adapt memory allocation name changes from upstream

lorelei-sakai:
Adapt Mike's upstream patch series renaming memory allocation functions so they don't look like macros. Also made a few minor edits for consistency. In particular change 12 (of 13) has been trimmed down to remove unrelated style changes.

The new patches in this series extend the renaming to the test and userspace code.


src/c++/vdo/base/block-map.c
slegendr, line 1: Should the UDS_FREE state name on line 281 be removed? I do not see it used anywhere else.

lorelei-sakai: The state name probably makes more sense as simply "FREE" rather than UDS_FREE". At any rate, Mike's original change didn't touch it so I left it alone.

It looks like that string, and the method it's in, are only used for logging one one error path, so I kind of wonder if we can cut it entirely. But separately from this patchset.

slegendr: Okay, that makes sense.

slegendr (approved):

PR #27, opened 2023-10-31
[VDO-5602] Feature/two line block brace

lorelei-sakai:
Fix various formatting issues in code flow. This doesn't necessarily include all instances that need to be fixed, but it should cover the vast majority of them.

Fixed issues:
Change unary increment from ++foo to foo++.
Add braces around code blocks that span more than one line, even if they contain a single statement.
Add more blank lines for readability.
Remove unnecessary void pointer casts.


bjohnsto (approved):
Looks good to me

slegendr (approved):

PR #28, opened 2023-11-02
Add script to create a kernel overlay branch from commit(s) or local branch.

slegendr:
Add commit_to_overlay.sh, which creates a new branch on the specified kernel tree repo with input changes overlayed. The inputs can be a commit, series of commits, or local branch name.  Full or partial commit SHAs can be used.

### DIFF
--- a/src/packaging/kpatch/commit_to_overlay.sh
+++ b/src/packaging/kpatch/commit_to_overlay.sh
@@ -1,0 +1,471 @@
+#!/bin/bash
+# commit_to_overlay.sh
+# Create a kernel tree overlay branch from a commit, series of commits, or local branch.
+#
+# This script is intended to be run from a developer fork of dm-vdo/vdo-devel after the relevant
+# PR(s) have been merged.
+#
+# Usage:
+# ./commit_to_overlay.sh <kernel_tree_repo_URL> <kernel_tree_branch_name>
+#                        <overlay_branch_name> [ -c | -b ] args...
+#
+# Input Type Options:
+#  -c  Process a commit or series of commits into a kernel overlay branch
+#  -b  Process all changes on a local branch into a kernel overlay branch
+#
+# Accepted Arguments:
+#  Commit SHA(s) ordered from oldest to newest
+#  Local branch name
+#
+# ** Multiple arguments must be delimited with a space **
+#
+# Notes:
+# - It is recommended that the SSH kernel_tree_repo_URL is input, as GitHub prompts for a
+#   username and password when an HTTPS URL is used.
+# - The local branch option (-b) relies on detecting differences between the input branch and
+#   origin/HEAD. It is only effective before a rebase of the fork is performed.
+# - Both the remote kernel tree and local vdo source tree will be cloned into a local absolute path
+#   during this process. By default, they will both be removed upon exiting the script, unless the
+#   DEBUG flag is set.
+# - Regardless of the DEBUG flag setting, the kernel tree clone will be retained if the user elects
+#   (when prompted) to manually push the successfully created overlay branch at a later time.
+# - Output from building the VDO tree while processing each change can be found in the
+#   /tmp/overlay_build_log* file. The log is removed at the end of each loop iteration upon success.
+#   It is retained until the commit_to_overlay.sh script is re-run upon failure, or when the DEBUG
+#   flag is set.
+#
+# Potential Future Modifications:
+# - Remove the kernel_tree_branch_name input argument if it is always vdo-next.
+# - Modify the kernel_tree_repo_URL input argument to accept a path to an existing clone instead of
+#   a URL. When a path is entered, bypass cloning the repo and use the repo at the path location.
+##
+
+if [[ -z ${DEBUG} ]]; then
+  DEBUG=1
+fi
+
+TOOL=$(basename $0)
+RUN_DIR=$(pwd)
+SOURCE_REPO_URL=$(git config --get remote.origin.url)
+MANUAL_PUSH=1
+COLOR_RED='\033[0;31m'
+NO_COLOR='\033[0m'
+
+ADDITIONAL_ARGS=()  # Array of command line arguments
+COMMIT_SHAS=()      # Array of commit SHAs to be processed
+CONFIG_MSG=()       # Array of strings to be used in the configuration message output to stdout
+INPUT_TYPE=         # Indicated input type
+OVERLAY_BRANCH=     # Name to give the kernel overlay branch being created
+KERNEL_REPO_URL=    # URL for the kernel tree repository where the overlay branch will be created
+KERNEL_BRANCH=      # Kernel tree branch to base the overlay on
+PROCESS_INPUT_FX=   # Function to call to process the input(s)
+SOURCE_BRANCH=      # Branch on the source repo where the relevant commits were made
+
+# Expand the arguments provided into the necessary parameters to operate on.
+process_args() {
+  if [ -z "$1" ] || [ -z "$2" ] || [ -z "$3" ] || [ -z "$4" ] || [ -z "$5" ]; then
+    printUsage
+  fi
+
+  KERNEL_REPO_URL=$1
+  KERNEL_BRANCH=$2
+  OVERLAY_BRANCH=$3
+  INPUT_TYPE=$4
+  shift 4
+  ADDITIONAL_ARGS=("$@")
+
+  # Remove any artifacts from previous runs
+  rm -f /tmp/overlay_build_log*
+
+  # Set parameters based on input type
+  case "${INPUT_TYPE}" in
+    "-c"|"--c"|"c")
+      PROCESS_INPUT_FX=process_commits
+      CONFIG_MSG=("commit" "Commit(s) to be used in overlay:")
+      ;;
+    "-b"|"--b"|"b")
+      PROCESS_INPUT_FX=process_branch
+      CONFIG_MSG=("branch" "Local branch to be used in overlay:")
+      SOURCE_BRANCH=${ADDITIONAL_ARGS[0]}
+
+      # Verify input branch exists
+      git ls-remote --exit-code --heads --refs ${SOURCE_REPO_URL} ${SOURCE_BRANCH} &>/dev/null
+      if [ $? == 2 ]; then
+        echo -e "${COLOR_RED}ERROR: Branch ${SOURCE_BRANCH} not found on repo" \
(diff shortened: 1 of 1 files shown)


src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -0,0 +1,459 @@
...
+
+  unset -f GITDATE GITAUTHOR CHANGE_LOG
+done
+
+# Push the kernel overlay branch to the remote kernel repository
+cd ${LINUX_SRC}
+prompt_user_push
+
+if [[ ${MANUAL_PUSH} != 0 ]]; then
+  git push -u origin ${OVERLAY_BRANCH}
```
lorelei-sakai, line 456: I feel like the MANUAL_PUSH flag is set backwards. Wouldn't MANUAL_PUSH =1 mean leave the push to the user, and MANUAL_PUSH=0 mean do the push automatically?

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -0,0 +1,459 @@
...
+  fi
+
+  # Check out the new version to build and post - do we need to do this - it puts us in a detached HEAD state, so commits will not be retained
+  #sudo git config --system advice.detachedHead "false"
+  echo "Checking out the change to build and post it"
+  git checkout ${change}
+
+  # Capture the commit message from the change
+  echo "Creating the initial commit file using the commit message from the change"
+  commit_file=$(mktemp /tmp/overlay_commit_file.XXXXX)
```
lorelei-sakai, line 376: Would it make more sense to put the logfile in ${RUN_DIR} instead of in /tmp?

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -0,0 +1,459 @@
...
+VDO_TREE=${RUN_DIR}/$(mktemp -d vdo-raw-XXXXXX)
+
+# Shallow clone the specified dm-linux repo branch into LINUX_SRC
+echo "Cloning kernel repo $(get_repo_name ${KERNEL_REPO_URL})/${KERNEL_BRANCH}"
+git clone --branch ${KERNEL_BRANCH} --depth 1 ${KERNEL_REPO_URL} ${LINUX_SRC}
+cd $LINUX_SRC
+git checkout -b ${OVERLAY_BRANCH} ${KERNEL_BRANCH}
+
+# Need to detect that our repo is x number of commits ahead of dm-vdo/vdo-devel
+# Clone the current repo as ${VDO_TREE}? Is VDO_TREE even needed? Aren't we working from the clone when this script is called? Need to make a copy?
```
lorelei-sakai, line 344: I agree we probably don't need this clone but we can figure that out later.

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -0,0 +1,459 @@
...
+  fi
+
+  cd ${VDO_TREE}
+
+  # Clean up from the last build, if necessary
+  if [[ "${change}" != "${COMMIT_SHAS[0]}" ]]; then
+    echo -en "Cleaning up from the last build... "
+    git clean -fdxq
+    echo "Done"
+  fi
```
lorelei-sakai, line 367: This bit seems unnecessary given we just cloned the repo. But it might be useful in the future, so no harm leaving it for now.

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -0,0 +1,459 @@
...
+
+  # Capture authorship information for use later
+  commit_author_email=$(git log --format=%aE -1 ${change})
+  commit_author_name=$(git log --format=%aN -1 ${change})
+  commit_date=$(git log --format=%ad -1 ${change})
+  gitauthor="--author \"${commit_author_name} <${commit_author_email}>\""
+  gitdate="--date \"${commit_date}\""
+
+  # Build the tree as normal
+  build_log_file=$(mktemp /tmp/overlay_build_log.XXXXX)
```
lorelei-sakai, line 390: Again, should the logs go in ${RUN_DIR}?

lorelei-sakai (changes requested):
Mostly looks good. A lot more polish than I was expecting. A few minor questions about documentation and logfile locations, otherwise all fine.

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -0,0 +1,459 @@
+#!/bin/bash
+# commit_to_overlay.sh
+# Create a kernel tree overlay branch from a commit, series of commits, or local branch.
+#
+# This script is intended to be run from a developer fork of dm-vdo/vdo-devel after the relevant
+# PR(s) have been merged.
+#
+# Usage:
+# ./commit_to_overlay.sh <kernel_tree_repo_URL> <kernel_tree_branch_name> [ -c | -b ] args...
```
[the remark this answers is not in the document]
slegendr: I will fix that.

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -0,0 +1,459 @@
...
+  if [ $return == 128 ]; then
+    echo -e "${COLOR_RED}ERROR: Input url (${KERNEL_REPO_URL}) is not a valid repository${NO_COLOR}"
+    print_usage
+  elif [ $return == 2 ]; then
+    echo -e "${COLOR_RED}ERROR: Branch ${KERNEL_BRANCH} not found on repo" \
+            "$(get_repo_name ${KERNEL_REPO_URL})${NO_COLOR}"
+    print_usage
+  fi
+}
+
```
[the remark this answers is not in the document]
slegendr: Good catch. I'm adding a check so it will print the usage info and exit if the branch already exists.

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -0,0 +1,459 @@
...
+# Print the script usage information to stdout
+print_usage() {
+  echo
+  echo "${TOOL}: Create a kernel tree overlay branch from a commit, series of commits," \
+       " or local branch"
+  echo
+  echo "  Usage: "
+  echo "  ./${TOOL} <kernel_tree_repo_URL> <kernel_tree_branch_name> [ -c | -b ] args..."
```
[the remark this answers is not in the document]
slegendr: Fixing.

src/packaging/kpatch/commit_to_overlay.sh
slegendr: For shell scripts, true is 0 and 1 is false. As such, I was keeping with that theme so MANUAL_PUSH true is 0, automatic push is MANUAL_PUSH false, which is 1.

src/packaging/kpatch/commit_to_overlay.sh
slegendr: It could. It is my understanding that systemd-tmpfiles applies an ageing concept to files in /tmp  and /var/tmp so that they are automatically cleaned at regular intervals. As such, there is less harm if the script somehow fails to clean up the commit files or build log.

src/packaging/kpatch/commit_to_overlay.sh
slegendr: The point of this is that it runs in between processing each commit. The clone persists across all commits being processed. However, it may not be necessary since we checkout the change we intend to build. Could be that the tree is already clean due to the detached HEAD state we enter when the change is checked out.

src/packaging/kpatch/commit_to_overlay.sh
slegendr: Potentially. I chose /tmp for the reasons mentioned above. Although I can change that if you think it is better in ${RUN_DIR}.

src/packaging/kpatch/commit_to_overlay.sh
lorelei-sakai: I ... can see the logic. Seems weird to me but I wouldn't want to make the style non-standard in a confusing way.

src/packaging/kpatch/commit_to_overlay.sh
lorelei-sakai: A new checkout won't erase build products since those aren't tracked by git (at least not the ones we've told .gitignore about, which is most of them). So the clean is probably necessary after all.

src/packaging/kpatch/commit_to_overlay.sh
lorelei-sakai: Having the in $RUN_DIR seems like it would make the easier to find. Does the script mention anywhere where the logs will be? That might be a reasonable compromise.

lorelei-sakai: Re-reading the script, the commit_file is temporary and we delete it on cleanup anyway, so /tmp is fine.

src/packaging/kpatch/commit_to_overlay.sh
lorelei-sakai: I see the error message directly names the build_log file, which allays my concern about finding it. So having it live in /tmp is fine.

src/packaging/kpatch/commit_to_overlay.sh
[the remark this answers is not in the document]
slegendr: See commit ec8a389 for the fixes.

src/packaging/kpatch/commit_to_overlay.sh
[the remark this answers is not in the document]
slegendr: See commit ec8a389 for the fixes.

src/packaging/kpatch/commit_to_overlay.sh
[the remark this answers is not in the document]
slegendr: See commit ec8a389 for the fixes.

src/packaging/kpatch/commit_to_overlay.sh
slegendr: True. Probably good to have the option to clone or enter a path to an existing clone, as we had discussed earlier. Maybe an option isn't needed, but that depends on the workflow we decide to follow.

src/packaging/kpatch/commit_to_overlay.sh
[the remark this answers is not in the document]
lorelei-sakai: Looks good.

lorelei-sakai (approved):
All looks good to me now.

bjohnsto (approved):
Change to restoring Makefile looks ok. Approved.

PR #31, opened 2023-11-08
[VDO-5602] Add more bracing and adjust newlines for legibility

lorelei-sakai:
When going upstream, this commit will be merged into change 94d1653b6972 "dm vdo: fix bracing and whitespace nits"

These additional changes are the result of reviewing Ken's large function reflowing commits. They are either changes that were missed earlier, or changes which don't actually require reflowing function calls.


bjohnsto (approved):
They look ok to me

raeburn (approved):

PR #32, opened 2023-11-08
[TECH-151] Improve the kernel overlay script

lorelei-sakai:
Streamline the overlay process by removing some unnecessary steps and some cruft that should not be added to the kernel tree.
Also add the ability to specify a series of commits by pulling them from a merge commit.

### DIFF
--- a/src/packaging/kpatch/commit_to_overlay.sh
+++ b/src/packaging/kpatch/commit_to_overlay.sh
@@ -427,10 +427,12 @@ for change in ${COMMIT_SHAS[@]}; do
   gitauthor="--author \"${commit_author_name} <${commit_author_email}>\""
   gitdate="--date \"${commit_date}\""
 
-  # Build the tree as normal
+  # Build the necessary parts of the tree
   build_log_file=$(mktemp /tmp/overlay_build_log.XXXXX)
-  echo -en "Building the VDO tree... "
-  make >> ${build_log_file} 2>&1
+  echo -en "Building the VDO perl directory... "
+  make -C src/perl >> ${build_log_file} 2>&1
+  echo -en "Generating VDO statistics files... "
+  make -C src/stats >> ${build_log_file} 2>&1
   echo "Done"
 
   if [[ $? != 0 ]]; then
--- a/src/packaging/kpatch/commit_to_overlay.sh
+++ b/src/packaging/kpatch/commit_to_overlay.sh
@@ -327,7 +327,7 @@ _cleanup() {
     rm -fv ${commit_file}
   fi
 
-  unset -f CHANGE_LOG DEBUG GITDATE GITAUTHOR LINUX_SRC
+  unset -f COMMIT_FILE DEBUG GITDATE GITAUTHOR LINUX_SRC
 }
 
 ###############################################################################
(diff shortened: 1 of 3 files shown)


src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -133,11 +139,12 @@ print_usage() {
   echo
   echo "  Usage: "
   echo "  ./${TOOL} <kernel_tree_repo_URL> <kernel_tree_branch_name>"
-  printf "%$((${indent}+3))s%s\n" ' ' '<overlay_branch_name> [ -c | -b ] args...'
+  printf "%$((${indent}+3))s%s\n" ' ' '<overlay_branch_name> [ -c | -b | -m ] args...'
   echo
   echo "  Input Type Options:"
   echo "     -c  Process a commit or series of commits into a kernel overlay branch"
   echo "     -b  Process all changes on a local branch into a kernel overlay branch"
+  echo "     -m  Process all changes merged in the name merge commit"
```
slegendr, line 147: 'name' should be 'named'.

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -417,14 +446,11 @@ for change in ${COMMIT_SHAS[@]}; do
 
   sed -r -i -e 's/^GITARGS=/GITARGS=\nGITAUTHOR ?= ""/g'        \
             -e 's/^GITARGS=/GITARGS=\nGITDATE ?= ""/g'          \
-            -e 's/commit -m/commit --file/g'                     \
-            -e 's/ commit / commit $(GITAUTHOR) $(GITDATE) /g'  \
-            -e 's/^(\s*&&).*format-patch.*/\1 echo committed/g' \
-            -e "s/ origin\/master / origin\/${KERNEL_BRANCH} /g" Makefile
+            -e 's/ commit / commit $(GITAUTHOR) $(GITDATE) /g' Makefile
```
slegendr, line 449: This will replace it globally, but you only need this in the new kernel-kpatch Makefile target. Could you make this more explicit to only catch the kernel-kpatch instance of 'commit' and not the existing kpatch one?

slegendr (changes requested):

src/packaging/kpatch/commit_to_overlay.sh
lorelei-sakai: I think this is fixed.

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -401,7 +430,8 @@ for change in ${COMMIT_SHAS[@]}; do
   # Build the tree as normal
   build_log_file=$(mktemp /tmp/overlay_build_log.XXXXX)
   echo -en "Building the VDO tree... "
```
slegendr, line 432: Since we're no longer building the entire tree, I feel that this message should be modified to indicate which portions are being built.

src/packaging/kpatch/commit_to_overlay.sh
slegendr: Okay, looks good.

src/packaging/kpatch/commit_to_overlay.sh
lorelei-sakai: I've added updated out out for each command, and update the section comment as well.

slegendr (approved):

PR #33, opened 2023-11-08
VDO-5610 fix RAID during emergency reboot

bjohnsto:
Our RAID device exists when the machine comes back up during emergency reboot, so we should not have mdadm --create in the activate method of RAID.pm. We will move that to the setup function, thus only creating once, and will move the stop method to the destroy method. We will remove the activate and stop functionality. This should be fine since RAID is a BOTTOM device and will not have anything underneath it that needs the RAID to be stopped during stop/start lifecycle.


lorelei-sakai (approved):
I think it's good to get in the habit of signing our commits, though for internal things it really doesn't matter.

PR #34, opened 2023-11-09
[VDO-5600] Apply io-submitter changes from upstream

slegendr:
Apply Mike's upstream patch series for renaming io-submitter functions and adjusting formatting. Also made a few edits to spacing for consistency in alignment of parameters. 

The new patch in this series renames submit_vio (previously process_vio_io) to vdo_submit_vio for consistency with the other non-static function names.

### DIFF
--- a/src/c++/vdo/base/io-submitter.c
+++ b/src/c++/vdo/base/io-submitter.c
@@ -323,13 +323,13 @@ static bool try_bio_map_merge(struct vio *vio)
 }
 
 /**
- * submit_data_vio_io() - Submit I/O for a data_vio.
+ * vdo_submit_data_vio() - Submit I/O for a data_vio.
  * @data_vio: the data_vio for which to issue I/O.
  *
  * If possible, this I/O will be merged other pending I/Os. Otherwise, the data_vio will be sent to
  * the appropriate bio zone directly.
  */
-void submit_data_vio_io(struct data_vio *data_vio)
+void vdo_submit_data_vio(struct data_vio *data_vio)
 {
 #ifdef VDO_INTERNAL
 	data_vio->vio.bio_submission_jiffies = jiffies;
--- a/src/c++/vdo/base/io-submitter.c
+++ b/src/c++/vdo/base/io-submitter.c
@@ -133,11 +133,6 @@ static void send_bio_to_device(struct vio *vio, struct bio *bio)
 	submit_bio_noacct(bio);
 }
 
-static sector_t get_bio_sector(struct bio *bio)
-{
-	return bio->bi_iter.bi_sector;
-}
-
 /**
  * process_vio_io() - Submits a vio's bio to the underlying block device. May block if the device
  *                    is busy. This callback should be used by vios which did not attempt to merge.
@@ -167,8 +162,10 @@ static struct bio *get_bio_list(struct vio *vio)
 	assert_in_bio_zone(vio);
 
 	mutex_lock(&bio_queue_data->lock);
-	vdo_int_map_remove(bio_queue_data->map, get_bio_sector(vio->bios_merged.head));
-	vdo_int_map_remove(bio_queue_data->map, get_bio_sector(vio->bios_merged.tail));
+	vdo_int_map_remove(bio_queue_data->map,
+			   vio->bios_merged.head->bi_iter.bi_sector);
+	vdo_int_map_remove(bio_queue_data->map,
+			   vio->bios_merged.tail->bi_iter.bi_sector);
 	bio = vio->bios_merged.head;
 	bio_list_init(&vio->bios_merged);
 	mutex_unlock(&bio_queue_data->lock);
@@ -211,7 +208,7 @@ static struct vio *get_mergeable_locked(struct int_map *map, struct vio *vio,
 					bool back_merge)
 {
 	struct bio *bio = vio->bio;
-	sector_t merge_sector = get_bio_sector(bio);
+	sector_t merge_sector = bio->bi_iter.bi_sector;
 	struct vio *vio_merge;
 
 	if (back_merge)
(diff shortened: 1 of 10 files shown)


bjohnsto (approved):
Looks ok to me. I'm not a fan of the __ function name thing, but it is a kernely thing to do.

lorelei-sakai (commented):
Please wait to merge this for a bit.

lorelei-sakai (changes requested):
Just some general notes. First, you'll need to rebase this after Ken's changes go in. I suspect a good number of the whitespace changes will be subsumed by that series. Second, I'm a little surprised that this didn't need any internal adjustments. I guess we don't use these interface directly in our tests?
Third, I think you should squash the final commit into the one before it. It's just a minor change, and given previous comments from Mike, I think he would have meant to do it, he just overlooked it. (On a side note, if you did leave it separate, the commit title should reflect match the preferred format and start with "dm vdo io-submitter:", and you should sign it my using commit -s.)

slegendr:
Squashed the last commit. Will rebase after the CI testing is done.

src/c++/vdo/base/io-submitter.h
```diff
@@ -24,24 +24,23 @@ void process_vio_io(struct vdo_completion *completion);
...
-				       bio_end_io_t callback, vdo_action *error_handler,
-				       unsigned int operation)
+static inline void vdo_submit_metadata_vio(struct vio *vio, physical_block_number_t physical,
+					   bio_end_io_t callback, vdo_action *error_handler,
+					   unsigned int operation)
 {
-	vdo_submit_metadata_io(vio, physical, callback, error_handler, operation,
-			       vio->data);
+	__submit_metadata_vio(vio, physical, callback, error_handler, operation, vio->data);
 }
```
lorelei-sakai, line 36: I think I decided to wrap this line even though Mike didn't. Be mindful of that when rebasing.

src/c++/vdo/base/io-submitter.c
```diff
@@ -386,8 +386,7 @@ void __submit_metadata_vio(struct vio *vio, physical_block_number_t physical,
 		return;
 	}
 
-	vdo_set_completion_callback(completion, process_vio_io,
-				    get_vio_bio_zone_thread_id(vio));
+	vdo_set_completion_callback(completion, vdo_submit_vio, get_vio_bio_zone_thread_id(vio));
 	vdo_launch_completion_with_priority(completion, get_metadata_priority(vio));
```
lorelei-sakai, line 390: You should not be changing the wrapping here.

lorelei-sakai (commented):
Okay, the upshot of this is that it basically all looks fine, but there are more formatting changes in MIke's series that I had realized. I have a PR out to pull those into the big formatting change, then when you rebase again this should all be fine.

A few things to note: There are a couple places where it looks like you might be changing the line wrapping from what it is in the formatting series, just be careful you're not doing that after the next rebase. Also, I think everything in commit two got subsumed so that commit can go away.

src/c++/vdo/base/io-submitter.c
slegendr: Modifying the wrapping to put it on one line was in Mike's change to "rename to submit_vio and submit_data_vio".

src/c++/vdo/base/io-submitter.c
```diff
@@ -234,31 +231,40 @@ static struct vio *get_mergeable_locked(struct int_map *map, struct vio *vio,
...
 	return (get_bio_sector(vio_merge->bios_merged.head) == merge_sector ?
+=======
+		return (vio_merge->bios_merged.tail->bi_iter.bi_sector == merge_sector ?
+			vio_merge : NULL);
+	}
+
+	return (vio_merge->bios_merged.head->bi_iter.bi_sector == merge_sector ?
+>>>>>>> 1f46120 (dm vdo io-submitter: remove get_bio_sector)
 		vio_merge : NULL);
 }
```
lorelei-sakai, line 240: Conflict marker here. I think you can keep the second block.

src/c++/vdo/base/io-submitter.c
```diff
@@ -307,9 +313,14 @@ static bool try_bio_map_merge(struct vio *vio)
...
 		merged = false;
+<<<<<<< HEAD
 		result = vdo_int_map_put(bio_queue_data->map,
 					 get_bio_sector(bio),
 					 vio, true, NULL);
+=======
+		result = vdo_int_map_put(bio_queue_data->map, bio->bi_iter.bi_sector, vio,
+					 true, NULL);
+>>>>>>> 1f46120 (dm vdo io-submitter: remove get_bio_sector)
 	} else if (next_vio == NULL) {
```
lorelei-sakai, line 311: Another conflict marker, please keep the same line-wrapping as  HEAD.

src/c++/vdo/base/io-submitter.c
lorelei-sakai: Ah I see you fixed this, but in the second patch instead of the first. Should be in the first patch, though.

src/c++/vdo/base/io-submitter.c
lorelei-sakai: Similarly, I agree with the fix, but it should be in the first commit.

src/c++/vdo/base/slab-depot.c
```diff
@@ -1204,9 +1204,9 @@ static void write_reference_block(struct waiter *waiter, void *context)
 	WRITE_ONCE(block->slab->allocator->ref_counts_statistics.blocks_written,
 		   block->slab->allocator->ref_counts_statistics.blocks_written + 1);
```
lorelei-sakai, line 1206: It's a minor think, but I'd rather not remove the space on line 1207.

lorelei-sakai (changes requested):
Okay, generally looks good. Just a few minor things.
a) In the first commit message, there's a typo. ("acces")
b) in all the commit messages, can you remove the lines about fixing tyle nits? It looks like we did successfully deal with those in previous changes.
c) The first patch still has conflict markers in it. There's fixed in the second patch but they need to be fixed in the first patch. you can use rebase /edit to fix it. I can walk you through that if you like; I at least find it a little confusing to keep track of where you are during that process.
d) I marked one place where I deliberately left a space in even though Mike removed it.

lorelei-sakai (approved):

PR #38, opened 2023-11-14
[VDO-5602] Fix more formatting issues

lorelei-sakai:
These commits represent a few more formatting fixes picked out of the hash-mp, io-submitter, and wait-queue series. I thought they should got rolled into the general formatting change, especially in places where the formatting change changed lines differently than Mike did, resulting in multiple changes to the same code statements.

When going upstream, these changes will be squashed into the large reformatting commits.

### DIFF
--- a/src/c++/vdo/base/slab-depot.c
+++ b/src/c++/vdo/base/slab-depot.c
@@ -1223,14 +1223,14 @@ static void reclaim_journal_space(struct slab_journal *journal)
 	WRITE_ONCE(journal->events->flush_count, journal->events->flush_count + 1);
 	if (length < journal->flushing_deadline) {
 		/* Schedule more writes the closer to the deadline we get. */
-		write_count = max_t(block_count_t,
-				    write_count / (journal->flushing_deadline - length + 1),
-				    1);
+		write_count /= journal->flushing_deadline - length + 1;
+		write_count = max_t(block_count_t, write_count, 1);
 	}
 
-	for (written = 0; written < write_count; written++)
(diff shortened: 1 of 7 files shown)


src/c++/vdo/base/slab-depot.c
```diff
@@ -1223,14 +1223,15 @@ static void reclaim_journal_space(struct slab_journal *journal)
 	WRITE_ONCE(journal->events->flush_count, journal->events->flush_count + 1);
 	if (length < journal->flushing_deadline) {
 		/* Schedule more writes the closer to the deadline we get. */
-		write_count = max_t(block_count_t,
-				    write_count / (journal->flushing_deadline - length + 1),
-				    1);
+		write_count =
+			max_t(block_count_t, (write_count /
+					      (journal->flushing_deadline - length + 1)), 1);
```
raeburn, line 1228: I think this might look better with the division and "1);" on one line, making it two lines (well, three if you keep the line break after "="), if you don't like the original three-line version. Looks to me like it'll still be under 100 columns.

raeburn (changes requested):

src/c++/vdo/base/slab-depot.c
lorelei-sakai: Why yes, I do believe you are correct. I am mirroring what Kine did it change 	2efa17775e4f2ee4f9da99dc160cac6094eebfb0 "dm vdo wait-queue: add proper namespace to interface"
Think it's worth changing?

raeburn (approved):

PR #41, opened 2023-11-15
VDO-5597: Remove pointer map and clean up int map

bjohnsto:
This change is based on Mike's patchset to merge int-map and pointer-map into hash-map. While adapting that change to vdo-devel, we noticed there was really only one user of pointer-map (apart from tests) which was in dedupe.c. We've instead decided to remove pointer-map and keep int-map, while changing dedupe.c to use an int-map. We will also be applying some additional cleanup Mike did in his patchset to this PR.


src/c++/vdo/base/int-map.h
```diff
@@ -24,9 +24,9 @@
 struct int_map;
 
 int __must_check
-vdo_make_int_map(size_t initial_capacity, unsigned int initial_load, struct int_map **map_ptr);
+vdo_int_map_create(size_t initial_capacity, unsigned int initial_load, struct int_map **map_ptr);
```
lorelei-sakai, line 28: We skipped reformatting int-map, so could we do it in this series? In the second change seems reasonable.

I think just the signatures for vdo_int_map_create() and vdo_int_map_put() need to get wrapped, in both the header and the .c file.

.... If you reorder the patches to remove the initial_load parameter before the name changes, you shouldn't have to reflow the int_map_create() lines. That'd probably be gruesome to resolve though, maybe just fix the int_map_create() wrapping knowing you'll be shortening the lines lafter.

src/c++/vdo/tests/asyncLayer.c
```diff
@@ -202,9 +202,9 @@ void initializeAsyncLayer(PhysicalLayer *syncLayer)
 {
   AsyncLayer *asyncLayer;
   VDO_ASSERT_SUCCESS(uds_allocate(1, AsyncLayer, __func__, &asyncLayer));
-  VDO_ASSERT_SUCCESS(vdo_make_int_map(0,
-                                      0,
-                                      &asyncLayer->completionEnqueueHooksMap));
+  VDO_ASSERT_SUCCESS(vdo_int_map_create(0,
+					0,
+					&asyncLayer->completionEnqueueHooksMap));
   VDO_ASSERT_SUCCESS(uds_init_mutex(&asyncLayer->mutex));
```
lorelei-sakai, line 207: Really tempted to unwrap this line as long as you're touching it.

lorelei-sakai (changes requested):
This basic approach looks right. I'm actaully surprised it touches so little in dedupe.c

I'm not sure we need a helper for the three pales we transform the hash into a key, though. Maybe add a field for the key to the struct hash_lock, or call out a local variable in the function that calls it twice.

And since we skipped int-map during the formatting pass, could oyou do some like wrapping, too? I mentioned the small number of lines I think are too long.

src/c++/vdo/base/dedupe.c
```diff
@@ -374,6 +374,17 @@ struct pbn_lock *vdo_get_duplicate_lock(struct data_vio *data_vio)
...
 
+/**
+ * get_hash_lock_key() - Get the hash lock key to use in int-map.
+ * @locl: The hash lock.
+ *
+ * Return: The short string representing the state
+ */
+static u64 get_hash_lock_key(struct hash_lock *lock)
+{
+  return get_unaligned_le64(&(lock->hash.name));
```
raeburn, line 385: Extra parens are unneeded here

raeburn (changes requested):

src/c++/vdo/base/dedupe.c
```diff
@@ -374,6 +374,17 @@ struct pbn_lock *vdo_get_duplicate_lock(struct data_vio *data_vio)
 	return data_vio->hash_lock->duplicate_lock;
 }
 
+/**
+ * get_hash_lock_key() - Get the hash lock key to use in int-map.
+ * @locl: The hash lock.
+ *
```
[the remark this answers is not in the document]
bjohnsto: fixed. also fixed comment in Return part of function header.

src/c++/vdo/base/dedupe.c
```diff
@@ -1943,11 +1955,10 @@ void vdo_release_hash_lock(struct data_vio *data_vio)
 	if (lock->registered) {
 		struct hash_lock *removed;
 
-		removed = vdo_pointer_map_remove(zone->hash_lock_map, &lock->hash);
-		ASSERT_LOG_ONLY(lock == removed,
-				"hash lock being released must have been mapped");
+		removed = vdo_int_map_remove(zone->hash_lock_map, get_hash_lock_key(lock));
+		ASSERT_LOG_ONLY(lock == removed, "hash lock being released must have been mapped");
 	} else {
```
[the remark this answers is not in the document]
bjohnsto: fixed. also added local key variable but only in part of code where there were two uses in if / else

src/c++/vdo/base/int-map.h
bjohnsto: int-map.c and int-map.h should be formatted now.

src/c++/vdo/tests/asyncLayer.c
bjohnsto: done.

src/c++/vdo/base/dedupe.c
bjohnsto: removed extra parens

lorelei-sakai (approved):
I'll fix the squashing issues in post.

raeburn (approved):

lorelei-sakai (approved):

PR #42, opened 2023-11-17
[VDO-5579] dm vdo: Correct migration test failures

slegendr:
Some of the tests in SimpleMigration.pm have been failing nightly for several weeks due to being unable to load the kvdo module. The migration tests are a special case of tests that involve building the kvdo module manually within the Upgrade device logic. As such, the root issue is that the named versions in the tests (particularly 8.2.1.3 and 8.2.1-current) are unable to build against newer kernels. 

To correct the issue, the SupportedVersions.yaml and SupportedScenarios.yaml files needed to be modified to include more recent module versions that are able to successfully build against recent kernels, and the associated SimpleMigration tests needed to be modified to use the more recent module versions.

### DIFF
--- a/src/perl/Permabit/SupportedScenarios.yaml
+++ b/src/perl/Permabit/SupportedScenarios.yaml
@@ -14,20 +14,15 @@ X86_FEDORA37_head:
   arch: X86_64
   moduleVersion: head
 
-X86_RHEL9_8.2.1.3:
+X86_RHEL9_8.2.3.2:
   rsvpOSClass: RHEL9
   arch: X86_64
-  moduleVersion: 8.2.1.3
+  moduleVersion: 8.2.3.2
 
-X86_RHEL9_8.2.1-current:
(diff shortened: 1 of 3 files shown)


lorelei-sakai (approved):
In SimpleMigration.pm, the MultipleMigration test doesn't seem to be doing much of anything for us now since it isn't migrating across multiple version any more.   But it seems reasonable enough to leave as a placeholder, for now.

slegendr:
@lorelei-sakai - True. I modified it so that there are fewer paths for failure when the scenario versions used now eventually become obsolete and are not buildable against newer kernels. The purpose of testMultipleMigration is to test reserving the correct number of hosts and migration between more than two machines, which it does. So, no need to modify the version, as migration and upgrade is handled in a separate test. We could ultimately remove the test if we feel it no longer serves a purpose, or choose to have it as the default and remove testSimpleMigration.

PR #43, opened 2023-11-20
[TECH-151] Fix unifdef artifact

lorelei-sakai:
Explicitly define VDO_UPSTREAM in the uds kernel directory so that it will get stripped from uds-threads.c.

This removes the last difference between the tree as prepared by the overlay script and the actual upstream repo.


slegendr (approved):
Looks good to me.

PR #47, opened 2023-11-29
[TECH-156] Adapt to changes in the 6.7 kernel

lorelei-sakai:
This is actually about making the mainline-next kernel builds work. The latest kernel appears to be 6.7rc3, and one of the mitigations should only compile on a 6.7 or later kernel.

### DIFF
--- a/src/c++/uds/kernelLinux/tests/Makefile
+++ b/src/c++/uds/kernelLinux/tests/Makefile
@@ -90,6 +90,11 @@ $(ZUB_PACKAGE): $(UDS_TGZ) $(ZUB_DKMS_SOURCES) $(ZUB_DKMS_HEADERS) \
 	cp -p $(ZUB_DKMS_SOURCES) $(ZUB_SUBDIR)
 	cp -p $(ZUB_DKMS_HEADERS) $(ZUB_SUBDIR)
 	$(ALLTESTS_COMMAND) >$(ZUB_SUBDIR)/AllTests_t1.c
+	ENDIF=`tail -n 1 $(ZUB_SUBDIR)/albtest.h`; \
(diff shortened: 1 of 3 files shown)


slegendr (approved):
Looks good to me. Will the VDO_USE_ALTERNATE changes also be needed on Chlorine and to be pushed to dm-vdo/kvdo?

C2Redhat:
Also, how do one choose to use  VDO_USE_ALTERNATE or VDO_USE_ALTERNATE_2 ?

lorelei-sakai:
> Looks good to me. Will the VDO_USE_ALTERNATE changes also be needed on Chlorine and to be pushed to dm-vdo/kvdo?

Chlorine is currently using RHEL9 only, not Fedora. So if/when these block changes are backported into the RHEL9 kernel, we will have to make that adjustment there. Since these changes are not even in a released Fedora version yet, it will be a while.
However, I realized thinking about this that I am not handling the RHEL9.4 case correctly in this change. It should be doing the same branching it was before rather than also checking the kernel version. I will fix that issue.

lorelei-sakai:
> Also, how do one choose to use VDO_USE_ALTERNATE or VDO_USE_ALTERNATE_2 ?

If we do not want to use the newest code, then I always set VDO_USE_ALTERNATE. Since there are two older versions, though, I additional set VDO_USE_ALTERNATE_2 to select which of the two older patterns to use. My original logic for RHEL versions was not correct, which may have caused some confusion here.

slegendr (approved):
Looks good. Yeah, now the setting of VDO_USE_ALTERNATE and VDO_USE_ALTERNATE_2 make more sense.

C2Redhat (approved):

PR #49, opened 2023-11-29
dm vdo: fix how dm_kcopyd_client_create() failure is checked

C2Redhat:
dm_kcopyd_client_create() returns an ERR_PTR so its return must be checked with IS_ERR().

This change is based on a suggestion from Mike Snitzer, in https://git.kernel.org/pub/scm/linux/kernel/git/device-mapper/linux-dm.git with commit id: 9b935003bbb46f7cefa8c6e7b17b0cf457a65483

We need to include <linux/err.h> to use IS_ERR and ERR_PTR.

### DIFF
--- a/src/c++/vdo/base/dm-vdo-target.c
+++ b/src/c++/vdo/base/dm-vdo-target.c
@@ -8,6 +8,7 @@
 #include <linux/completion.h>
 #include <linux/delay.h>
 #include <linux/device-mapper.h>
+#include <linux/err.h>
 #include <linux/module.h>
 #include <linux/mutex.h>
 #include <linux/spinlock.h>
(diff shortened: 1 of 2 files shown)


raeburn (approved):

lorelei-sakai (changes requested):
The code changes look good. The commit message should be written with the upstream audience in mind. So keep the first setence, but cut the section about "base on a suggestion from Mike" We should either be using his change directly (i.e., he is the author) or not mention it.

The comment about including linux/err.h is probably obvious enough that we don't need it, also.

Both of those notes can be in the PR descripton, but they should not be in the commit message that would go upstream.

lorelei-sakai (approved):

PR #51, opened 2023-12-01
Add additional cleanup and small improvements

bjohnsto:
Have done some additional cleanup to the ci action. 

1. display info at start of every ci run. also fixed log file name at this point.
2. assign reviewers even for draft prs
3. remove trailing spaces
4. fix indenting
5. remove unused REPO variable

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,8 +209,6 @@ jobs:
       - name: Copy log file to /permatbit/user/bunsen/artifacts
         if: failure()
         run: |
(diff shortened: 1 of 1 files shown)


lorelei-sakai (commented):
These changes all seems reasonable.

The public testing failure is a bit concerning, though. Maybe the workflow around having a draft PR that is then made real isn't set up quite right? It's an http 403 error, which makes me think maybe the secrets aren't properly accessible for this PR somehow.

lorelei-sakai (approved):
Looks good, pending testing.

PR #54, opened 2023-12-06
[VDO-5591] Try again to fix log file based on context dump

bjohnsto:
We want to output useful information about the PR that is firing the CI job. The info includes the org and repo the pr is for. We also want to use org/repo as part of a unique name for log files. 

Unfortunately we've been using the wrong field names for those values. After printing out the github event context when the event fired, we were able to determine that the context for what we want is github.event.repository.name and github.event.repository.owner.login for repo name and repo organization, respectively.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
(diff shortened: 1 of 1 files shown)


lorelei-sakai:
Could you add more detail to the commit message? A good rule of thumb is to explain why you're making this change, to someone who is reading this comment in six months and doesn't remember any of the relevant context. (The content looks fine, assuming that the fields are now correct; without a pointer the context structure itself I can't check that.)

lorelei-sakai (approved):
Looks good now.

PR #62, opened 2024-01-09
Add makefile target to support testing dm-linux repo

bjohnsto:
Add a makefile target to the kernel makefile to support building a dkms module using the source code from the dm-linux repo so that we can test it using our perl testing infrastructure.

### DIFF
--- a/src/c++/vdo/kernel/Makefile
+++ b/src/c++/vdo/kernel/Makefile
@@ -59,7 +59,7 @@ DKMS_FILES =	$(addprefix $(VDO_BASE_DIR)/,$(KERNEL_SOURCES))		\
 		$(filter-out $(VDO_BASE_DIR)/list.h,			\
 			$(wildcard $(VDO_BASE_DIR)/*.h))
 
-# The kernel Makefile treats V undefined the same as V equals 0 
+# The kernel Makefile treats V undefined the same as V equals 0
 V ?= 0
 
 # If VDO_PRESERVE_DKMS is defined then leave the intermediate products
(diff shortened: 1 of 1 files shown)


src/c++/vdo/kernel/Makefile
```diff
@@ -101,6 +101,20 @@ endif
 build-against-latest-kernel:
 	$(MAKE) $(DKMS_TGZ) KERNEL_SOURCE_DIR=/permabit/not-backed-up/kernels/mainline/
 
+.PHONY: build-using-dm-linux
+build-using-dm-linux:
+	$(MAKE) $(DKMS_TGZ) VDO_PRESERVE_DKMS=1
+	rm -fr $(VDO_SUBDIR)/*.[cho]
```
lorelei-sakai, line 107: I could be misreading this, but it looks like this build target is doing way more work that it needs to do. It looks like you're building the regular dkms module, then deleting that module and replacing all the files before building the module you really want.

I think it would be cleaner to call out a new dkms preparation target to capture the common bits between the normal dkms build and your new one, and then have both targets rely on that common setup target. Then your dm-linux build target can just copy in the files it wants without erasing things you don't need.

Along similar lines, I think it would make sense to have a custom Makefile for this build (similar to the Makefile.module.in for the regular dkms build) that you can just copy in, instead of trying to edit the existing Makefile inline. I think that would make it easier to update the Makefile later if we need to. This step could be put off for now, though, since Mike has a pending change to give dm-vdo upstream its own Makefile, and that would affect what you do here. So we can clean the Makefiles up after that change goes in.

lorelei-sakai (changes requested):
Could we cut down the new make target so it doesn't rely on doing the whole normal dkms build first? I think that means factoring out the common parts into a ne wphny make target.

PR #66, opened 2024-01-23
Allow fsck exit status 1 (filesystem errors corrected). [VDO-5638]

raeburn:
Fedora 39 has fsck support for checking the "orphan file" status in ext4, and will report if the orphan_present flag is set but the orphan file is empty. It's a reasonable and normal state if file system updates are interrupted in between metadata blocks, as they are for rebuild tests, since multiple blocks can't get updated atomically.

### DIFF
--- a/src/perl/Permabit/FileSystem.pm
+++ b/src/perl/Permabit/FileSystem.pm
@@ -133,7 +133,19 @@ sub fsck {
   my ($self) = assertNumArgs(1, @_);
   my $machine = $self->{machine};
   my $devPath = $self->{device}->getSymbolicPath();
-  $machine->runSystemCmd("sudo fsck -t $self->{fsType} -a $devPath");
+  my $fsckCommand = "sudo fsck -t $self->{fsType} -a $devPath";
+  $log->debug("$machine->{hostname}: $fsckCommand");
+  my $errno = $machine->sendCommand($fsckCommand);
+  if ($errno > 1) {
+    confess("fsck failed on $machine->{hostname}: $errno\n"
+            . $machine->_getCmdOutput());
+  } elsif ($errno == 1) {
+    # Some "problems" like metadata in two different locations not
+    # getting updated together atomically can come up naturally and
+    # reasonably, and may be reported by fsck. (VDO-5638)
+    $log->info("fsck fixed issues on $machine->{hostname}:\n"
+               . $machine->_getCmdOutput());
+  }
 }
 
 #############################################################################


lorelei-sakai:
Ken, can you rebase this PR to get rid of the merge commit? Unfortuantely the default button for "update this branch" makes an extra merge commit and it'll make our history even more of a mess if we include those. Thanks!

src/perl/Permabit/FileSystem.pm
```diff
@@ -133,7 +133,20 @@ sub fsck {
...
+  my $errno = $machine->sendCommand($fsckCommand);
+  if ($errno > 1) {
+    confess("fsck failed on $machine->{hostname}: $errno\n"
+            . $machine->_getCmdOutput());
+  } elsif ($errno == 1) {
+    # Some "problems" like metadata in two different locations not
+    # getting updated together atomically can come up naturally and
+    # reasonably, and may be reported by fsck. (VDO-5638: "feature
+    # orphan_present is set but orphan file is clean")
+    $log->info("fsck fixed issues on $machine->{hostname}:\n"
```
lorelei-sakai, line 147: We probably don't need the ticket title; the ticket ID should be reference enough.

src/perl/Permabit/FileSystem.pm
```diff
@@ -133,7 +133,20 @@ sub fsck {
   my ($self) = assertNumArgs(1, @_);
   my $machine = $self->{machine};
   my $devPath = $self->{device}->getSymbolicPath();
-  $machine->runSystemCmd("sudo fsck -t $self->{fsType} -a $devPath");
+  my $fsckCommand = "sudo fsck -t $self->{fsType} -a $devPath";
+  $log->debug("$machine->{hostname}: $fsckCommand");
```
lorelei-sakai, line 137: You basically just unrolled runSystemCommand, didn't you? That's probably fine for now but in the long run it might make sense to add an extra parameter or an extra method to allow error codes besides 0.

lorelei-sakai (changes requested):
The test fix looks reasonable. Some small comments inline.

Could you get rid of the second (merge) commit, though? I think it'll make our history messy for no good reason.

src/perl/Permabit/FileSystem.pm
raeburn: Yes. We've got an annoying random assortment of logging vs not logging and allowing nonzero exit status vs blowing up on failure. A variant to log the command and return the exit status would be helpful.

PR #69, opened 2024-01-24
[VDO-5642] Disable Volume_n4 StumblingGet test

lorelei-sakai:
This test is failing intermittently since we moved to the pmifarms. Disable it so that it doesn't interfere with upstream development.

### DIFF
--- a/src/c++/uds/src/tests/Volume_n4.c
+++ b/src/c++/uds/src/tests/Volume_n4.c
@@ -525,10 +525,11 @@ static void testMultiThreadStress4Async(void)
 static const CU_TestInfo tests[] = {
   {"Invalid Read Queue", testInvalidateReadQueue},
   {"SequentialGet",      testSequentialGet},
-  {"StumblingGet",       testStumblingGet},
   {"Full Read Queue",    testFullReadQueue},
   {"MT Stress 1 async",  testMultiThreadStress1Async},
   {"MT Stress 4 async",  testMultiThreadStress4Async},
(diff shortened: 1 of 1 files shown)


src/c++/uds/src/tests/Volume_n4.c
```diff
@@ -525,11 +525,12 @@ static void testMultiThreadStress4Async(void)
 static const CU_TestInfo tests[] = {
   {"Invalid Read Queue", testInvalidateReadQueue},
   {"SequentialGet",      testSequentialGet},
-  {"StumblingGet",       testStumblingGet},
   {"Full Read Queue",    testFullReadQueue},
   {"MT Stress 1 async",  testMultiThreadStress1Async},
   {"MT Stress 4 async",  testMultiThreadStress4Async},
   CU_TEST_INFO_NULL,
+  {"StumblingGet",       testStumblingGet},
+  CU_TEST_INFO_NULL,
```
raeburn, line 533: The extra null entry isn't needed since the scanning won't get that far.
OTOH it doesn't hurt anything either; just an observation, I'll approve either way.

raeburn (approved):

src/c++/uds/src/tests/Volume_n4.c
lorelei-sakai: I like to leave the final null in so you don't forget it when re-enabling the test. I agree it isn't necessary, but i'm going to leave it in.

PR #70, opened 2024-01-24
[VDO-5655] dm vdo data-vio: rename is_trim flag to is_discard

lorelei-sakai:
Eliminate use of "trim" in favor of "discard" since it reflects the top-level Linux discard primative rather than the ATA specific ditto.


slegendr (approved):
Looks good to me.

PR #71, opened 2024-01-25
[VDO-5656] Properly handle kernel-only files from upstream dm-vdo

lorelei-sakai:
Now that dm-vdo upstream has its own Makefile and Kconfig file, we need to account for these in our tooling.

For ingestion, create a local copy of these files in the kpatch directory, and apply any upstream changes to our local copies.
For the overlay, add these files to the upstream sources before overlaying the new files.


slegendr (approved):
Looks good to me.

PR #72, opened 2024-01-25
Add support to test dm-linux using vdo-devel perl infra

bjohnsto:
In order to test dm-vdo/dm-linux using our current perl infrastructure, we're going to create a kvdo srpm using source files from the dm-vdo/dm-linux repo. Then we will run vdotests using that srpm instead of the normal one built using dm-vdo/vdo-devel sources. This PR contains the support to build the srpm and run the tests using it. The main target to do all this is in src/Makefile and will be called by the CI for dm-linux. That work is not part of this PR and instead will be in a PR for the dm-vdo/dm-linux repo once this code is checked in. The changes for this PR consist of the following:
src/Makefile - the target to be called from the CI to create the SRPM and run the tests.
src/packaging/dmlinux/* - the code to create the SRPM using the dm-linux sources. Requires setting of DMLINUX_SOURCE_DIR to point at the dm-linux sources.
src/perl/vdotest/* - changes to the test infrastructure to run the test using the SRPM we've created.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -133,6 +133,16 @@ jenkins-parallel: clean
 	$(MAKE) check-style
 	$(MAKE) test-vdo-distribution
 
+.PHONY: dmlinux-jenkins
+dmlinux-jenkins:
+	$(MAKE) -j$(NUM_JOBS) dmlinux-jenkins-parallel
+dmlinux-jenkins-parallel: export VDO_TESTINDEX=$(TEST_INDEX)
+dmlinux-jenkins-parallel: clean
+	$(MAKE) all
+	$(MAKE) -C packaging/dmlinux clean all
+	@echo "Using index name of $$VDO_TESTINDEX"
+	$(MAKE) DMLINUX=1 vdotests
+
 .PHONY: archive
 archive:
 	$(MAKE) -C $(ARCHIVE)
--- a/src/packaging/dmlinux/Makefile
+++ b/src/packaging/dmlinux/Makefile
@@ -1,0 +1,56 @@
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Copyright Red Hat
+#
+
+SRC_DIR = ../..
+include ../defines
+
+RPMBUILD      = rpmbuild
+RPMBUILD_FLAG = -bs
+BUILD_DIR    ?= $(realpath .)/build
+WORK_DIR     ?= $(realpath .)/work
+PACKAGE       = kvdo-$(VDO_VERSION)
+PREPARED_DIR  = $(WORK_DIR)/$(PACKAGE)
+SOURCES       = $(BUILD_DIR)/SOURCES
+
+SPECS         = $(BUILD_DIR)/SPECS
+
+# rpmbuild fails to clean up after itself on RHEL8 if BUILDROOT is under
+# BUILD_DIR in NFS so just pick a temporary local directory for it to use.
+BUILDROOT := $(shell mktemp -du BUILDROOT-XXXXXXXXXX --tmpdir)
+
+RPMBUILD_ARGS = --define '_topdir $(realpath $(BUILD_DIR))' \
+		--buildroot=$(BUILDROOT)
+
+.PHONY: all
+all: kvdo-srpm
+
+.PHONY: clean
+clean:
+	rm -rf $(BUILD_DIR) $(WORK_DIR) prepare.out
+
+prepare: prepare.out
+
+.PHONY: FORCE
+prepare.out: FORCE
+	$(if $(DMLINUX_SOURCE_DIR),,$(error Must set DMLINUX_SOURCE_DIR))
+	rm -rf $(WORK_DIR)
+	mkdir -p $(WORK_DIR)/$(PACKAGE)/dm-vdo
+	sed -e "s/@VERSION@/$(VDO_VERSION)/g" kvdo.spec >$(WORK_DIR)/kvdo.spec
+	cp kernel/Makefile $(WORK_DIR)/$(PACKAGE)
+	cp $(DMLINUX_SOURCE_DIR)/drivers/md/dm-vdo/*.[ch] $(WORK_DIR)/$(PACKAGE)/dm-vdo > prepare.out
+
+$(SOURCES): prepare
+	mkdir -p $(SOURCES)
+	tar -C $(WORK_DIR) -zcvf $(SOURCES)/$(PACKAGE).tgz ./$(PACKAGE)
+
+SRPM =	mkdir -p $(SPECS);					\
+	cp $(WORK_DIR)/$(1) $(SPECS) && cd $(SPECS) &&		\
+	$(RPMBUILD) $(RPMBUILD_FLAG) $(RPMBUILD_ARGS) $(1)
+
+
+.PHONY: kvdo-srpm
+kvdo-srpm: $(SOURCES)
+	$(call SRPM,kvdo.spec)
+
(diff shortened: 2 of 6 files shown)


src/Makefile
```diff
@@ -11,7 +11,7 @@ include $(CURRENT_VERSION_FILE)
 export VDO_VERSION
 
 SUBDIRS                 := perl c++ python tools
-CLEAN_SUBDIRS           := $(SUBDIRS) stats packaging
+CLEAN_SUBDIRS           := $(SUBDIRS) stats packaging packaging/dmlinux
 CHECK_SUBDIRS           := perl tools python/vdo
```
lorelei-sakai, line 15: Please don't skip levels like this. Instead, update the Makefile in src/packaging to clean the dmlinux directory.

src/Makefile
```diff
@@ -133,6 +133,16 @@ jenkins-parallel: clean
...
 	$(MAKE) test-vdo-distribution
 
+.PHONY: dmlinux-jenkins
+dmlinux-jenkins:
+	$(MAKE) -j$(NUM_JOBS) dmlinux-jenkins-parallel
+dmlinux-jenkins-parallel: export VDO_TESTINDEX=$(TEST_INDEX)
+dmlinux-jenkins-parallel: clean
+	$(MAKE) all
+	$(MAKE) -C packaging/dmlinux clean all
+	@echo "Using index name of $$VDO_TESTINDEX"
```
lorelei-sakai, line 143: You sohuldn't need "clean" here; the clean target should have done that for you. You also might not need "all" (since it should be the default, or you can set it to be).

src/packaging/dmlinux/Makefile
```diff
@@ -0,0 +1,60 @@
...
+	cp $(WORK_DIR)/$(1) $(SPECS) && cd $(SPECS) &&		\
+	$(RPMBUILD) $(RPMBUILD_FLAG) $(RPMBUILD_ARGS) $(1)
+
+
+.PHONY: kvdo-srpm
+kvdo-srpm: $(SOURCES)
+	$(call SRPM,kvdo.spec)
+
+.PHONY: srpms
+srpms:
```
lorelei-sakai, line 58: Again, srpms only has one step, it doesn't need to be a separate target.

src/packaging/dmlinux/kernel/Makefile
```diff
@@ -0,0 +1,20 @@
...
+EXTRA_CFLAGS =	-std=gnu11					\
+		-fno-builtin-memset				\
+		-fno-omit-frame-pointer				\
+		-fno-optimize-sibling-calls			\
+		-Werror						\
+		$(if $(CONFIG_KASAN),,-Wframe-larger-than=400)	\
+		-DVDO_VERSION=\"$(VDO_VERSION)\"		\
+		-DSTATIC=					\
+		-DVDO_INTERNAL					\
+		$(INCLUDES)
```
lorelei-sakai, line 16: We shoudl reconsider these compile flags to more closely reflect what the kernel does. I suppose we can defer that step to a later change, though.

At a minimum, the STATIC and VDO_INTERNAL lines should not be needed because all that stuff should already have been replaced in the kernel sources. (And we don't want to mask stuff if we make a mistake.) The VDO_VERSION line is similarly unnecessary because the upstream version current hard-codes a version number.

src/packaging/dmlinux/kvdo.spec
```diff
@@ -0,0 +1,79 @@
...
+%clean
+rm -rf $RPM_BUILD_ROOT
+
+%files
+%defattr(644,root,root,755)
+%{_usr}/src/%{kmod_name}-%{version}
+
+%changelog
+* Fri Aug 04 2023 - Susan LeGendre-McGhee <slegendr@redhat.com> - 8.3.0.0-1
+- Initial spec file for internal test rpms. Spec file is distribution agnostic.
```
lorelei-sakai, line 79: If you're going to include a changelog in the spec file, please update it.

src/packaging/dmlinux/Makefile
```diff
@@ -0,0 +1,60 @@
...
+include ../defines
+
+RPMBUILD      = rpmbuild
+RPMBUILD_FLAG = -bs
+BUILD_DIR    ?= $(realpath .)/build
+WORK_DIR     ?= $(realpath .)/work
+PACKAGE       = kvdo-$(VDO_VERSION)
+PREPARED_DIR  = $(WORK_DIR)/$(PACKAGE)
+SOURCES       = $(BUILD_DIR)/SOURCES
+
```
lorelei-sakai, line 16: Where does VDO_VERSION come from, here? It's not really defined upstream (other than the kernel version itself), so I want to make sure we're not relying on something irrelevant.

lorelei-sakai (changes requested):
Generally seems reasonable. I had some questions about where the version number comes from, and there are some  Makefile things that could be simplified away.

src/packaging/dmlinux/kernel/Makefile
bjohnsto: I don't think we want to mess with the version number for now. We use it in the other srpms/rpms to depend on a particular version or higher. If we try to plug in the kernel version, things are gonna break. I think we need to look at the versioning stuff as a whole later on.

src/Makefile
bjohnsto: I don't think we should have clean and all from packaging contain dmlinux. dmlinux is very separate from the default make stuff and shouldnt add time to those targets for what is a very specific thing. When I decided this, I meant to remove packaging/dmlinux from CLEAN_SUBDIRS but forgot. This is especially true for all since dmlinux all requires DMLINUX_SOURCE_DIR to be set. If we want to add dmlinux just to the clean part of packaging I could maybe get behind that, but I think having clean all in the make target dmlinux-jenkins-parallel itself makes the most sense

src/Makefile
bjohnsto: See above on my thoughts

src/packaging/dmlinux/Makefile
bjohnsto: Will fix.

src/packaging/dmlinux/kvdo.spec
bjohnsto: Will fix.

src/Makefile
lorelei-sakai: Just removing packaging/dmlinux from this list seems fine.

src/Makefile
lorelei-sakai: If dmlinux realy is going to be a separate thing, you shouldn't need the dependence on the "clean" target, and you shouldn't nee  the generic "make all" before building the dmlinux stuff. Maybe it'll be clearer once you update it.

src/packaging/dmlinux/kernel/Makefile
lorelei-sakai: My first thought is that if we've got other rpms relying on this then we have a bigger problem to sort out. The vdo rpms already have to work with real kernels so there should be some way to address the issue. But we can defer it to a later change.

src/Makefile
bjohnsto: I was trying to say its not part of the normal build and test of vdo-devel. the dmlinux target itself still needs all the other things to be built (I think) so that the testing will work after we build the dmlinux source rpm. So I'm saying I don't think the dmlinux packaging dir belongs in the packaging Makefile (the default build and test) but instead should just be handled in the particular dmlinux target.

src/packaging/dmlinux/Makefile
bjohnsto: I believe it comes from the include of the defines file

src/Makefile
lorelei-sakai: Okay, I guess this seems fine.

lorelei-sakai (approved):
This looks fine for now. We can figure out how to clean this up more as we start using it.

PR #74, opened 2024-01-26
[VDO-5592] Remove nr_requests check

bjohnsto:
Remove nr_requests from Sysfs.pm test, as the kernel source has been changed so that the req based drivers are the only ones that have that field now.


lorelei-sakai (approved):
I think it would be slightly better to add more detail to the actual git commit message as well, but it's not a big deal here.

PR #75, opened 2024-01-26
Incorporate struct renaming patches from Mike Snitzer [VDO-5659,VDO-5660]

raeburn:
Mike's patches and fixups to internal/test code


lorelei-sakai (approved):
All looks good to me.

PR #76, opened 2024-01-27
Copy Makefile and Kconfig during prepare step

lorelei-sakai:
Copy our Makefile and Kconfig files into the dm-vdo directory in the prepare step instead of in checkpatch. This fixes an error from PR vdo-devel/71, which put this operation in the wrong target.


slegendr (approved):
Looks good to me.

PR #77, opened 2024-01-29
[VDO-5661] Ingest sparse warning patches from upstream

slegendr:
Ingest Mike's upstream patch series to fix sparse warnings.

The additional commits in this series apply the upstream changes to userspace, correcting build failures that arose.


lorelei-sakai (changes requested):
Mostly looks good! Just a few small tweaks.

I think patch 7 should get squashed into patch 6. (Mike shoudl have nitced and fixed that in his change.)
Change nine (adding the macro definitions) should probably either be broken up so the definition of __releases() is added right after it's used, or possibly the whole change could get moved to before commit 6 (adding the typedefs before any code uses them).
To be really nitpicky, the title of commit 4 shouldn't end in a period, but you don't have to re-push just to fix that.

slegendr:
@lorelei-sakai The requested changes were made.

lorelei-sakai (approved):
This looks good.

PR #78, opened 2024-02-01
[VDOSTORY-283] Replace generated stats files with static files

C2Redhat:
The stats generation code is somewhat cumbersome, and the generated files
 are not easy to keep in sync with upstream, especially when contributors
 want to edit them. It is likely simpler from a maintenance point-of-view
 to check in static versions for the generated files, and bear the cost of
 remembering to keep the various files in sync in the rare cases when we
 add or remove a statistics field.

In this PR, we added all the files that were generated before in the
repo and updated comments in each file.
     
We also reflect these changes by updating .gitignore, Makefile and defines.
Stats files are no longer generated, .gitignore will not ignore those
generated files any longer.

And in the second commit, we removed files that we longer needed.


src/c++/vdo/base/message-stats.c
```diff
@@ -0,0 +1,1216 @@
+// SPDX-License-Identifier: GPL-2.0-only
+/*
+ * Copyright 2023 Red Hat
+ *
+ */
+
```
lorelei-sakai, line 6: Please tighten up the header comment by removing the blank line on line 4. It should look like the header message in other vdo/base .c files.

src/c++/vdo/base/pool-sysfs-stats.c
```diff
@@ -0,0 +1,2684 @@
+// SPDX-License-Identifier: GPL-2.0-only
+/*
+ * Copyright 2023 Red Hat
+ *
+ */
```
lorelei-sakai, line 5: Same comment about fixing the copyright comments.

src/c++/vdo/base/statistics.h
```diff
@@ -0,0 +1,289 @@
...
+ *
+ * ATTENTION!!! ATTENTION!!! ATTENTION!!! ATTENTION!!! ATTENTION!!!
+ *
+ * If you add new statistics, be sure to update the following files:
+ *
+ * ../../../c++/vdo/base/message-stats.c
+ * ../../../c++/vdo/base/pool-sysfs-stats.c
+ * ../../../c++/vdo/user/messageStatsReader.c
+ * ../../../c++/vdo/user/vdoStatsWriter.c
+ * ../../../perl/Permabit/Statistics/Definitions.pm
```
lorelei-sakai, line 13: These paths can be shortened. The vdo/base files are in this directory. The vd/user files  only need to be  ../user/[foo.c]

ALso pleaes remove the ATTENTION line  (lines 5-6)

src/c++/vdo/user/Makefile
```diff
@@ -68,10 +65,10 @@ DMEVENT_LIB = libdevmapper-event-lvm2vdo.so
 LOGDIR ?= $(CURDIR)
 
 .PHONY: all
-all: $(STATS_HEADER) $(BIN_PROGS) $(DMEVENT_LIB)
+all: $(BIN_PROGS) $(DMEVENT_LIB)
 
-$(STATS_HEADER) $(STATS_FILES): $(VDO_STATS_DEPS)
-	$(MAKE) -C $(VDO_STATS_DIR)
+#$(STATS_HEADER) $(STATS_FILES): $(VDO_STATS_DEPS)
+#	$(MAKE) -C $(VDO_STATS_DIR)
```
lorelei-sakai, line 72: We should actually remove this make rule instead of just commenting it out.

src/c++/vdo/user/messageStatsReader.c
```diff
@@ -0,0 +1,1272 @@
...
+ *
+ * If you add new statistics, be sure to update the following files:
+ *
+ * ../../../c++/vdo/base/statistics.h
+ * ../../../c++/vdo/base/message-stats.c
+ * ../../../c++/vdo/base/pool-sysfs-stats.c
+ * ../../../c++/vdo/user/vdoStatsWriter.c
+ * ../../../perl/Permabit/Statistics/Definitions.pm
+ *
+ */
```
lorelei-sakai, line 15: Similar comment on shortening these paths. Alnd please remove the ATTENTION line and any unnecessary blank lines (like line 14).

src/c++/vdo/user/vdoStatsWriter.c
```diff
@@ -0,0 +1,1447 @@
...
+ * ATTENTION!!! ATTENTION!!! ATTENTION!!! ATTENTION!!! ATTENTION!!!
+ *
+ * If you add new statistics, be sure to update the following files:
+ *
+ * ../../../c++/vdo/base/statistics.h
+ * ../../../c++/vdo/base/message-stats.c
+ * ../../../c++/vdo/base/pool-sysfs-stats.c
+ * ../../../c++/vdo/user/messageStatsReader.c
+ * ../../../perl/Permabit/Statistics/Definitions.pm
+ *
```
lorelei-sakai, line 14: Same ccoment as the messageStatsReader.c, just noting this comment needs to change, too.

src/perl/Permabit/Statistics/Definitions.pm
```diff
@@ -0,0 +1,1029 @@
...
+#
+# If you add new statistics, be sure to update the following files:
+#
+# ../../../c++/vdo/base/statistics.h
+# ../../../c++/vdo/base/message-stats.c
+# ../../../c++/vdo/base/pool-sysfs-stats.c
+# ../../../c++/vdo/user/messageStatsReader.c
+# ../../../c++/vdo/user/vdoStatsWriter.c
+#
+##
```
lorelei-sakai, line 24: The perl file is really downstream form the .c c changes, so I think we don't need the file list here. This would mean cutting lines 13-23.

src/perl/Permabit/Statistics/Makefile
```diff
@@ -5,4 +5,3 @@ MANPREFIX := Permabit::Statistics::
 include $(ROOT)/Makefile.common
```
lorelei-sakai, line 5: It kind of seems like we don't need the perl/Permabit/Statistics makefile at all, now. Can we just get rid of it? Maybe as part of the second commit...

src/perl/Permabit/Statistics/.gitignore
```diff
@@ -1 +0,0 @@
-Definitions.pm
```
lorelei-sakai, line 1: This file (src/perl/Permabit/Statistics/.gitignore) should really be part of the first commit since it's about not ignoring the newly-committed perl file.

lorelei-sakai (changes requested):
In addition to the specific comments, I think the second commit can also remove all the perl stats Generation code. I think that's everything in the src/perl/statsgenerator directory.

src/perl/Permabit/Statistics/Makefile
C2Redhat: I was wondering if we should get rid of it.  I left it there in case we think people want to add things in the future.  I will remove it.

src/c++/vdo/base/statistics.h
raeburn: The version in the upstream submission shouldn't mention the non-kernel files. Maybe two comments, one inside an #ifdef ?

lorelei-sakai: Oh, Ken makes a good point. I'd maybe put the entire note about other files in an #ifdef. Especially since pool-sysfs-stats.c is also soon going to not be in the kernel build.

lorelei-sakai (changes requested):
Mostly looks good. Just Ken's note on statistics.h, and make sure it passes "'make jenkins"

src/c++/vdo/user/vdoStatsWriter.c
```diff
@@ -0,0 +1,1444 @@
...
+ * If you add new statistics, be sure to update the following files:
+ *
+ * ../base/statistics.h
+ * ../base/message-stats.c
+ * ../base/pool-sysfs-stats.c
+ * ./messageStatsReader.c
+ * ../../../perl/Permabit/Statistics/Definitions.pm
+ */
+
+
```
lorelei-sakai, line 14: extra blank line at 13-14.

lorelei-sakai (approved):
This all looks good to me. Private testing failed in BlockMapTreeWrite_t1 though.  That test shouldn't be affected by this series, so I restarte the tests, and if they pass this time it's all good.

I hope we don't have another rare race, though.

src/c++/vdo/user/vdoStatsWriter.c
C2Redhat: I have updated the file and removed the blank line.  Thanks.

lorelei-sakai (approved):
Looks good now.

PR #81, opened 2024-02-08
[VDO-5679] Restyle formerly generated files now checked in

raeburn:


lorelei-sakai (approved):
The code changes look good. Thank you for doing this.

One minor nit, I find the commit title for the message-stats.c commit a bit confusing, but I may just edit it when moving upstream. So, approved.

PR #85, opened 2024-02-13
[VDO-5678] Ingest formatting patches from upstream

slegendr:
Ingest Mike's upstream patches regarding small nits and formatting fixes. 

The first commit was modified to make necessary includes being removed in the upstream patch available in userspace and internal testing.


lorelei-sakai (approved):
There's a typo in the second commit message ("refelcts") that's in Mike's oiginal. You can fix it, or I can fix it when moving it upstream. Everything else looks good.

PR #87, opened 2024-02-13
[VDO-5677] dm vdo logger: update logging to start with "device-mapper: vdo"

bjohnsto:
[VDO-5677] Ingest commits: logging changes

Stops short of actually using DM's various logging macros (e.g. DMERR, DMINFO, etc) because VDO's logger isn't quite compatible with them.

Also switch emit_log_message_to_kernel() from open-coding printk with log-level to using corresponding pr_ macro.

### DIFF
--- a/src/c++/uds/src/uds/logger.h
+++ b/src/c++/uds/src/uds/logger.h
@@ -98,9 +98,6 @@ int uds_vlog_strerror(int priority, int errnum, const char *module, const char *
 #define uds_log_info_strerror(errnum, ...) \
 	uds_log_strerror(UDS_LOG_INFO, errnum, __VA_ARGS__)
 
-#define uds_log_notice_strerror(errnum, ...) \
-	uds_log_strerror(UDS_LOG_NOTICE, errnum, __VA_ARGS__)
-
 #define uds_log_warning_strerror(errnum, ...) \
 	uds_log_strerror(UDS_LOG_WARNING, errnum, __VA_ARGS__)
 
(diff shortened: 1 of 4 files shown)


lorelei-sakai (changes requested):
This really needs to have the other logging change in front of it so we don't get merge conflicts when rebasing.

bjohnsto:
> This really needs to have the other logging change in front of it so we don't get merge conflicts when rebasing.

done

src/c++/uds/src/uds/logger.h
```diff
@@ -37,11 +38,8 @@
...
 
 #ifdef __KERNEL__
-#if defined(MODULE)
-#define UDS_LOGGING_MODULE_NAME THIS_MODULE->name
-#else /* compiled into the kernel */
-#define UDS_LOGGING_MODULE_NAME "vdo"
-#endif
+#define DM_MSG_PREFIX "vdo"
+#define UDS_LOGGING_MODULE_NAME DM_NAME ": " DM_MSG_PREFIX
 #else /* userspace */
```
lorelei-sakai, line 43: I would like to keep the #if defined(MODULE) clause, inside an #ifndef VDO_UPSTREAM, in case some of our perl tests rely on the logging that produces.  This can be cone in this commit (the second), or it can be done with a follow-up internal only patch. But I'd prefer not to lose it.

lorelei-sakai (changes requested):
We talked about keeping the old UDS_LOGGING_MODULE_NAME internally, but it looks like you didn't make that change. And the first comit should be signed.

Everything else looks good.

src/c++/uds/src/uds/logger.h
bjohnsto: I think this is a bad reason to keep this for internal. We should use and test what the kernel uses. If tests break, we should fix the tests.

lorelei-sakai (approved):
Looks good.

PR #88, opened 2024-02-14
[VDO-5683] Move semaphore and barrier code into sparse cache

lorelei-sakai:
As far as the kernel is concerned, sparse_cache is the only file using a barrier, and the barrier is the only code using semaphores. So Mike simplified it down and moved it all into the sparse cache.

This is fine for the barriers; the only other internal user is a unit test, which we I removed. There's no good way to hook into the new simplified barrier code, and the tests of the sparse cache will still test the overall correctness. In order to run userspace uds tests, however, we still need to have a userspace implementation of the barrier concept.

For semaphores it's a different story. A number of our tests use semaphores for their own synchronization so we need to keep an implementation around. Fortunately due to the code simplification, upstream no longer uses our old methods at all, so we can simply hide it all behind #ifdefs. (We also have some tests that test that semaphores work the way we expect on our test machines. I ultimately decided to keep those since we need would still need the test implementation of semaphores either way.)


slegendr (approved):
Looks good.

PR #89, opened 2024-02-14
[VDO-5689] Ingest upstream assertion fix

slegendr:
Ingest the upstream patch from Harshit Mogalapalli to fix an assertion in volume-index.c.


lorelei-sakai (approved):
Looks good.

PR #92, opened 2024-02-17
[VDO-5684][VDO-5685][VDO-5686] Restructure and simplify various thread and locking components

lorelei-sakai:
This series restructures several components of uds-threads (renamed to thread-utils).

Commits 1-4 rename uds-threads.[ch] and uds.h to thread-utils.[ch] and indexer.h, respectively.
Commits 5-8 simplify functions using struct cond_var and move them into the indexer. [VDO-5685]
Commits 9-10 replace functions using struct mutex with kernel functions. [VDO-5684]
Commits 11-17 do low-level cleanup in threads.[ch] and rename  functions to have a vdo_ prefix instead of a uds_ one. [VDO-5686]
Finally, commit 18 adjusts the patch translation script to account for deleted and renamed files.

The first four commits are vaguely associated with VDO-5667, so they don't have a separate ticket. They represent two patches from Mike:
dm vdo: rename uds-threads.[ch] to thread-utils.[ch]
dm vdo indexer: rename uds.h to indexer.h

Patch 5 represents only part of the patch matching its name, as the rest was committed as part of my previous semaphore series.

Several of these patches are kind of long, but those mostly involve mechanical renaming so hopefully it is not too onerous to review. They get shorter near the end, too.


slegendr (changes requested):

src/c++/uds/src/uds/funnel-queue.c
```diff
@@ -8,7 +8,6 @@
 #include "cpu.h"
 #include "memory-alloc.h"
 #include "permassert.h"
-#include "uds.h"
```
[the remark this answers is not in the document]
lorelei-sakai: I didn't comment this separately, but funnel-queue.c doesn't actually need this include.

src/c++/uds/src/uds/thread-utils.h
```diff
@@ -8,20 +8,19 @@
...
-#include <linux/mutex.h>
 #include <pthread.h>
 #include <sched.h>
 #include <semaphore.h>
 #include <stdbool.h>
-#endif
+#endif /* __KERNEL__ */
+#if !(defined __KERNEL__) || defined(TEST_INTERNAL)
 
 #include "errors.h"
```
slegendr, line 23: The original patch removed the include of errors.h. So does this need to be within an ifdef?

slegendr:
The comment above is the only remaining one. Otherwise, looks good to me.

src/c++/uds/kernelLinux/tests/RequestQueue_p1.c
```diff
@@ -484,14 +484,14 @@ static void testQuadTime(const char *label,
     td[i].sections  = 4;
     td[i].producer  = producer;
     td[i].qb        = qb;
-    UDS_ASSERT_SUCCESS(uds_create_thread(producerThread, &td[i], "producer",
+    UDS_ASSERT_SUCCESS(vdo_create_thread(producerThread, &td[i], "producer",
```
bjohnsto, line 487: Are we gonna need to change UDS_ASSERT_SUCCESS to VDO_ASSERT_SUCCESS eventually?

src/c++/uds/src/uds/thread-utils.h
lorelei-sakai: It is in an #if clause; I moved the #if line above the errors.h include. This is need fomr the uds_initialize_semaphore() function among others, but I could potentially get away with making it just TEST_INTERNAL.

src/c++/uds/kernelLinux/tests/RequestQueue_p1.c
lorelei-sakai: This fuction is test-only, so there's no need to do that. We could do it if you think it's useful,though.

bjohnsto (approved):
YAY

slegendr (approved):
Looks good.

PR #98, opened 2024-02-21
[VDO-5690] Clean up scnprintf usage

C2Redhat:
[VDO-5690] Ignore scnprintf return status since it is not necessary.
Change write_* functions type from int to void since we no longer
return any result. Also, clean up any codes that check or use any
scnprintf return results.
check uds_allocate return code which was previous ignored, return
and log error when uds_allocate failed.

Note:
message-stats.c used to be generated and we added the file to our 
tree, VDOSTORY-283.
And snprintf was replaced by scnprinf because snprintf could return
-1 and it can cause problem, [vdo-4907]. 

Reported-by: Dan Carpenter <dan.carpenter@linaro.org>
Signed-off-by: Chung Chung <cchung@redhat.com>

### DIFF
--- a/src/c++/vdo/base/message-stats.c
+++ b/src/c++/vdo/base/message-stats.c
@@ -11,749 +11,407 @@
 #include "thread-device.h"
 #include "vdo.h"
 
-static int write_u64(char *prefix, u64 value, char *suffix, char **buf,
-		     unsigned int *maxlen)
-{
-	int count = scnprintf(*buf, *maxlen, "%s%llu%s", prefix == NULL ? "" : prefix,
-			      value, suffix == NULL ? "" : suffix);
-	*buf += count;
-	*maxlen -= count;
-	if (count >= *maxlen)
-		return VDO_UNEXPECTED_EOF;
-	return VDO_SUCCESS;
-}
-
-static int write_u32(char *prefix, u32 value, char *suffix, char **buf,
-		     unsigned int *maxlen)
-{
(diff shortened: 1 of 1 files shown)


src/c++/vdo/base/message-stats.c
```diff
@@ -11,749 +11,407 @@
...
-	int result;
-
-	result = write_string(prefix, "{ ", NULL, buf, maxlen);
-	if (result != VDO_SUCCESS)
-		return result;
+	write_string(prefix, "{ ", NULL, buf, maxlen);
 	/* Number of compressed data items written since startup */
-	result = write_u64("compressedFragmentsWritten : ",
+	write_u64("compressedFragmentsWritten : ",
 			   stats->compressed_fragments_written, ", ", buf, maxlen);
```
raeburn, line 132: Indentation of the continuation line needs adjusting

raeburn (changes requested):

raeburn:
I don't think the "note" part in the description is needed. That's all visible in our git history, and not terribly important for this change.

src/c++/vdo/base/message-stats.c
C2Redhat: I have made the changes.

C2Redhat:
> I don't think the "note" part in the description is needed. That's all visible in our git history, and not terribly important for this change.

Matt ask me to add the info in description, but it is not in the commit, so it only appear in the PR.

raeburn (approved):

PR #100, opened 2024-02-21
[VDO-5667] Move UDS files to indexer sub-directory

lorelei-sakai:
Move the core UDS files into a sub-directory upstream. Since the files do not move locally, that means adjusting the spelling of several includes lines, and rewriting some Makefiles to adapt to the new file structure.

The biggest change in the infrastructure is that the packaging/github directory now defines the VDO_UPSTREAM flag and uses the upstream file structure. This allows us to test the upstream build arrangement a little more directly, without having to push changes to the dm-linux repository. It also means that if we do continue to push changes to our own open-source repositories (such as dm-vdo/kvdo), we will reflect the new directory structure there as well as in the kernel.

### DIFF
--- a/src/packaging/dmlinux/Makefile
+++ b/src/packaging/dmlinux/Makefile
@@ -36,10 +36,12 @@ prepare: prepare.out
 prepare.out: FORCE
 	$(if $(DMLINUX_SOURCE_DIR),,$(error Must set DMLINUX_SOURCE_DIR))
 	rm -rf $(WORK_DIR)
-	mkdir -p $(WORK_DIR)/$(PACKAGE)/dm-vdo
+	mkdir -p $(WORK_DIR)/$(PACKAGE)
 	sed -e "s/@VERSION@/$(VDO_VERSION)/g" kvdo.spec >$(WORK_DIR)/kvdo.spec
 	cp kernel/Makefile $(WORK_DIR)/$(PACKAGE)
-	cp $(DMLINUX_SOURCE_DIR)/drivers/md/dm-vdo/*.[ch] $(WORK_DIR)/$(PACKAGE)/dm-vdo > prepare.out
+	cd $(DMLINUX_SOURCE_DIR)/drivers/md && 		\
+	find dm-vdo -type f -name '*.[ch]'		\
+	-exec cp --parents {} $(WORK_DIR)/$(PACKAGE) ';' > prepare.out
 
 $(SOURCES): prepare
 	mkdir -p $(SOURCES)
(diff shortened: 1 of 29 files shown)


src/packaging/dmlinux/Makefile
```diff
@@ -36,10 +36,10 @@ prepare: prepare.out
 prepare.out: FORCE
 	$(if $(DMLINUX_SOURCE_DIR),,$(error Must set DMLINUX_SOURCE_DIR))
 	rm -rf $(WORK_DIR)
-	mkdir -p $(WORK_DIR)/$(PACKAGE)/dm-vdo
+	mkdir -p $(WORK_DIR)/$(PACKAGE)
 	sed -e "s/@VERSION@/$(VDO_VERSION)/g" kvdo.spec >$(WORK_DIR)/kvdo.spec
 	cp kernel/Makefile $(WORK_DIR)/$(PACKAGE)
-	cp $(DMLINUX_SOURCE_DIR)/drivers/md/dm-vdo/*.[ch] $(WORK_DIR)/$(PACKAGE)/dm-vdo > prepare.out
+	cp -r $(DMLINUX_SOURCE_DIR)/drivers/md/dm-vdo $(WORK_DIR)/$(PACKAGE)/dm-vdo > prepare.out
```
bjohnsto, line 42: Is there a reason we're not just copying source files anymore? I'm worried about copying object files and things if the kernel source happened to have been built before running this target.

lorelei-sakai: It's a newly-cloned repo when CI runs the tests, right? 

I just thought this was easier that having to copy each level of the directory tree explicitly. Perhaps there's a way to just copy the .[ch] files, though.

lorelei-sakai: I've replaced the cp command with a slightly more sophisticated find command that copies just the *.c and *.h files in the whole directory structure.

src/packaging/kpatch/Makefile.upstream
```diff
@@ -1,50 +1,39 @@
 # SPDX-License-Identifier: GPL-2.0-only
 
+ccflags-$(CONFIG_DM_VDO) := -I${src} -I${src}/indexer
```
raeburn, line 3: I think this should be ccflags-y, based on reading Documentation/kbuild/makefiles.rst .

The ccflags-$(CONFIG_FOO) style works for optional flags wanted depending on some additional boolean y/n flag like CONFIG_ACPI_DEBUG. But CONFIG_DM_VDO is tristate and can be y/m/n.

raeburn (changes requested):

bjohnsto (approved):
Change looks good to me

src/packaging/kpatch/Makefile.upstream
lorelei-sakai: *sigh* Fixed. I wish the documentation were a little clearer about that.

raeburn (approved):

PR #105, opened 2024-02-28
[VDO-5697] Standardize on VDO_SUCCESS in utility code

lorelei-sakai:
This series ingests eight of Mike's suggested patches, all around the theme of using VDO_SUCCESS instead of UDS_SUCCESS in the utility files. I have added some internal-only commits, propagating Mike's change to internal code, and I have also added two new commits, one centered of removing unused error codes and making VDO_SUCCESS available to UDS code, and one centered on completing the VDO_ASSERT transition that Mike only did for VDO files but not for indexer files.

This series is built on top of, and depends on, the commits in PR 104.


src/c++/uds/src/uds/errors.h
```diff
@@ -21,7 +21,8 @@
 /* Valid status codes for internal UDS functions. */
 enum uds_status_codes {
 	/* Successful return */
-	UDS_SUCCESS = 0,
+	VDO_SUCCESS = 0,
```
bjohnsto, line 24: This feels weird to me. In my mind, I was assuming there was a common include that defined VDO_SUCCESS = 0 and then we would include that everywhere it was needed, including the uds ones. I know that sort of goes against how we have uds feed into vdo currently, but is that something we are looking at changing down the road?

src/c++/vdo/tests/DmKcopydFake_t1.c
```diff
@@ -86,7 +86,7 @@ static void testDmKcopyd(block_count_t regionSize)
 
   // Verify that the original data has not been touched.
   char *buffer;
-  UDS_ASSERT_SUCCESS(vdo_allocate(VDO_BLOCK_SIZE * regionSize, char,
+  VDO_ASSERT_SUCCESS(vdo_allocate(VDO_BLOCK_SIZE * regionSize, char,
                                   "verification buffer", &buffer));
   VDO_ASSERT_SUCCESS(layer->reader(layer, 0, regionSize, buffer));
   UDS_ASSERT_EQUAL_BYTES(buffer, data, VDO_BLOCK_SIZE * regionSize);
```
bjohnsto, line 92: Are we planning to convert these types of asserts as well?

src/c++/uds/userLinux/tests/Permassert_t1.c
```diff
@@ -99,15 +99,15 @@ static void checkNotFound(const char *str)
 /**********************************************************************/
 static void testAssertionSuccess(void)
 {
-  UDS_ASSERT_SUCCESS(ASSERT(true, "true"));
+  UDS_ASSERT_SUCCESS(VDO_ASSERT(true, "true"));
```
bjohnsto, line 102: I think this should be VDO_ASSERT_SUCCESS? Or are we only converting tests to just use VDO_ASSERT?

bjohnsto (commented):
Looks good but I did have a couple questions / comments

src/c++/uds/kernelLinux/uds/funnel-requestqueue.c
```diff
@@ -275,5 +275,5 @@ void uds_request_queue_finish(struct uds_request_queue *queue)
 
 	uds_free_funnel_queue(queue->main_queue);
```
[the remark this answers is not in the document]
lorelei-sakai: Yes, uds_free_funnel_queue is an example of a function that Mike converted that should not be converted in this commit. This particular method will get converted, but in a later commit that deals with all the funnel_queue functions.

src/c++/vdo/tests/DmKcopydFake_t1.c
lorelei-sakai: Which type of assert are you talking about here?

src/c++/uds/userLinux/tests/Permassert_t1.c
lorelei-sakai: This test is in the uds directory, so it doesn't have VDO_ASSERT_SUCCESS defined. As I mentioned in the commit message, I didn't think that was worth doing.

src/c++/vdo/tests/DmKcopydFake_t1.c
bjohnsto: UDS_ASSERT_EQUAL_BYTES

src/c++/uds/src/uds/errors.h
bjohnsto: I wasn't suggesting a single file. I assumed we might have a very common file somewhere. If thats errors.h, then I might consider moving VDO_SUCCESS = 0 out of the uds_status_codes enum and put it at the top of the file.

lorelei-sakai: Hm, could do.

bjohnsto: very

bjohnsto (approved):
Looks good

PR #109, opened 2024-02-29
[VDO-5695] dm vdo funnel-queue: change from uds_ to vdo_ namespace 

C2Redhat:
This PR contains two commits.  The first one is to apply Mike Snitzer's uds to vdo namespace change
in the kernel side.  The second commit is to adapt those changes in our test code.


lorelei-sakai (approved):
All looks good.

PR #113, opened 2024-03-01
[VDO-5635] Add vdo-setup-dmtest package

slegendr:
The purpose of the vdo-setup-dmtest package is to set up the repositories and environment necessary
for developing or running VDO tests through the dmtest-python utility. It will install and configure dmtest-python, along with its blk-archive and bufio-test dependencies, as they may be used for future VDO testing and/or test development. 

More information can be found in the README.md file included in this change. The config.yaml is the default configuration, and will likely need to be modified as we know more about how we intend to add a vdo target to dmtest-python.


src/python/vdo-setup-dmtest/config.yaml
```diff
@@ -0,0 +1,28 @@
+# Devices to be used by dmtest-python
+DATA_DEVICE: /dev/mapper/scratch-vdo_scratch
+METADATA_DEVICE: /dev/mapper/scratch-vdo_scratch
+
+# Dmtest-python and dependency repos
+# May be defined with URL or full path. If URL, automatic cloning will be performed.
+DMTEST_PYTHON_REPO: https://github.com/dm-vdo/dmtest-python
```
rhawalsh, line 7: Should this point at our fork, or the parent?  I'd assume we would want to track with the current version that everyone else is using, right?

src/python/vdo-setup-dmtest/source/Logger.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python3
+"""
+  Logger - Sets up the vdo_setup_dmtest logger
+
+  Copyright (c) 2023 Red Hat
```
rhawalsh, line 5: 2024?  (or no date at all)

src/python/vdo-setup-dmtest/source/Utils.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env python3
+"""
+  Utils - General vdo_setup_dmtest constants and utilities
+
+  Copyright (c) 2023 Red Hat
```
rhawalsh, line 5: 2024? or no date..

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
```diff
@@ -0,0 +1,512 @@
+#! /usr/bin/env python3
+"""
+  Installs and sets up a dmtest-python environment for running vdo tests.
+
+  Copyright (c) 2023 Red Hat
```
rhawalsh, line 5: 2024?  Or no date at all...

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
```diff
@@ -0,0 +1,512 @@
...
+    self.verifyLinuxSource()
+
+  def parseArgs(self, args):
+    """
+    Parse any input arguments and apply defaults
+    """
+    parser = argparse.ArgumentParser(
+      prog='vdo-setup-dmtest',
+      description='Installs and sets up a dmtest-python environment for running vdo tests.',
+      epilog='For additional information see README.md within the module directory.')
```
rhawalsh, line 79: You're using double quotes everywhere else in this def.  Perhaps you can change these to doubles to be consistent with the rest of this def and the script?

src/python/vdo-setup-dmtest/source/Logger.py
```diff
@@ -0,0 +1,74 @@
...
+            logging.INFO: self.fmt,
+            logging.WARNING: self.yellow + self.fmt + self.reset,
+            logging.ERROR: self.red + self.fmt + self.reset,
+            logging.CRITICAL: self.bold_red + self.fmt + self.reset
+        }
+
+    def format(self, record):
+        log_fmt = self.FORMATS.get(record.levelno)
+        formatter = logging.Formatter(log_fmt)
+        return formatter.format(record)
```
rhawalsh, line 39: This appears to be using tab (or 4-space) indents.  The rest of the script(s) appear to be using 2-space indents.  Please adjust.

src/python/vdo-setup-dmtest/source/Utils.py
```diff
@@ -0,0 +1,125 @@
...
+        [ req_distinguished_name ]
+        CN = Build time autogenerated kernel key for test modules
+        [ myexts ]
+        basicConstraints=critical,CA:FALSE
+        keyUsage=digitalSignature
+        subjectKeyIdentifier=hash
+        authorityKeyIdentifier=keyid
+        """))
+
+    result = runCommand(['sudo cp', genkeyFile, certsDir], logger)
```
rhawalsh, line 50: This command format doesn't match the rest of the script where you're passing things to runCommand all as a single element.

Additionally, 'sudo cp'.. shouldn't that be `'sudo', 'cp'` in this context?

Personally I'd just suggest making this match the rest of the script, but if you decide to convert everything to use this method I won't complain either.

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
```diff
@@ -0,0 +1,512 @@
...
+
+    with open(configFile, 'r') as file:
+      loadedConfig = yaml.safe_load(file)
+
+    if not bool(loadedConfig):
+      self.log.error("Unable to load config file ({0})".format(configFile))
+      return False
+
+    for key in loadedConfig.keys():
+      self.config[key] = loadedConfig[key]
```
rhawalsh, line 131: Why not just use something like `self.config = loadedConfig`?  It doesn't appear to me that you're using self.config anywhere else before this.

src/python/vdo-setup-dmtest/config.yaml
```diff
@@ -0,0 +1,28 @@
+# Devices to be used by dmtest-python
+DATA_DEVICE: /dev/mapper/scratch-vdo_scratch
+METADATA_DEVICE: /dev/mapper/scratch-vdo_scratch
```
rhawalsh, line 3: Are these OK to be the same device?

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
```diff
@@ -0,0 +1,512 @@
...
+          msg = 'does not match any of the expected substrings (BUFIO | BLK_ARCHIVE | DMTEST)'
+          self.log.warning("Repo key '{key}' {error} - Skipping".format(key = key, error = msg))
+      except Exception as error:
+        self.log.error("Failed during {name} installation: {reason}".format(name = repoName,
+                                                                            reason = error))
+        # If dmtest-python install failed, do not continue any further
+        if (key.upper().find('DMTEST') != -1):
+          sys.exit(1)
+        else:
+          self.log.warning("Manual install of {name} required".format(name = repoName))
```
rhawalsh, line 287: Is it worth noting the location that it should be installed to?

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
```diff
@@ -0,0 +1,512 @@
...
+
+    # Determine the rhel release branch to use for installation
+    if (os.path.exists('/etc/fedora-release')):
+      osVersion = 9
+    elif (os.path.exists('/etc/redhat-release')):
+      result = runCommand('cut -d " " -f6 /etc/redhat-release | cut -d"." -f1')
+      osVersion = result.stdout.rstrip()
+    elif (os.path.exists('/etc/centos-release')):
+      result = runCommand('cut -d " " -f4 /etc/centos-release')
+      osVersion = result.stdout.rstrip()
```
rhawalsh, line 320: What if I'm running this on Ubuntu or some other distro?  Perhaps there should be a block that catches that.
```
else:
  raise OSError "Unsupported OS in use"
```
... or something

src/python/vdo-setup-dmtest/source/Utils.py
```diff
@@ -0,0 +1,125 @@
...
+  valid = False
+
+  # Evaluate if the certificate will expire in the next day
+  try:
+    result = runCommandIgnoringErrors(['sudo openssl x509 -checkend', daySeconds, '-noout -in',
+                                       certPath])
+    if ('will not expire' in result.stdout):
+      valid = True
+    elif ('will expire' in result.stdout):
+      logger.info("Certificate will expire within 24 hours - recreation needed")
```
rhawalsh, line 121: Is there any significance to 24 hours?  How long is reasonable for dmtest to actually run?  Can it exceed 24 hours?

Perhaps `recreation needed` can be rephrase?  I like to play Dodgeball, but I don't think it'll help here ;)

Maybe `please re-generate certificate` ??

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
```diff
@@ -0,0 +1,512 @@
...
+      genkeyPath = os.path.join(certsDir, 'x509.genkey')
+      if not (os.path.exists(genkeyPath)):
+        assert createX509Genkey(certsDir) == True, 'Unable to create x509.genkey'
+
+      self.log.info("Creating the certificate/key pair needed for signing the module")
+      keyPath = os.path.join(certsDir, 'signing_key.x509')
+      opensslCmd = ['sudo openssl req -new -nodes -utf8 -sha512 -days 36500 -batch -x509',
+                    '-config', genkeyPath, '-outform PEM -out', certPath, '-keyout', keyPath]
+      result = runCommand(opensslCmd)
+      assert result.returncode == 0, "Creating OpenSSL certificate failed"
```
rhawalsh, line 351: Since the other cert operations are in their own functions, perhaps this is also a candidate for moving into its own function?

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
```diff
@@ -0,0 +1,512 @@
...
+    if isExecutable(localExePath) and isExecutable(cargoExePath):
+      self.log.debug("{0} is already installed - skipping".format(repoName))
+      if userCargoPath not in os.environ["PATH"]:
+        self.log.warning("PATH must include '{path}' for {repo} to function properly".format(
+          path = userCargoPath, repo = repoName))
+
+        # Append strings to necessary lists for reporting downstream
+        self.envVars.append('PATH=$PATH' + os.pathsep + userCargoPath)
+        self.dmtestHealthWarn.append('Dependency blk-archive will display as missing until PATH' \
+                                     + ' is updated')
```
rhawalsh, line 377: Perhaps these strings can be created without + concatenation?

```
self.envVars.append('PATH=$PATH{pathsep}{userCargoPath}'.format(pathsep=os.pathsep, userCargoPath=userCargoPath))
self.dmtestHealthWarn.append('Dependency blk-archive shows as missing until PATH is updated')
```

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
```diff
@@ -0,0 +1,512 @@
...
+                'cargo install --path $PWD']
+    for cmd in commands:
+      result = runCommand(cmd)
+      assert result.returncode == 0, 'Non-zero return status received'
+    self.log.warning("PATH must be updated to include '{0}'".format(userCargoPath))
+
+    # Append strings to necessary lists for reporting downstream
+    self.envVars.append('PATH=$PATH' + os.pathsep + userCargoPath)
+    self.dmtestHealthWarn.append('Dependency blk-archive will display as missing until PATH' \
+                                 + ' is updated')
```
rhawalsh, line 391: Perhaps these strings can be created without + concatenation?

```
self.envVars.append('PATH=$PATH{pathsep}{userCargoPath}'.format(pathsep=os.pathsep, userCargoPath=userCargoPath))
self.dmtestHealthWarn.append('Dependency blk-archive shows as missing until PATH is updated')
```

src/python/vdo-setup-dmtest/config.yaml
slegendr: My understanding from discussions with Matt is that our primary intention is to use this internally. As such, it should point to our fork so that our team developers are set up to add tests to dmtest-python that we can later push upstream to the parent.

src/python/vdo-setup-dmtest/source/Utils.py
slegendr: I don't understand. There are instances in vdo-setup-dmtest.py where runCommand is presented with a list and not a single element. I wrote runCommand so that it could handle either string or list input in order to make it adaptable.

For this context, it could be written as you suggest, or as it is. Within runCommand(), it will evaluate if the input is a string or a list and concatenate as appropriate.

src/python/vdo-setup-dmtest/config.yaml
slegendr: We do not know yet. That is something I am trying to figure out.

src/python/vdo-setup-dmtest/source/Utils.py
slegendr: There is no significance to 24 hours beyond giving it a date to compare to for determining certificate validity.

As for the "recreation needed", that can be rephrased, but we regenerate the certificate from within vdo-setup-dmtest in that case, so "please re-generate certificate" wouldn't be appropriate. Maybe "Certificate will expire within 24 hours - regenerating".

src/python/vdo-setup-dmtest/source/Utils.py
slegendr: Fixed in latest push.

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
slegendr: Fixed in latest push.

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
slegendr: Fixed in latest push.

src/python/vdo-setup-dmtest/source/Logger.py
slegendr: Fixed in latest push.

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
slegendr: Fixed in latest push.

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
slegendr: Fixed in latest push.

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
slegendr: Fixed in latest push.

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
slegendr: Fixed in latest push.

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
slegendr: Fixed in latest push.

src/python/vdo-setup-dmtest/config.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+# Dmtest-python and dependency repos
+# May be defined with URL or full path. If URL, automatic cloning will be performed.
+DMTEST_PYTHON_REPO: https://github.com/dm-vdo/dmtest-python
+BUFIO_TEST_REPO: https://github.com/bmarzins/bufio-test.git
+BLK_ARCHIVE_REPO: https://github.com/jthornber/blk-archive.git
```
bjohnsto, line 9: why do we need blk_archive? What do we plan to use it for?

slegendr: Strictly speaking, we do not need bufio-test or blk-archive. I was asked to include them in the script so we could more easily explore if we want to use them in VDO tests we add to dmtest-python.

bjohnsto: This seems weird to me. Not sure why we would want to use them in our tests; especially blk-archive, which is an attempt to create a compress / dedupe type device as well. Seems like just excess time / space to d\l this stuff. But I'll not block if someone specifically asked for it.

src/python/vdo-setup-dmtest/vdo-setup-dmtest.py
slegendr: Fixed in latest push.

src/python/vdo-setup-dmtest/source/Utils.py
slegendr: Updated the message in the latest push to note that it's regenerating the certificate when it detects it will expire within a day.

slegendr:
Closing PR to move to dm-vdo/dmtest-python.

PR #114, opened 2024-03-02
[VDO-5700][VDO-5619][VDO-5709] Adopt various minor fixes

lorelei-sakai:
[VDO-5700] The first commit removes vdo_perform_once() from the kernel build. It is not necessary because we can initialize what we need at module load time. I left vdo_perform_once() in for userspace because the test runners and userspace utilities don't have a clear initialization bottleneck and I didn't want to worry about accidentally breaking something subtle.

[VDO-5619] The second commit removes the CURRENT_VERSION constant from upstream. Upstream the only version we care about is the kernel version, so a separate field in meaningless at best and confusing at worst. However, we do still want an identifiable version for internal tests and for RHEL8/9 releases, so this is only removed for VDO_UPSTREAM.

[VDO-5709] This is a tiny format fix that came in to the mailing list this morning.

These really should be three separate PRs, but I'm trying to save time because the merge window is so near.

### DIFF
--- a/src/c++/uds/kernelLinux/uds/thread-utils.c
+++ b/src/c++/uds/kernelLinux/uds/thread-utils.c
@@ -20,7 +20,6 @@
 
 static struct hlist_head thread_list;
 static struct mutex thread_mutex;
-static atomic_t thread_once = ATOMIC_INIT(0);
 
 struct thread {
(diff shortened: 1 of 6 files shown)


slegendr (approved):
Looks good to me.

PR #116, opened 2024-03-12
dm-vdo: Add code to adapt kernel next block device changes

C2Redhat:
In linux next kernel, blkdev_get_by_dev has been redefined and blkdev_put is obsolete.  
Update getTestIndexNames.c to adapt the kernel changes by adding VDO_USE_ALTERNATE_3 
and rearranging VDO_USE_ALTERNATE marco.


src/c++/uds/kernelLinux/tests/getTestIndexNames.c
```diff
@@ -74,20 +87,25 @@ struct block_device **getTestMultiBlockDevices(void)
 /**********************************************************************/
 void putTestBlockDevice(struct block_device *bdev)
 {
-#ifndef VDO_USE_ALTERNATE
+
```
lorelei-sakai, line 90: Please remove this line. We do not need a blank line at the beginning of a function.

src/c++/uds/kernelLinux/tests/getTestIndexNames.c
```diff
@@ -30,21 +36,28 @@ enum { BLK_FMODE = FMODE_READ | FMODE_WRITE };
...
 	struct bdev_handle *bdev_handle;
 
 	bdev_handle = bdev_open_by_path(name, BLK_FMODE, NULL, &hops);
 	bdev = bdev_handle->bdev;
+#endif /* VDO_USE_ALTERNATE_2 */
+#else
+	const struct blk_holder_ops hops = { NULL };
+
+	bdev = file_bdev(bdev_file_open_by_path(name, BLK_FMODE, NULL, &hops));
 #endif /* VDO_USE_ALTERNATE */
```
lorelei-sakai, line 61: Looking at the way cleanup happens, I think we need to do something a little subtler here.
I think after we call file_bdev(), we should call get_device(bdev) to get a reference in the block device, then close the struct file object immediately.

We will still have a reference to the bdev (because of get_device()), but not to the file object.  Then we can release the device reference later with put_device().

lorelei-sakai (changes requested):
Generally looks good. Thinking abotu this some more, though, I think it's worth being a bit more diligent about the file handling.

src/c++/uds/kernelLinux/tests/getTestIndexNames.c
```diff
@@ -10,16 +10,25 @@
 
 #undef VDO_USE_ALTERNATE
 #undef VDO_USE_ALTERNATE_2
-#ifdef RHEL_RELEASE_CODE
+#undef VDO_USE_ALTERNATE_3
+
+/*
+  COMMENT HERE
+ */
```
lorelei-sakai, line 17: This comment says COMMENT HERE. It should probably be a real comment.

src/c++/uds/kernelLinux/tests/getTestIndexNames.c
```diff
@@ -10,16 +10,25 @@
...
+#if defined(RHEL_RELEASE_CODE)
 #define VDO_USE_ALTERNATE
```
lorelei-sakai, line 19: What about the code to have the fake RHEL_RELEASE_CODE 9.99? Did it turn out that you don't need it?

lorelei-sakai (changes requested):
This version fixes the previsou issues but it looks to be a little incomplete still.

lorelei-sakai (approved):
Looks good. We'll have to rework this with some changes coming upstream, but they should simplify things.

PR #118, opened 2024-03-18
Clean up properly after module build failure. [VDO-5603]

raeburn:
The previous implementation for logging an error in postinstall commands would treat the "rpm -i" as failing, and not remove the binary package, although RPM considered the installation successful enough to leave the package there. This caused checkServer.pl failures.

Replace the "checker" approach with an additional step run after "rpm -i", checking the DKMS output in the new step. If an error is detected, the previous "successful" steps are cleaned up, including first removing the package just installed.

The assertion call logs the matching string, which includes the full text of the build log, when failing.


bjohnsto (approved):
Looks good

PR #119, opened 2024-03-20
dm vdo: use kernel byteswapping routines instead of GCC ones [VDO-5720]

raeburn:
The murmurhash3.c code was imported from a user-mode library that explicitly tested byte order macros and called __builtin functions that aren’t well supported in the kernel on all architectures.

Reported-by: Guenter Roeck <linux@roeck-us.net>

### DIFF
--- a/src/c++/uds/src/uds/murmurhash3.c
+++ b/src/c++/uds/src/uds/murmurhash3.c
@@ -8,6 +8,10 @@
 
 #include "murmurhash3.h"
 
+#ifdef __KERNEL__
+#include <asm/byteorder.h>
+
+#endif
 static inline u64 rotl64(u64 x, s8 r)
 {
 	return (x << r) | (x >> (64 - r));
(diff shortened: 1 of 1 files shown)


src/c++/uds/src/uds/murmurhash3.c
```diff
@@ -8,6 +8,10 @@
 
 #include "murmurhash3.h"
 
+#ifdef __KERNEL__
+#include <asm/byteorder.h>
+#endif
+
 static inline u64 rotl64(u64 x, s8 r)
```
lorelei-sakai, line 15: Don't add an extra blank line here for userspace.

lorelei-sakai (changes requested):
The code change looks basically correct. A few comments on the chnage description.
The title line should not include capitalized letters or punctuation. (It's not a sentence.)
Also I'd prefer if the rest of the message were phrased to not mention userspace. I'm not sure the rest of the commit mesage beyond the title is even needed; any reader of the patch can see what you're doing. Maybe mention that the __builtin support is not universal  (by way of motivation) and nothing else.

lorelei-sakai (approved):
Looks good.

PR #120, opened 2024-03-20
[VDO-5643] Fix reboot issues with lvm by always uses use_devicesfile=0

bjohnsto:
Move the use of use_devicesfile=0 to always be used. It seems like when we are now under a raid, we can't properly reactivate lvm volumes after a reboot. Eventually, we should start using the devices file itself by adding and removing our devices during a test, but it requires a bit of refactoring with ISCSI.pm and I think we can delay that work for now.


lorelei-sakai (approved):
Seems reasonable.

PR #121, opened 2024-03-21
Murmurhash byteswapping fix v2 [VDO-5720]

raeburn:
Per suggestion from Eric Biggers on dm-devel, don't use cpu_to_le64 and friends, use get_unaligned_le64 and friends. Call them directly rather than via a wrapper function, since we don't need the conditional mess. Use the same code for user mode, as well.


lorelei-sakai (approved):
It seems the ROTL64 macro doesn't serve much purpose. But that's unrelated. This change looks good.

PR #125, opened 2024-03-29
dm-vdo: Changes are needed to build and test RAWHIDE

C2Redhat:
[VDO-5715] RAWHIDE updated some of the system's .h files. We need to update our .h files to reflect thoes changes. Also, removed some of the obsolete definitions that are no longer needed but causes problems.

Also, we discover that missing-declarations and missing-prototypes were not turned on by default in gcc before 14.0.1.  Some of our codes need to be updated since it is now turned on.

### DIFF
--- a/src/c++/vdo/fake/linux/bio.h
+++ b/src/c++/vdo/fake/linux/bio.h
@@ -11,8 +11,11 @@
 #define __LINUX_BIO_H
 
 /* struct bio, bio_vec and BIO_* flags are defined in blk_types.h */
+#include <asm/unaligned.h>
 #include <linux/blk_types.h>
+#include <linux/kernel.h>
 #include <linux/limits.h>
+#include <linux/minmax.h>
 #include <linux/version.h>
 
 #define BIO_DEBUG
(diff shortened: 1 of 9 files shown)


src/c++/vdo/fake/linux/bvec.h
```diff
@@ -11,6 +11,7 @@
 #define __LINUX_BVEC_H
 
 #include <linux/highmem.h>
+#include <linux/min_heap.h>
 #include <linux/minmax.h>
```
lorelei-sakai, line 15: Do you actually need min_heap.h here? I thought we worked around that.

src/c++/vdo/fake/linux/bio.h
```diff
@@ -11,8 +11,12 @@
 #define __LINUX_BIO_H
 
 /* struct bio, bio_vec and BIO_* flags are defined in blk_types.h */
+#include <asm/unaligned.h>
 #include <linux/blk_types.h>
+#include <linux/kernel.h>
 #include <linux/limits.h>
+#include <linux/min_heap.h>
+#include <linux/minmax.h>
 #include <linux/version.h>
```
lorelei-sakai, line 21: Do we actually need all these new includes? I hope we don't need min-heap.h, at least.

lorelei-sakai (changes requested):
I think you may be added more header includes than are strictly necessary.

Also, in the commit messages, we should not be in the habit of including ticket references. Please restrict those to the PR title and description.

src/c++/vdo/fake/linux/bvec.h
C2Redhat: WARN_ONCE is defined in min_heap.h so

lorelei-sakai (approved):
Looks good now.

PR #128, opened 2024-04-03
dm-vdo: Add extra tests to verify if a system is RHEL.

C2Redhat:
Start from Fedora 3.7, Fedora system has RHEL_RELEASE_CODE defined and RHEL_MINOR macro set to 99 in
/usr/src/kernels/`uname -r`/include/generated/uapi/linux/version.h which can cause problems.
Add additional tests to make sure we have the correct VDO_USE_ALTERNATE define.


lorelei-sakai (changes requested):
The change itself looks correct. However, if we do this I think we should do it everywhere we check for RHEL_RELEASE_CODE, or else tests might not pass (even if the code compiles). Examples include uds/kernelLinux/uds/thread-utils.c, vdo/base/data-vio.c, and c++/devices/common.h. Grep for RHEL_RELEASE_CODE to be sure you update all the places we use it.

There are also two typos in the commit message: s/Occasionaly/Occasionally/ and s/marco/macro/

lorelei-sakai (approved):
Thanks for adding the extra handling. Looks good.

bjohnsto (approved):
I would prefer to use something besides minor as our distinguisher between Fedora and RHEL (maybe RHEL_RELEASE) but I also would like us to understand why Fedora started outputting stuff that makes us think its RHEL (RHEL_RELEASE_VERSION, etc). Still this change will get us compiling, so I say YAY.

PR #131, opened 2024-04-17
[VDO-5743] dm vdo test: manage extra block-related handles correctly

lorelei-sakai:
Recent changes to the block interface mean that our test code can't access the struct block_device directly. Instead, tests obtain another intermediate structure (struct bdev_handle or struct file) first. These intermediate structures must have the same lifetime as the struct block_device itself. This is handled by creating a new block_device_context structure to hold any additional information we need to track. 

The existing implementation of the 6.7.0 - 6.9.0 case did not remember the struct bdev_handle obtained when opening the device, which resulted in leaking that structure, and more importantly causing bdev_release() to call kfree() on an address in the stack. (It is surprising to me that this does not fail more often than it does.)

Although this code has been used by our tests for months, it has only become noticeable recently. Perhaps a recent Fedora update has changed the handling of stack-allocated memory, or perhaps there is just stricter checking now.

### DIFF
--- a/src/c++/uds/kernelLinux/tests/getTestIndexNames.c
+++ b/src/c++/uds/kernelLinux/tests/getTestIndexNames.c
@@ -20,7 +20,7 @@
 #else /* !RHEL_RELEASE_CODE */
 #if (LINUX_VERSION_CODE < KERNEL_VERSION(6,9,0))
 #define VDO_USE_ALTERNATE
-#if (LINUX_VERSION_CODE < KERNEL_VERSION(6,8,0))
+#if (LINUX_VERSION_CODE < KERNEL_VERSION(6,7,0))
 #define VDO_USE_ALTERNATE_2
 #if (LINUX_VERSION_CODE < KERNEL_VERSION(6,5,0))
 #define VDO_USE_ALTERNATE_3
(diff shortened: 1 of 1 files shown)


src/c++/uds/kernelLinux/tests/getTestIndexNames.c
```diff
@@ -31,83 +31,125 @@
...
-	bdev = blkdev_get_by_path(name, BLK_FMODE, NULL);
+	context->block_device = blkdev_get_by_path(name, BLK_FMODE, NULL);
 #else
 	const struct blk_holder_ops hops = { NULL };
 
-	bdev = blkdev_get_by_path(name, BLK_FMODE, NULL, &hops);
+	context->block_device = blkdev_get_by_path(name, BLK_FMODE, NULL, &hops);
 #endif /* VDO_USE_ALTERNATE_3 */
 #else
 	const struct blk_holder_ops hops = { NULL };
```
raeburn, line 62: At least some flavors of the bdev open functions save the "hops" pointer away. I think the structures need to be "static const" if they're defined within the function body, or moved to a top-level definition, because they'll be on the stack otherwise.

raeburn (changes requested):

src/c++/uds/kernelLinux/tests/getTestIndexNames.c
lorelei-sakai: I don't think we care whether the hops are on the stack, because they are under the caller's control (i.e., our control). The problem with the device_handle is that the kernel controls the memory, so it tries to free it. But with the hops, that doesn't happen.

I could move the hops into the block_device_context, but I really think it can't matter.

raeburn: There are calls into the holder ops from bdev.c, I'm not convinced we fully control it all.
Since the values are constants, I don't think they need to move, they just need static storage allocation. Though, from other examples in the kernel, I think passing a null pointer may be allowed too.

lorelei-sakai: Okay, fine, I've added the static.

raeburn (approved):

PR #132, opened 2024-04-18
[VDO-5594] Use grubby for kernel option changes on RHEL 9.

raeburn:
It appears that grubby is the new way of updating kernel options, and has been preferred for a while, but in RHEL 9 there are kernel options used that are not in the grub config file, making grubby the only reliable way of updating options without losing others.

Switch only RHEL 9 to grubby, for now.


lorelei-sakai (approved):
It's a little weird to do the install twice in the non-RHEL9 modify case, but I suppose it's not an issue.

PR #133, opened 2024-04-18
[VDO-5639] Disable writable storage before failing

raeburn:
Disable writable storage before failing

Leaving the VDO pool writable seems to be invisible to LVM and so it won't clean up the pool while we try to tear down the storage stack.


lorelei-sakai (approved):
Looks good to me.

PR #134, opened 2024-04-19
[VDO-5733] dm-vdo: Update various files for VDO package upstreaming

C2Redhat:
In order to push vdo package upstream, we need to adapt various changes:

. Add bash_completions_dir variable in vdo/Makefile . Add _defaultlicensedir variable to user/Makefile . Rename adaptLVMVDO.sh to adaptlvm
. Update MANIFEST.yaml and vdo.spec files to reflect
  adaptLVMVDO.sh name change
. Update runAdaptLVMScript in DebugTools.pm to call adaptlvm


lorelei-sakai (commented):
The adaptlvm name change really sedms like it would be clearer as its own separate commit. But it's not a big deal.

i thinnk there's a  ticket for Fedora packagnig; it's probably worth tagging the PR.

PR #135, opened 2024-04-23
dm-vdo: Remove SystemTap test

C2Redhat:
Remove SystemTap tests and files that are related to it.


src/perl/vdotest/VDOTest.pm
```diff
@@ -226,48 +221,6 @@ sub new {
...
-# Launch a SystemTap script to record the kernel page cache's
-# retention of blocks in our device.
-#
-# @param devMajor  The block major device number
-# @param devMinor  The block minor device number
-##
-sub _monitorPageCacheStats {
-  my ($self, $devMajor, $devMinor) = assertNumArgs(3, @_);
-  if (!$self->{_pageCacheMonitor}) {
-    my $script = "monitor-page-cache.stp";
```
lorelei-sakai, line 260: This monitor-page-cache.stp script... does it also need to be removed? Does it still exist?

Either way, it's probably worth mentioning in the commit message.that this feature is also being removed.

lorelei-sakai (changes requested):
A tag for the Jira ticket(s) should be added to the PR description, in addition to the other comments.

src/perl/vdotest/VDOTest.pm
C2Redhat: I did a find on the src and I don't see the script.  "monitorPageCacheStats   => 0", so I think the stp file has been removed when we set that to 0. 

We have jira ticket for systemtap not working.  I am not sure if that is the right jira ticket to add.
I will create a new jira ticket for removing systemtap.

lorelei-sakai (approved):

raeburn (approved):

PR #139, opened 2024-04-30
[OPS-20441] Made user 'bunsen' use 'jenkins' RSVP

rhawalsh:
To prevent resource starvation while nightly is in process, automated CI testing should use the jenkins resources by default, which are not utilized by nightly.

### DIFF
--- a/src/perl/udstest/Makefile
+++ b/src/perl/udstest/Makefile
@@ -12,7 +12,7 @@ ifdef LOGDIR
   TEST_ARGS += --logDir=$(LOGDIR) --saveServerLogDir=$(LOGDIR)
 endif
 
-ifeq ($(USER),continuous)
+ifeq ($(USER),$(filter $(USER),bunsen continuous))
   RSVP_HOST = PRSVP_HOST=jenkins
 endif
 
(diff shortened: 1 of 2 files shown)


rhawalsh:
While public testing is running, it occurs to me that perhaps we need to modify the public environment to ensure that the hostname 'jenkins' also resolves to the RSVP host.  Otherwise, it's going to try to use 'jenkins' which doesn't otherwise exist until it reaches private testing.

raeburn:
> While public testing is running, it occurs to me that perhaps we need to modify the public environment to ensure that the hostname 'jenkins' also resolves to the RSVP host. Otherwise, it's going to try to use 'jenkins' which doesn't otherwise exist until it reaches private testing.

Sounds like that needs to happen before this change goes in?

rhawalsh:
> > While public testing is running, it occurs to me that perhaps we need to modify the public environment to ensure that the hostname 'jenkins' also resolves to the RSVP host. Otherwise, it's going to try to use 'jenkins' which doesn't otherwise exist until it reaches private testing.
> 
> Sounds like that needs to happen before this change goes in?

Yes.  Will need to make the tweaks.  The environment gets reimaged tomorrow, so it shouldn't be too long.

raeburn (approved):

PR #142, opened 2024-05-03
[VDO-5576] Shut down NFS server after stopping export.

raeburn:
The NFS server appears to hold on to references to some files for a bit otherwise.

Also, assume "systemctl" support and drop "service" path.

### DIFF
--- a/src/perl/Permabit/FileSystem.pm
+++ b/src/perl/Permabit/FileSystem.pm
@@ -109,16 +109,8 @@ sub exportNfs {
   assertTrue($self->{mounted});
   if (!$self->{exported}) {
     my $machine = $self->{machine};
-    # Start the nfs service. Systems with systemd use systemctl; others
-    # use service
     my $command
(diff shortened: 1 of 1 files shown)


lorelei-sakai (approved):
I wonder if all our lab machines now have systemctl support? It doesn't look like there' any reason to merge this fix back to older branches, though, so it doesn't really matter.

raeburn:
> I wonder if all our lab machines now have systemctl support? It doesn't look like there' any reason to merge this fix back to older branches, though, so it doesn't really matter.

I checked Fedora 38 and RHEL 8 and they have systemctl, so I think we're okay.

lorelei-sakai (changes requested):
Oh, wait. shouldn't the nfs server shutdown happen in the unexport() function instead of the unmount one?

I'm wondering what happens in the filesystem was not exported in the first place.

lorelei-sakai (approved):
Looks god to me. Yay.

PR #143, opened 2024-05-08
Delete MegaRaid support, since we don't have MegaRaid cards any more

raeburn:


src/perl/vdotest/VDOTest/VDOFIOPerfTest.pm
```diff
@@ -73,9 +73,7 @@ sub testFioBenchmark {
   my $rate  = $wrate + $rrate;
   my $io    = $wio   + $rio;
 
-  # Print out the SSD wearout indicators
-  # XXX Assert that we're not too worn out?  (worn out SSDs == invalid test?)
-  $machine->printSSDWearout();
+  # To do: Print out the SSD wearout indicators
```
lorelei-sakai, line 77: Given that we don't even have physical hardware any more, can we have meaningful SSD wearout indicators? I'd probably cut the comment, too.

lorelei-sakai (changes requested):
Mostly looks good, just a minor nit in the perf test.

src/perl/vdotest/VDOTest/VDOFIOPerfTest.pm
raeburn: I was thinking of data collection for analysis for performance characteristics when testing on real hardware. But in our main lab testing environment with only virtual systems it's not meaningful...

lorelei-sakai (approved):

PR #144, opened 2024-05-08
dm vdo test: remove flushnop test device

lorelei-sakai:
The flushnop device was intended to speed up tests in sync mode. Now that sync mode has been removed, there is no use for it. Upstream is also unlikely to find utility in this device.


raeburn (approved):

slegendr (approved):
Looks good. There's also a flushnop line in our common perl code for checkserver constants that will need to be removed.

PR #150, opened 2024-05-16
[VDO-5770] Fix tests that use vdoFillIndex

lorelei-sakai:
The first commit adds some error-handling to vdoFillIndex so that if it encouters an error, it doesn't leak an open file descriptor for the VDO pool device. This file descriptor can prevent any other process from accessing the file, potentially making it impossible to restart VDO until it cleaned up.

The second commit fixes the actual test failures, by setting a size in the fake block_device that vdoFillIndex constructs. This should have been fixed as part of PR vdo-devel/148, but I didn't notice this use of struct block_device.

### DIFF
--- a/src/c++/vdo/user/vdoFillIndex.c
+++ b/src/c++/vdo/user/vdoFillIndex.c
@@ -244,6 +244,7 @@ static struct block_device *parse_device(const char *name)
   }
 
   device->fd = fd;
+  device->size = SIZE_MAX;
   return device;
 }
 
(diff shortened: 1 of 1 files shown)


bjohnsto (approved):
Changes look good. I believe Matt plans to create a ticket to better handle cleanup of this scenario, where we enable the vdo storage device for writing, but fail before we disable and reset the full lvm vdo device properly.

lorelei-sakai:
> Changes look good. I believe Matt plans to create a ticket to better handle cleanup of this scenario, where we enable the vdo storage device for writing, but fail before we disable and reset the full lvm vdo device properly.

Yes. I have created VDO-5771 to improve the writable storage issue and make it harder for tests to get in a bad device stack state.

PR #151, opened 2024-05-18
[VDO-5735] Remove unused user tools

lorelei-sakai:
Remove vdodmeventd, vdosetuuid, vdoregenerategeometry, and vdodumpconfig from the user tools package, and from the code.

Commit 5 also respells vdoFillIndex to simplify its interface and remove its dependence on vdodumpconfig. I could have left the latter as a test-only utility, but the restructuring seemed better.


bjohnsto (commented):
I just want to make sure some of these (like vdodumpconfig) aren't used by LVM, before we get rid of them. I've got a question out to Zdenek on this.

raeburn (approved):

PR #152, opened 2024-05-22
[VDO-5771] Fix non-standard device stacks during vdo stop

lorelei-sakai:
Add flags to record when a VDO device enters a non-standard device state (usually to run user programs), and add cleanup code to stop() to revert those changes before doing the rest of the device cleanup.


bjohnsto (changes requested):
Change looks good. It looks like you added some additional spacing in the properties for the new properties. Do you think that should be mentioned in the commit message?

lorelei-sakai:
I was just realigning the delimiters, it didn't seem worth mentioning. The new variable names are pretty long, but I'm open to suggestions for making them shorter.

bjohnsto (approved):

PR #153, opened 2024-05-22
Remove Managed VDO block device

lorelei-sakai:
Also clean up related test, code, and comments.


src/perl/Permabit/BlockDevice/VDO/Upgrade.pm
```diff
@@ -83,17 +83,6 @@ sub _setupData {
   $self->{_versionData} = $SUPPORTED_VERSIONS->{$versionKey};
 }
 
-########################################################################
-# Start a managed VDO
-##
-sub startManagedVDO {
-  my ($self) = assertNumArgs(1, @_);
-  # We need to suppress checking for blocked tasks, because the Fluorine
-  # release still has blocked tasks reported during "dmsetup create".
```
bjohnsto, line 92: So we switched upgrade to use LVMManaged but it looks like we never dealt with this override from Managed here. Do we still do flourine upgrades? It seems like we were trying to block errors here. Do you think we need to  override the activate part of LVMManaged to do something similar?

bjohnsto (changes requested):
I had one question about Upgrade. Otherwise the change looks good.

src/perl/Permabit/BlockDevice/VDO/Upgrade.pm
lorelei-sakai: The Upgrade device and tests were moved to lvm-vdo quite a while ago, if we were missing some handling, we almost certainly would have noticed it. This particular method is not define for an LVMManaged device so this "override" is not getting run.

Fluorine specifically was three versions before Magnesium (== RHEL 7) so it's long obsolete.

bjohnsto (approved):

PR #155, opened 2024-05-23
[VDO-5776] Fix FullWarnDmEventd

lorelei-sakai:
Update FullWarnDmEventd test expectations. The test is checking for a log message created by our custom dmeventd_lvm trigger, which no longer exists. Now that we no longer register a custom trigger, the default LVM VDO plugin will trigger instead, but its warning message is spelled slightly differently.

Additionally, wrap the dumpMetadata call in FullBase in an eval clause. When the test fails this way, the dumpMetadata operation fails, which has the side-effect of short-circuiting cleanup and log collection.

### DIFF
--- a/src/perl/vdotest/VDOTest/FullBase.pm
+++ b/src/perl/vdotest/VDOTest/FullBase.pm
@@ -67,7 +67,9 @@ sub tear_down {
   if ($self->failedTest() && defined($device)
       && $device->isa("Permabit::BlockDevice::VDO")) {
     # Upon failure, save the metadata from the VDO device
-    $device->dumpMetadata();
+    eval {
+      $device->dumpMetadata();
+    };
   }
   $self->SUPER::tear_down();
 }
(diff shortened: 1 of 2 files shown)


lorelei-sakai:
I should be clear that while I expect this fix to work, I can't test it any more. the LVM VDO plugin seems to not exist on most of our pfarms at the moment. (I might have erased some of them while doing diagnostic testing.) We might have to recreate the pfarms or reinstall lvm on them for the test to really start working again.

slegendr:
> I should be clear that while I expect this fix to work, I can't test it any more. the LVM VDO plugin seems to not exist on most of our pfarms at the moment. (I might have erased some of them while doing diagnostic testing.) We might have to recreate the pfarms or reinstall lvm on them for the test to really start working again.

What do you mean by the lvm vdo plugin?

lorelei-sakai:
> > I should be clear that while I expect this fix to work, I can't test it any more. the LVM VDO plugin seems to not exist on most of our pfarms at the moment. (I might have erased some of them while doing diagnostic testing.) We might have to recreate the pfarms or reinstall lvm on them for the test to really start working again.
> 
> What do you mean by the lvm vdo plugin?

The lvm repo has a [dmeventd plugin for VDO]( https://github.com/lvmteam/lvm2/tree/main/daemons/dmeventd/plugins), as Bruce discovered. However, the library that shoudl be built from that doesn't seem to appear on our pfarms, even though the other lvm plugins in the same directory are installed. I don't know if we need to do something additional to enable this plugin for our farms, or if it's there by default but has somehow been removed. Notably, I did see it working in my testing before 2pm this afternoon, so I suspect the latter scenario.

PR #156, opened 2024-05-24
Internalize struct uds_attribute

lorelei-sakai:
This mirrors a change proposed upstream by David Alan Gilbert. Although it has not been accepted upstream yet, it doesn't hurt anything for us to incorporate now in anticipation that it will be accepted.

It also was an oversight that should have been included with PR vdo-devel/99. Since we didn't actually remove the sysfs interfaces internally, this change internalizes this structure also rather than removing it.


slegendr (approved):
Looks good.

PR #157, opened 2024-05-29
Use swap macro from kernel in swap_open_chapter()

lorelei-sakai:
This reflects a change suggested upstream by Jiapeng Chong. It's going to be applied to the dm-devel tree, so apply it to vdo-devel also. If people think it's important I can defer this until we have a proper commit to ingest.


slegendr (approved):
Looks good.

PR #165, opened 2024-06-07
[VDO-5789] Update RHEL 9 migration scenarios

lorelei-sakai:
Update the scenarios used by the migration tests to include the latest from the Chlorine branch, since the older versions no longer build on our lab machines due to kernel updates in RHEL 9.5.


bjohnsto (approved):
Looks good to me. Numbers all match up

PR #167, opened 2024-06-12
Fix the Separated.pm test.

bjohnsto:
In order to fix this test, we need to create a "dummy" test so we can install dmtest. That way we can run dmtest list to get the set of tests we want to run. Then we can take that list and split it up into a single test per item in the list.

### DIFF
--- a/src/perl/dmtest/DMTest/Separated.pm
+++ b/src/perl/dmtest/DMTest/Separated.pm
@@ -13,6 +13,7 @@ use Log::Log4perl;
 
 use Permabit::Assertions qw(assertNumArgs);
 use Permabit::SystemUtils qw(assertSystem);
+use Permabit::UserMachine;
 use Permabit::Utils qw(makeFullPath);
 
 use base qw(DMTest::Grouped);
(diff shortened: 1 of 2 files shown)


C2Redhat (approved):

src/perl/dmtest/DMTest/Separated.pm
```diff
@@ -24,22 +25,32 @@ my $log = Log::Log4perl->get_logger(__PACKAGE__);
 sub suite {
   my ($package) = assertNumArgs(1, @_);
 
+  my @testNames = ();
+  my $machine = undef;
   my $options = $package->makeDummyTest();
```
lorelei-sakai, line 30: It doesn't look like this $machine variable gets used for anything so you should just remove it from the function.

src/perl/dmtest/DMTest/Separated.pm
```diff
@@ -24,22 +25,32 @@ my $log = Log::Log4perl->get_logger(__PACKAGE__);
...
-  if (defined($options->{dmtestName})
-      && ($options->{dmtestName} ne "")
-      && ($options->{dmtestName} !~ /[\[\]{}*?]/)) {
-    # If unitTestName names isn't a pattern but appears to name exactly one
-    # test, just use it. If the name is bad, we'll find out soon enough.
-    $testList = $options->{dmtestName};
-  } else {
-    # TODO: Get all tests from dmtest
-    $testList = "vdo";
-  }
```
lorelei-sakai, line 38: Note that by removing this default, DMTest/Separated will default to running all tests defined in dmtest.

You probably want to update dmtests.suites to pass " --dmTesName =vdo"  as part of the Full alias like the other test aliases do.

lorelei-sakai (changes requested):
You'll want to update the Full alias in the dmtest.suites file as noted in another comment. Also, I think the commit message would be abit better if you specified somwhere that this is the DMTest Separated test (either in the title, or in the text somewhere); I was a little confused when I first looked at it.

Oh, and the wording of the first sentence seem a bit garbled, too. (extra words?)

src/perl/dmtest/DMTest/Separated.pm
bjohnsto: fixed

src/perl/dmtest/DMTest/Separated.pm
bjohnsto: fixed

src/perl/dmtest/DMTest/Separated.pm
```diff
@@ -48,5 +55,4 @@ sub suite {
   }
   return $suite;
 }
-
```
lorelei-sakai, line 51: Shoot, forgot to mention this: could you restore the blank line here, before the final 1; please?

lorelei-sakai (changes requested):
Just one thing I forgot earlier, everything else looks good.

lorelei-sakai (approved):
Looks good, thanks.

src/perl/dmtest/DMTest/Separated.pm
bjohnsto: fixed

PR #168, opened 2024-06-17
Replace use of obsolete vdo-devel mailing list

lorelei-sakai:
Use the dm-devel list instead of vdo-devel, as the latter no longer exists. The author field of the framework is only used for changelog entries in the published spec file, but it should still reflect an address that might work.


slegendr (approved):
Looks good.

PR #169, opened 2024-06-18
[VDO-5793] Add tests to make sure we don't accidentally overwrite VDOs.

bjohnsto:
VDOManagerTestClobber was a test we previously had when vdo manager was our main tool for creating and managing VDO volumes. It consisted of several sub tests that made sure we didn't overwrite existing VDO volumes on a storage device, when calling vdoformat or vdo manager commands like "vdo create".

When we moved to LVM managed VDOs, we removed this test, but we've decided since then that we would like to bring back some of the sub tests.

VDOClobber.pm is the test that attempts that to do that. It uses Unmanaged.pm, which creates VDO volumes directly using dmsetup create and also calls vdoformat directly as well to replicate the previous tests features.

### DIFF
--- a/src/perl/vdotest/VDOTest/VDOClobber.pm
+++ b/src/perl/vdotest/VDOTest/VDOClobber.pm
@@ -1,0 +1,114 @@
+##
+# Test that creating a vdo doesn't clobber a file system or another VDO volume
+#
+# $Id: $
+##
+package VDOTest::VDOClobber;
+
+use strict;
+use warnings FATAL => qw(all);
+use English qw(-no_match_vars);
+use Log::Log4perl;
+
+use Permabit::Assertions qw(
+  assertEvalErrorMatches
+  assertNumArgs
+  assertRegexpMatches
+);
+use Permabit::Constants;
+
+use base qw(VDOTest);
+
+my $log = Log::Log4perl->get_logger(__PACKAGE__);
+
+##
+# @paramList{getProperties}
+our %PROPERTIES
+  = (
+     # @ple set up an Unmanaged device
+     deviceType => "vdo",
+    );
+##
+
+#############################################################################
+##
+sub checkClobberOnlyIfForce {
+  my ($self) = assertNumArgs(1, @_);
+  my $device = $self->getDevice();
+  my $storage = $device->{storageDevice};
+
+  # Try running vdoformat without force. Should complain about signatures.
+  eval {
+    $device->formatVDO({ force => undef });
+  };
+  assertEvalErrorMatches(qr/existing signature|already containing/);
+  # Now create a new VDO using vdoformat. This should work.
+  $device->{_formatted} = 0;
+  $device->start();
+  # Wait till start finishes then stop the device and wipe the geometry block.
+  $device->getMachine->runSystemCmd("sudo udevadm settle");
+  $device->stop();
+  $storage->ddWrite((bs => 4096, count => 1, if => "/dev/zero"));
+}
+
+#############################################################################
+# Test that creating a vdo won't scribble over various types of existing
+# content, unless --force is given to vdoformat.
+##
+sub testClobber {
+  my ($self) = assertNumArgs(1, @_);
+  my $device = $self->getDevice();
+  my $storage = $device->{storageDevice};
+  my $storagePath = $storage->getDevicePath();
+  my $machine = $device->getMachine();
+
+  $device->stop();
+  $self->checkClobberOnlyIfForce();
+
+  my @filesystems = qw(ext3 vfat xfs);
+  foreach my $fstype (@filesystems) {
+    # mkfs.xfs seems to need root privileges
+    $machine->runSystemCmd("sudo mkfs -t $fstype $storagePath");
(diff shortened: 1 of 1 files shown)


src/perl/vdotest/VDOTest/VDOClobber.pm
```diff
@@ -0,0 +1,121 @@
...
+
+use Permabit::Assertions qw(
+  assertEvalErrorMatches
+  assertNumArgs
+  assertRegexpMatches
+);
+use Permabit::CommandString::VDOFormat;
+use Permabit::Constants;
+use Permabit::PlatformUtils qw(getDistroInfo);
+
```
lorelei-sakai, line 21: The VDOFormat ComandString doesn't sem to get used in theses tests.
I'm not usre what you're using from Constants, that might also not be needed?
getDistroInfo() is used, but as I comment below I don't think it's necessary, either.

src/perl/vdotest/VDOTest/VDOClobber.pm
```diff
@@ -0,0 +1,121 @@
...
+use base qw(VDOTest);
+
+my $log = Log::Log4perl->get_logger(__PACKAGE__);
+
+##
+# @paramList{getProperties}
+our %PROPERTIES
+  = (
+     # @ple set up a Linear device
```
lorelei-sakai, line 30: This comment is incorrect. Since these test rely on having an unmanaged device, it might be worth saying something about that. (But it doesn't need an extensive comment.)

src/perl/vdotest/VDOTest/VDOClobber.pm
```diff
@@ -0,0 +1,121 @@
...
+  my $machine = $device->getMachine();
+
+  $device->stop();
+  $self->checkClobberOnlyIfForce();
+
+  my @filesystems = qw(ext3 vfat xfs);
+  # RHEL8 does not support btrfs.
+  if (getDistroInfo($machine->getName()) ne "RHEL8") {
+    push(@filesystems, "btrfs");
+  }
```
lorelei-sakai, line 74: This tree does not work on RHEL8 and we're not going to backport it that far. If you want to use btrfs, just make it part of the normal list.

src/perl/vdotest/VDOTest/VDOClobber.pm
```diff
@@ -0,0 +1,121 @@
...
+    $device->formatVDO();
+  };
+  assertRegexpMatches(qr/The device .* is in use/, $machine->getStderr(),
+                      "vdoformat failed to properly check open device");
+}
+
+#############################################################################
+# Test that vdoformat won't scribble over an no open vdo volume, unless
+# if --force is given.
+##
```
lorelei-sakai, line 106: "unless if" sounds awkwards, doesn't it? I'd s/unless if/unless/

src/perl/vdotest/VDOTest/VDOClobber.pm
```diff
@@ -0,0 +1,121 @@
...
+sub testClobberVDOFormatNotOpen {
+  my ($self) = assertNumArgs(1, @_);
+  my $device = $self->getDevice();
+  my $machine = $device->getMachine();
```
lorelei-sakai, line 110: I think $machine is unused in this test.

lorelei-sakai (changes requested):
Generally these tests make sense. In addition to the separate comments, I'd prefer that  the ticket number is not in the commit message. For test changes like this it doesn't really matter, but I'd like to get in the habit of writing commits that we could push upstream if we need to. (The PR description and the merge commit it creates are sufficient for our internal records, I think.)

src/perl/vdotest/VDOTest/VDOClobber.pm
bjohnsto: I've removed VDOFormat, Constants and getDistroInfo since I removed the btrfs as well

src/perl/vdotest/VDOTest/VDOClobber.pm
bjohnsto: fixed

src/perl/vdotest/VDOTest/VDOClobber.pm
bjohnsto: Removed RHEL8

src/perl/vdotest/VDOTest/VDOClobber.pm
bjohnsto: Agreed, fixed

src/perl/vdotest/VDOTest/VDOClobber.pm
bjohnsto: Removed

src/perl/vdotest/VDOTest/VDOClobber.pm
lorelei-sakai: Constants is still included but it might be needed for something, sure.

lorelei-sakai (approved):
As I commented before, I would prefer that we not include ticket identifiers in commit messages (only in the PR descriptions), but it doesn't matter too much for test changes.

PR #171, opened 2024-07-09
[VDO-5791] Update VDO based on dm discard change

bjohnsto:
Commit 4f563a64732d ("block: add a max_user_discard_sectors queue limit") changed block core to set max_discard_sectors to:
 min(lim->max_hw_discard_sectors, lim->max_user_discard_sectors)

Commit 825d8bbd2f32cb229c3b6653bd454832c3c20acb made fixes to dm to deal with this, by replacing the use of max_discard_sectors with max_hw_discard_sectors. Unfortunately, vdo did not get fixed at that time. I don't think vdo was 
in the kernel at the time of the original fix.

This fix replaces the use of max_discard_sectors in vdo's io_hint function with the use of max_hw_discard_sectors. This will be true for all versions that are 6.10.0 and above.

### DIFF
--- a/src/c++/vdo/base/dm-vdo-target.c
+++ b/src/c++/vdo/base/dm-vdo-target.c
@@ -18,6 +18,9 @@
 #include <linux/kernel.h>
 #include <linux/kstrtox.h>
 #endif /* INTERNAL */
+#ifndef VDO_UPSTREAM
+#include <linux/version.h>
+#endif /* VDO_UPSTREAM */
 
 #include "admin-state.h"
 #include "block-map.h"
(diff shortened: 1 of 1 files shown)


slegendr (approved):
Looks good to me.

lorelei-sakai (changes requested):
Aside from the RHEL version section, the commit message should be written as if it going upstream (which it will): The title should have the correct format and a description of the change for upstream. Commit references should be short hash + title. There should be a "Fixes:" tag that mentions Mike's change 825d8bbd2f32.

Also please rewrite the commit message to say what the change does, and then explain why we need to do that afterwards. Also internal information, such as the paragraph about affecting 6.10, should be in the PR description instead of in the commit message.

lorelei-sakai (changes requested):
Code content looks great.

The commit message still needs some work:
1) The title should start with "dm vdo: "
2) Can you make the title a bit shorter?
3) Use short hashes (12 characters)  and the title when referring to commit IDs (two places)
4) Start the commit message with the thing you fixed. i.e. "Replace use of max_discard_sectors with max_mw_discard_sectors." Move the explanation of why this is needed after to after that. (Alternatiely, update the title to clearly and succinctly describe the change.) (Perhaps "dm vdo: set new hardware discard limit")

Separately from that, can you add a VDO ticket tag to the PR title?

lorelei-sakai (approved):

lorelei-sakai (approved):
Yay

PR #174, opened 2024-07-16
Ingest change to fix vdo documentation formatting

lorelei-sakai:
Patch from Masatake Yamato upstream, plus infrastructure improvements to handle doc files.


slegendr (approved):
Looks good to me.

PR #175, opened 2024-07-17
[VDO-5795] Add dmsetup message for returning config info

bjohnsto:
In previous versions of VDO, we had a user tool called vdoDumpConfig which output configuration info from our geometry block and uds index. We have removed that tool and are replacing the functionality with a new dmsetup message called config, which will output a yaml string containing the same information. The yaml will contain a version number to handle future
modifications to the string.

We have updated the VDO module number so that clients of VDO will know when they can call the new dmsetup message.

### DIFF
--- a/doc/vdo.rst
+++ b/doc/vdo.rst
@@ -251,7 +251,13 @@ The messages are:
 		by the vdostats userspace program to interpret the output
 		buffer.
 
-        dump:
+	config:
+		Outputs useful vdo configuration information. Mostly used
+		by users who want to recreate a similar VDO volume and
+		want to know the creation configuration used.
+
+
+	dump:
 		Dumps many internal structures to the system log. This is
 		not always safe to run, so it should only be used to debug
 		a hung vdo. Optional parameters to specify structures to
--- a/src/c++/vdo/base/dm-vdo-target.c
+++ b/src/c++/vdo/base/dm-vdo-target.c
@@ -1269,6 +1269,14 @@ static int vdo_message(struct dm_target *ti, unsigned int argc, char **argv,
 			*result_buffer = '\0';
 #endif /* __KERNEL__ */
 		result = 1;
+	} else if ((argc == 1) && (strcasecmp(argv[0], "config") == 0)) {
+#ifdef __KERNEL__
+		vdo_write_config(vdo, &result_buffer, &maxlen);
+#else /* not __KERNEL__ */
+		if (maxlen > 0)
+			*result_buffer = '\0';
+#endif /* __KERNEL__ */
+		result = 1;
 	} else {
 		result = vdo_status_to_errno(process_vdo_message(vdo, argc, argv));
 	}
@@ -3074,7 +3082,7 @@ static void vdo_resume(struct dm_target *ti)
 static struct target_type vdo_target_bio = {
 	.features = DM_TARGET_SINGLETON,
 	.name = "vdo",
-	.version = { 9, 0, 0 },
+	.version = { 9, 1, 0 },
 #ifdef __KERNEL__
 	.module = THIS_MODULE,
 #endif /* __KERNEL__ */
--- a/src/c++/vdo/base/message-stats.c
+++ b/src/c++/vdo/base/message-stats.c
@@ -2,8 +2,8 @@
 /*
  * Copyright 2023 Red Hat
  */
-
 #include "dedupe.h"
+#include "indexer.h"
 #include "logger.h"
 #include "memory-alloc.h"
 #include "message-stats.h"
@@ -428,5 +428,52 @@ int vdo_write_stats(struct vdo *vdo, char *buf, unsigned int maxlen)
 	vdo_fetch_statistics(vdo, stats);
 	write_vdo_statistics(NULL, stats, NULL, &buf, &maxlen);
 	vdo_free(stats);
+	return VDO_SUCCESS;
+}
+
+static void write_index_memory(u32 mem, char **buf, unsigned int *maxlen)
+{
+	char *prefix = "memorySize : ";
+
+	/* Convert index memory to fractional value */
+	if (mem == (u32)UDS_MEMORY_CONFIG_256MB)
+		write_string(prefix, "0.25, ", NULL, buf, maxlen);
+	else if (mem == (u32)UDS_MEMORY_CONFIG_512MB)
+		write_string(prefix, "0.50, ", NULL, buf, maxlen);
+	else if (mem == (u32)UDS_MEMORY_CONFIG_768MB)
+		write_string(prefix, "0.75, ", NULL, buf, maxlen);
+	else
+		write_u32(prefix, mem, ", ", buf, maxlen);
+}
+
+static void write_index_config(struct index_config *config, char **buf,
+			       unsigned int *maxlen)
+{
+	write_string("index :  ", "{ ", NULL, buf, maxlen);
+	/* index mem size */
+	write_index_memory(config->mem, buf, maxlen);
+	/* whether the index is sparse or not */
+	write_bool("isSparse : ", config->sparse, ", ", buf, maxlen);
+	write_string(NULL, "}", ", ", buf, maxlen);
+}
+
+int vdo_write_config(struct vdo *vdo, char **buf, unsigned int *maxlen)
+{
+	struct vdo_config *config = &vdo->states.vdo.config;
+
+	write_string(NULL, "{ ", NULL, buf, maxlen);
+	/* version */
+	write_u32("version : ", 1, ", ", buf, maxlen);
+	/* physical size */
+	write_block_count_t("physicalSize : ", config->physical_blocks * VDO_BLOCK_SIZE, ", ",
+			    buf, maxlen);
+	/* logical size */
+	write_block_count_t("logicalSize : ", config->logical_blocks * VDO_BLOCK_SIZE, ", ",
+			    buf, maxlen);
+	/* slab size */
+	write_block_count_t("slabSize : ", config->slab_size, ", ", buf, maxlen);
+	/* index config */
+	write_index_config(&vdo->geometry.index_config, buf, maxlen);
+	write_string(NULL, "}", NULL, buf, maxlen);
 	return VDO_SUCCESS;
 }
--- a/src/c++/vdo/base/message-stats.h
+++ b/src/c++/vdo/base/message-stats.h
@@ -8,6 +8,7 @@
 
 #include "types.h"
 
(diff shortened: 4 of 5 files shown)


doc/vdo.rst
```diff
@@ -251,7 +251,16 @@ The messages are:
...
-        dump:
+	config:
+		Outputs useful vdo configuration information in the form:
+		{ version: 1, physicalSize: <bytes>, logicalSize: <bytes>,
+		  slabSize: <bytes>, index: { memorySize: <gb>,
+		  isSparse: <0 or 1>, },
+		}
+
+{ version: 1, physicalSize, logical }
+
```
lorelei-sakai, line 262: Other message documentation doesn't provide details about their output. I think that is better because we won't have to update documentation every time we make a minor change to the message output. Also, I am not sure what purpose line 261 serves, is it here intentionally?

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -3074,7 +3082,7 @@ static void vdo_resume(struct dm_target *ti)
 static struct target_type vdo_target_bio = {
 	.features = DM_TARGET_SINGLETON,
 	.name = "vdo",
-	.version = { 9, 0, 0 },
+	.version = { 9, 0, 1 },
 #ifdef __KERNEL__
```
lorelei-sakai, line 3086: I feel like this iis more of a y-version change, so move it to 9.1.0 instead of 9.0.1. Any particular reason for this choice?

src/c++/vdo/base/message-stats.c
```diff
@@ -2,8 +2,10 @@
 /*
  * Copyright 2023 Red Hat
  */
+#include <linux/limits.h>
```
lorelei-sakai, line 6: Do you need to include limits.h? It's not clear to me what requires it.

src/c++/vdo/base/message-stats.c
```diff
@@ -430,3 +432,59 @@ int vdo_write_stats(struct vdo *vdo, char *buf, unsigned int maxlen)
 	vdo_free(stats);
 	return VDO_SUCCESS;
 }
+
+static void write_index_memory(char *prefix, u32 mem, char *suffix,
+			       char **buf, unsigned int *maxlen)
+{
+	/* Determine if the index memory value is fractional */
+	if (mem > 1024) {
+		/* Convert index memory to fractional value */
```
lorelei-sakai, line 441: I think this will read cleaner (and be simpler) if you check for the three special values first, and otherwise print out the bare number. (The 1024 value is also kind of magical here, but you shouldn't need it if you check the special values first).

src/c++/vdo/base/message-stats.c
```diff
@@ -430,3 +432,59 @@ int vdo_write_stats(struct vdo *vdo, char *buf, unsigned int maxlen)
...
+	write_block_count_t("physicalSize : ", config->physical_blocks * VDO_BLOCK_SIZE, ", ",
+			    buf, maxlen);
+	/* logical size */
+	write_block_count_t("logicalSize : ", config->logical_blocks * VDO_BLOCK_SIZE, ", ",
+			    buf, maxlen);
+	/* slab size */
+	write_block_count_t("slabSize : ", config->slab_size, ", ", buf, maxlen);
+	/* index config */
+	write_index_config("index : ", vdo, ", ", buf, maxlen);
+	write_string(NULL, "}", NULL, buf, maxlen);
```
lorelei-sakai, line 483: Is there a good reason to make this a nested structure? It might be simple just to put the two index values at the same level as everything else. Also, I'm not sure this needs three functions to write out the config. The existing code is written this way in order to make the code generation simpler but we don't need to cater to that any more.

src/c++/vdo/base/message-stats.h
```diff
@@ -8,6 +8,8 @@
 
 #include "types.h"
 
+int vdo_write_config(struct vdo *vdo, char *buf, unsigned int maxlen);
+
 int vdo_write_stats(struct vdo *vdo, char *buf, unsigned int maxlen);
```
lorelei-sakai, line 13: We probably don't need a blank line between these signature? Not a big deal either way.

src/perl/vdotest/VDOTest/Dmsetup.pm
```diff
@@ -64,6 +65,35 @@ sub testBasicOps {
...
+
+  my $output = $device->runOnHost("sudo dmsetup message $path 0 config");
+  $output = "config : " . $output;
+  my $yaml = yamlStringToHash($output);
+
+  my $slabBits = $self->{slabBits};
+  if (!defined($slabBits)) {
+    # default slab bits from vdoformat
+    $slabBits = 19;
+  }
```
lorelei-sakai, line 85: Can you make sure this is testing with a non-default number of slab bits? It's worth making sure we're actually reading the config.

src/perl/vdotest/VDOTest/Dmsetup.pm
```diff
@@ -64,6 +65,35 @@ sub testBasicOps {
...
+
+  my $config = $yaml->{"config"};
+
+  assertEq($config->{"physicalSize"}, int($device->{storageDevice}->getSize()));
+  assertEq($config->{"logicalSize"}, $device->{logicalSize});
+  assertEq($config->{"slabSize"}, 1 << $slabBits);
+  assertEq($config->{"index"}->{"memorySize"}, $self->{memorySize});
+  assertEq($config->{"index"}->{"isSparse"}, $device->{sparse});
+
+}
```
lorelei-sakai, line 95: You don't need a blank line at the end of a block.

lorelei-sakai (changes requested):
So I have some nitpicky comments about a couple of details but the main issue is that this new message will get pushed upstream, so the commits adding it need to be in a suitable form. This means properly formatting the commit messages, and tailoring the content of the commit messages to the upstream audience.

It is probably best to split this change into a couple of separate commits: one for he message itself, one for the documentation changes, and one for the internal test changes.  (Doc and implementation may be fine as one commit.) The upstream commit message should not refer to anything the upstream version cannot see, such as any version of our user tools. You can refer to the github workflow document for guidance on tailoring commit messages for upstream.

src/c++/vdo/base/message-stats.c
bjohnsto: left over from when I was trying to use U32_MAX. Will remove.

src/c++/vdo/base/message-stats.c
bjohnsto: Its useful to separate it like this, since index: is a sub item in the yaml. This allows it to be index: {  foo }. We could maybe merge everything together, but I much prefer it to be separate.

src/perl/vdotest/VDOTest/Dmsetup.pm
bjohnsto: Ok. I can make another test.

src/perl/vdotest/VDOTest/Dmsetup.pm
bjohnsto: Will remove

src/c++/vdo/base/message-stats.c
bjohnsto: This was how vdo manager did it, so I was just following the pattern. I can rework it if you think its a big deal

src/c++/vdo/base/dm-vdo-target.c
bjohnsto: Not particularly I guess. This seems a minor change so I made it 9.0.1

doc/vdo.rst
bjohnsto: Ok Ill remove the output. I thought it might be useful to clients so they know how to parse it.

src/c++/vdo/base/message-stats.h
bjohnsto: Removed

src/c++/vdo/base/dm-vdo-target.c
bjohnsto: Fixed to 9.1.0

lorelei-sakai (changes requested):
In addition to the specific nits, the commit titles are not sentences; they should not be capitalized and not end with a period. (I think Mike's original guidance said no uppercase letters, no punctuation besides the single colon.)

Also for the first commit, "The output will contain ... " => "The output contains ..." Present tense, written from the world where this comit is part of the canon.

I can fix commit messages later, especially for minor things, but it's better if I don't have to.

src/c++/vdo/base/dm-vdo-target.c
bjohnsto: the 9.1.0 is more of a version that lets clients know when features are changed/added. In this case, we are saying that you can now call this new dmsetup message. LVM uses this to decide when to do various actions with VDO. I still think having a version for the config also makes sense. If we change the format, we don't need to update the module version (this is not something we should be changing often) and just use the format version instead.

doc/vdo.rst
bjohnsto: alright i'll add some more description

src/perl/vdotest/VDOTest/Dmsetup.pm
```diff
@@ -64,6 +65,63 @@ sub testBasicOps {
...
+  my $device = $self->getDevice();
+  my $path = $device->getVDODevicePath();
+
+  my $output = $device->runOnHost("sudo dmsetup message $path 0 config");
+  $output = "config : " . $output;
+  my $yaml = yamlStringToHash($output);
+  my $config = $yaml->{"config"};
+
+  my $slabBits = $self->{slabBits};
+
```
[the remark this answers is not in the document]
bjohnsto: Will turn the two into one test and fix the local variable

src/c++/vdo/base/message-stats.c
```diff
@@ -430,3 +430,52 @@ int vdo_write_stats(struct vdo *vdo, char *buf, unsigned int maxlen)
...
+		write_string(prefix, "0.50", suffix, buf, maxlen);
+	else if (mem == (u32)UDS_MEMORY_CONFIG_768MB)
+		write_string(prefix, "0.75", suffix, buf, maxlen);
+	else
+		write_u32(prefix, mem, suffix, buf, maxlen);
+}
+
+static void write_index_config(char *prefix, struct vdo *vdo, char *suffix,
+			       char **buf, unsigned int *maxlen)
+{
```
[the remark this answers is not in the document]
bjohnsto: Why do you want me to do this differently than all the other callers in this file that use things like "{ " for prefix?  Do you want me to redo the entire file?

PR #178, opened 2024-07-29
dm vdo: Use VDO_USE_NEXT code with RAWHIDE distro

C2Redhat:
The linux-next min_heap code is now integrated into RAWHIDE. Add MACRO to test if a system is RAWHIDE, and use linex-next min_heap code if it is. We also replaced the VDO_USE_ALTERNATE variable with VDO_USE_NEXT if the code block is for the linux-next kernel to clarify the usage.

We also reorder code blocks so that the older code comes first and then the newest code.

### DIFF
--- a/src/c++/vdo/base/repair.c
+++ b/src/c++/vdo/base/repair.c
@@ -28,16 +28,18 @@
 #include "wait-queue.h"
 
 #ifndef VDO_UPSTREAM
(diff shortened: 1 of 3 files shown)


lorelei-sakai (approved):
The commit message could be a bit clearer, but it's fine. Code content all looks correct.

PR #179, opened 2024-07-29
dm vdo: don't refer to dedupe context after releasing it [RHEL-42515,VDO-5794]

raeburn:
Clear the dedupe context pointer in a data VIO whenever ownership of the context is lost, so that we can't examine it accidentally.

### DIFF
--- a/src/c++/vdo/base/dedupe.c
+++ b/src/c++/vdo/base/dedupe.c
@@ -750,6 +750,7 @@ static void process_update_result(struct data_vio *agent)
 	enter_histogram_sample(vdo_from_data_vio(agent)->histograms.update_histogram,
 			       jiffies - context->submission_jiffies);
 #endif /* VDO_INTERNAL */
+	agent->dedupe_context = NULL;
 	release_context(context);
 }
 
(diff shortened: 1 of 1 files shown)


src/c++/vdo/base/dedupe.c
```diff
@@ -750,6 +750,7 @@ static void process_update_result(struct data_vio *agent)
 	enter_histogram_sample(vdo_from_data_vio(agent)->histograms.update_histogram,
 			       jiffies - context->submission_jiffies);
 #endif /* VDO_INTERNAL */
+	agent->dedupe_context = NULL;
 	release_context(context);
```
lorelei-sakai, line 754: Arguably we should use vdo_forget(context) since that's what that sugar does. Or is that awkward here? (This applies to all three lines added.)

lorelei-sakai (changes requested):
Is using vdo_forget() appropriate here?  Otherwise, looks good.

src/c++/vdo/base/dedupe.c
raeburn: Yep, good idea

raeburn: Not with the local variable, but with the struct slot.

raeburn: Actually, I think we don't want to do the store of NULL until (unless) the atomic state change succeeds and we release the context. So the place where we're previously fetching the value from the struct isn't appropriate. And re-fetching the value from the struct with the macro later so that we can use the macro to store the null seems gratuitous...

lorelei-sakai: Yeah, that was kind of the question I was asking. Okay.

lorelei-sakai (approved):
Looks good, then.

PR #180, opened 2024-07-30
[VDO-5742] Improve handling of dirty volumes with old recovery journal format

slegendr:
In the third commit, you will notice that asyncLayer.c was modified to remove a vdo_free() line. This is to correct compilation failures regarding the VDO target being double freed in the case of load failure. Function stopAsyncLayer() already frees the target in the TABLE_LOADED case.

### DIFF
--- a/src/c++/vdo/base/dm-vdo-target.c
+++ b/src/c++/vdo/base/dm-vdo-target.c
@@ -2533,6 +2533,14 @@ static void handle_load_error(struct vdo_completion *completion)
 		return;
 	}
 
+	if ((completion->result == VDO_UNSUPPORTED_VERSION) &&
+	    (vdo->admin.phase == LOAD_PHASE_MAKE_DIRTY)) {
+		vdo_log_error("Aborting load due to unsupported version");
+		vdo->admin.phase = LOAD_PHASE_FINISHED;
+		load_callback(completion);
+		return;
+	}
+
 	vdo_log_error_strerror(completion->result,
 			       "Entering read-only mode due to load error");
 	vdo->admin.phase = LOAD_PHASE_WAIT_FOR_READ_ONLY;
@@ -2981,6 +2989,19 @@ static int vdo_preresume_registered(struct dm_target *ti, struct vdo *vdo)
 		vdo_log_info("starting device '%s'", device_name);
 		result = perform_admin_operation(vdo, LOAD_PHASE_START, load_callback,
 						 handle_load_error, "load");
+		if (result == VDO_UNSUPPORTED_VERSION) {
+			 /*
+			  * A component version is not supported. This can happen when the
+			  * recovery journal metadata is in an old version format. Abort the
+			  * load without saving the state.
+			  */
+			vdo->suspend_type = VDO_ADMIN_STATE_SUSPENDING;
+			perform_admin_operation(vdo, SUSPEND_PHASE_START,
+						suspend_callback, suspend_callback,
+						"suspend");
+			return result;
+		}
+
 		if ((result != VDO_SUCCESS) && (result != VDO_READ_ONLY)) {
 			/*
 			 * Something has gone very wrong. Make sure everything has drained and
@@ -3058,7 +3079,8 @@ static int vdo_preresume(struct dm_target *ti)
 
 	vdo_register_thread_device_id(&instance_thread, &vdo->instance);
 	result = vdo_preresume_registered(ti, vdo);
-	if ((result == VDO_PARAMETER_MISMATCH) || (result == VDO_INVALID_ADMIN_STATE))
+	if ((result == VDO_PARAMETER_MISMATCH) || (result == VDO_INVALID_ADMIN_STATE) ||
+	    (result == VDO_UNSUPPORTED_VERSION))
 		result = -EINVAL;
 	vdo_unregister_thread_device_id();
 	return vdo_status_to_errno(result);
--- a/src/c++/vdo/base/repair.c
+++ b/src/c++/vdo/base/repair.c
@@ -1613,10 +1613,36 @@ static int parse_journal_for_recovery(struct repair_completion *repair)
 	sequence_number_t i, head;
 	bool found_entries = false;
 	struct recovery_journal *journal = repair->completion.vdo->recovery_journal;
+	struct recovery_block_header header;
+	enum vdo_metadata_type expected_format;
 
 	head = min(repair->block_map_head, repair->slab_journal_head);
+	header = get_recovery_journal_block_header(journal, repair->journal_data, head);
+	expected_format = header.metadata_type;
 	for (i = head; i <= repair->highest_tail; i++) {
-		struct recovery_block_header header;
+		header = get_recovery_journal_block_header(journal, repair->journal_data, i);
+		if (header.metadata_type != expected_format) {
+			/* There is a mix of old and new format blocks, so we need to rebuild. */
+			vdo_log_error_strerror(VDO_CORRUPT_JOURNAL,
+					       "Recovery journal is in an invalid format, a read-only rebuild is required.");
+			vdo_enter_read_only_mode(repair->completion.vdo, VDO_CORRUPT_JOURNAL);
+			return VDO_CORRUPT_JOURNAL;
+		}
+
+		if (!is_exact_recovery_journal_block(journal, &header, i, expected_format)) {
+			/* A bad block header was found so this must be the end of the journal. */
+			break;
+		}
+	}
+
+	if (expected_format == VDO_METADATA_RECOVERY_JOURNAL) {
+		/* All journal blocks have the old format, so we need to upgrade. */
+		vdo_log_error_strerror(VDO_UNSUPPORTED_VERSION,
+				       "Recovery journal is in the old format. Downgrade and complete recovery, then upgrade with a clean volume");
+		return VDO_UNSUPPORTED_VERSION;
+	}
+
+	for (i = head; i <= repair->highest_tail; i++) {
 		journal_entry_count_t block_entries;
 		u8 j;
 
@@ -1628,13 +1654,6 @@ static int parse_journal_for_recovery(struct repair_completion *repair)
 		};
 
 		header = get_recovery_journal_block_header(journal, repair->journal_data, i);
-		if (header.metadata_type == VDO_METADATA_RECOVERY_JOURNAL) {
-			/* This is an old format block, so we need to upgrade */
-			vdo_log_error_strerror(VDO_UNSUPPORTED_VERSION,
-					       "Recovery journal is in the old format. Downgrade and complete recovery, then upgrade with a clean volume");
-			return VDO_UNSUPPORTED_VERSION;
-		}
-
 		if (!is_exact_recovery_journal_block(journal, &header, i,
 						     VDO_METADATA_RECOVERY_JOURNAL_2)) {
 			/* A bad block header was found so this must be the end of the journal. */
--- a/src/c++/vdo/base/repair.c
+++ b/src/c++/vdo/base/repair.c
@@ -1631,9 +1631,7 @@ static int parse_journal_for_recovery(struct repair_completion *repair)
 		if (header.metadata_type == VDO_METADATA_RECOVERY_JOURNAL) {
 			/* This is an old format block, so we need to upgrade */
 			vdo_log_error_strerror(VDO_UNSUPPORTED_VERSION,
-					       "Recovery journal is in the old format, a read-only rebuild is required.");
-			vdo_enter_read_only_mode(repair->completion.vdo,
-						 VDO_UNSUPPORTED_VERSION);
+					       "Recovery journal is in the old format. Downgrade and complete recovery, then upgrade with a clean volume");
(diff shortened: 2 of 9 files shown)


src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -2981,6 +2989,18 @@ static int vdo_preresume_registered(struct dm_target *ti, struct vdo *vdo)
 		vdo_log_info("starting device '%s'", device_name);
 		result = perform_admin_operation(vdo, LOAD_PHASE_START, load_callback,
 						 handle_load_error, "load");
+		if (result == VDO_UNSUPPORTED_VERSION) {
+			/*
+			 * A component version is not supported and the load has been aborted.
+			 */
+			vdo_log_error("Start failed, suspending the VDO");
```
lorelei-sakai, line 2996: A minor thiing, I think "suspending VDO" is better than "suspending _the_ VDO". I'm also not sure we need to log anything here given we are generating a log message at every level of the stack for this case.

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -2981,6 +2989,18 @@ static int vdo_preresume_registered(struct dm_target *ti, struct vdo *vdo)
...
+			vdo->suspend_type = VDO_ADMIN_STATE_SUSPENDING;
+			perform_admin_operation(vdo, SUSPEND_PHASE_START,
+						suspend_callback, suspend_callback,
+						"suspend");
+			return result;
```
lorelei-sakai, line 3001: You could just return -EINVAL directly here instead of changing the result in the calling function. There's precedent in the clause above (line 2985).

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -3058,7 +3078,8 @@ static int vdo_preresume(struct dm_target *ti)
 
 	vdo_register_thread_device_id(&instance_thread, &vdo->instance);
 	result = vdo_preresume_registered(ti, vdo);
-	if ((result == VDO_PARAMETER_MISMATCH) || (result == VDO_INVALID_ADMIN_STATE))
+	if ((result == VDO_PARAMETER_MISMATCH) || (result == VDO_INVALID_ADMIN_STATE) ||
+	    (result == VDO_UNSUPPORTED_VERSION))
 		result = -EINVAL;
```
lorelei-sakai, line 3083: As noted above, you don't need to change this line; just return the value you really want from vdo_preresume_registere().

src/c++/vdo/base/repair.c
```diff
@@ -1611,8 +1611,37 @@ static int parse_journal_for_recovery(struct repair_completion *repair)
...
 	bool found_entries = false;
 	struct recovery_journal *journal = repair->completion.vdo->recovery_journal;
+	enum vdo_metadata_type expected_format;
 
 	head = min(repair->block_map_head, repair->slab_journal_head);
+	expected_format = get_recovery_journal_block_header(journal,
+							    repair->journal_data,
+							    head).metadata_type;
+	for (i = head; i <= repair->highest_tail; i++) {
+		struct recovery_block_header header;
```
lorelei-sakai, line 1621: I woudl move the "struct recovery_block_header" declaration outside this loop. It's used in both loops so you'll only have to declare once that way. Also, having the intermeiate 'header' variable will make spelling less awkward when setting the expected_format (line 1617).

src/c++/vdo/base/repair.c
```diff
@@ -1611,8 +1611,37 @@ static int parse_journal_for_recovery(struct repair_completion *repair)
...
+
+		header = get_recovery_journal_block_header(journal, repair->journal_data, i);
+		if (header.metadata_type != expected_format) {
+			/* There is a mix of old and new format blocks, so we need to rebuild */
```
lorelei-sakai, line 1625: Nitpick: This is a sentence, comment should end with a period.

src/c++/vdo/base/repair.c
```diff
@@ -1611,8 +1611,37 @@ static int parse_journal_for_recovery(struct repair_completion *repair)
...
+		}
+
+		if (!is_exact_recovery_journal_block(journal, &header, i, expected_format)) {
+			/* A bad block header was found so this must be the end of the journal. */
+			break;
+		}
+	}
+
+	if (expected_format == VDO_METADATA_RECOVERY_JOURNAL) {
+		/* All journal blocks have the old format, so we need to upgrade */
```
lorelei-sakai, line 1639: Same nitpick: Comment is a sentence, should end with a period.

src/c++/vdo/tests/OldRecoveryJournal_t1.c
```diff
@@ -50,7 +51,8 @@ static const char *RECOVERED = "testdata/vdo.old.rj.recovered";
 static char *crashedPath;
 static char *recoveredPath;
 
-static bool generateFiles = false;
+static bool  generateFiles = false;
+static char *expectedData = NULL;
```
lorelei-sakai, line 55: Nitpick: There would be no reason to change the format of line 54 if you put the new expectedData pointer with the other static char* variables, say after line 52.

src/c++/vdo/tests/OldRecoveryJournal_t1.c
```diff
@@ -141,6 +143,40 @@ static void generate(void)
...
+static void readVDOFromDisk(const char *fileName)
+{
+  int fd;
+  VDO_ASSERT_SUCCESS(open_file(fileName, FU_READ_ONLY, &fd));
+
+  off_t vdoSize;
+  VDO_ASSERT_SUCCESS(get_open_file_size(fd, &vdoSize));
+
+  VDO_ASSERT_SUCCESS(vdo_allocate(vdoSize, char, __func__, &expectedData));
+  ssize_t bytesRead;
```
lorelei-sakai, line 158: This just looks awkward. This is a short enough function, can you pull all the local variable definitions up to the top of the function?

src/c++/vdo/tests/OldRecoveryJournal_t1.c
```diff
@@ -141,6 +143,40 @@ static void generate(void)
...
+/**
+ * Check if the mismatch detected for a given pbn is an acceptable mismatch.
+ *
+ * Implements MismatchChecker.
+ **/
+static void mismatchChecker(physical_block_number_t pbn,
+                            char *expectedBlock,
+                            char *actualBlock)
+{
+  /* There are no acceptable mismatches */
```
lorelei-sakai, line 174: This comment feels at odds with the  function comment at line 166. And you really don't need two comments. Update the function comment to describe this checker. (The 'Implements' note on line 168 is fine.)

src/c++/vdo/tests/OldRecoveryJournal_t1.c
```diff
@@ -141,6 +143,40 @@ static void generate(void)
...
+  CU_FAIL("Unexpected mismatch at pbn %" PRIu64 ": expected '%s', actual '%s'", pbn,
```
lorelei-sakai, line 175: The line wrapping on this line looks weird, just break after the string format and put al three parameters on the next line.

src/c++/vdo/tests/OldRecoveryJournal_t1.c
```diff
@@ -178,13 +214,14 @@ static void testDirty(void)
 {
   parameters.backingFile = crashedPath;
   initializeTest(&parameters);
-  setStartStopExpectation(VDO_READ_ONLY);
-  startVDO(VDO_DIRTY);
+  readVDOFromDisk(crashedPath);
```
lorelei-sakai, line 217: Given the way it's used, 'readVDOFromDisk()" would probably be more descriptively named "saveExpectedFileContents()" or something. The current name doesn't mention that it saves the file contents elsewhere, so that's a bit of an unexpected side-effect with the current name.

Although it's not necessary, you could make "expectedData" a local variable and pass it into readVDOFromDisk() as a parameter. That would also make its usage less magical.

src/c++/vdo/tests/OldRecoveryJournal_t1.c
```diff
@@ -178,13 +214,14 @@ static void testDirty(void)
...
+
+  // VDO_UNSUPPORTED_VERSION is translated to -EINVAL upon load failure
+  startVDOExpectError(-EINVAL);
+
+  checkRAMLayerContents(getSynchronousLayer(), expectedData, mismatchChecker);
   stopVDO();
```
lorelei-sakai, line 223: You should stop VDO, then check the contents. Stopping VDO could write out the superblock or make other modifications.

src/c++/vdo/tests/RecoveryJournal_t2.c
```diff
@@ -305,6 +305,38 @@ static BlockPattern partialHeaderPattern[JOURNAL_BLOCKS] = {
   { 11, 15, GOOD_COUNT, USE_NONCE, FULL_BLOCK,  true,  normalSectors     },
 };
 
+/**
+ * A non-wrapped journal that has been corrupted by the old header
+ * version being written in block 4. The reap head is 1 and the
+ * highest sequence number is 5.
+ *
```
lorelei-sakai, line 312: Blank line in the comment should be removed.

lorelei-sakai (changes requested):
I do like the split into two code commits, I think they are clearer than the previous single commit.

Beside the specific comments, I would like to refine the commit messages more. Rather than try to rewrite them here, can we discuss them tomorrow?

src/c++/vdo/base/dm-vdo-target.c
slegendr: I added it for consistency in the code. I can remove it.

src/c++/vdo/base/dm-vdo-target.c
slegendr: I thought that because this has a status code associated with it, it should be handled by addition to the list of statuses translated to -EINVAL in the calling function. The above precedent does not have a status code associated and would thus be less consistent with the existing check in the calling function.

src/c++/vdo/base/dm-vdo-target.c
slegendr: I don't agree. Because we have a status code associated with the failure, I think it makes the most sense to add it here to the list translated to -EINVAL. That way any status code that results in -EINVAL being returned is listed in a single location. It maintains consistency and code clarity, in my opinion.

src/c++/vdo/base/repair.c
slegendr: Ah yes - had meant to do this and forgot.

src/c++/vdo/tests/OldRecoveryJournal_t1.c
slegendr: I put it here because its declared and defined, versus just declared as the other char * variables are.

src/c++/vdo/tests/OldRecoveryJournal_t1.c
slegendr: Yeah. I had copied it from another test. Will fix.

src/c++/vdo/tests/OldRecoveryJournal_t1.c
slegendr: Updated the function name and input parameters. Made expectedData local to testDirty().

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -2533,6 +2533,14 @@ static void handle_load_error(struct vdo_completion *completion)
 		return;
 	}
 
+	if ((completion->result == VDO_UNSUPPORTED_VERSION) &&
+	    (vdo->admin.phase == LOAD_PHASE_MAKE_DIRTY)) {
+		vdo_log_error("Aborting load due to unsupported version");
+		vdo->admin.phase = LOAD_PHASE_FINISHED;
+		load_callback(vdo_forget(completion));
+		return;
```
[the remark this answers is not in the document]
slegendr: Fixed.

src/c++/vdo/tests/OldRecoveryJournal_t1.c
slegendr: Fixed.

src/c++/vdo/tests/OldRecoveryJournal_t1.c
slegendr: Fixed.

src/c++/vdo/tests/OldRecoveryJournal_t1.c
slegendr: Fixed.

src/c++/vdo/tests/RecoveryJournal_t2.c
slegendr: Fixed.

src/c++/vdo/base/dm-vdo-target.c
lorelei-sakai: Yeah, okay, that makes sense. This spelling is fine.

lorelei-sakai (approved):
Great. All looks good now.

PR #181, opened 2024-07-31
[TECH-170] Add upstreamTests suite for FEDORANEXT OS class

C2Redhat:
Rename rawhideTests to upstreamTests since we will use the same test suite for both kernels. Also, add upstreamTests in nightly run for both RAWHIDE and FEDORANEXT OS class.

### DIFF
--- a/src/perl/vdotest/VDOTest.pm
+++ b/src/perl/vdotest/VDOTest.pm
@@ -253,15 +253,17 @@ sub _minimumMemory {
   # Fedora 38 measured 2023-07-24 needed 400M for pre-VDO operation.
   # Fedora 39 not measured; assume it needs 400M for pre-VDO operation.
   # Fedora 40 not measured; assume it needs 400M for pre-VDO operation.
+  # FEDORANEXT not measured; assume it needs 400M for pre-VDO operation.
   # RHEL8 measured 2023-07-24 needed 400M for pre-VDO operation.
   # RHEL9 measured 2023-07-24 needed 400M for pre-VDO operation. (VDO-5559)
   my %DISTRO_MEMORY_REQUIREMENTS
     = (
-       FEDORA39 => 400 * $MB,
-       FEDORA40 => 400 * $MB,
-       RAWHIDE  => 400 * $MB,
-       RHEL8    => 400 * $MB,
-       RHEL9    => 400 * $MB,
+       FEDORA39   => 400 * $MB,
+       FEDORA40   => 400 * $MB,
+       FEDORANEXT => 400 * $MB,
+       RAWHIDE    => 400 * $MB,
+       RHEL8      => 400 * $MB,
+       RHEL9      => 400 * $MB,
       );
   my $distro   =  getDistroInfo($host);
   assertDefined($DISTRO_MEMORY_REQUIREMENTS{$distro});
--- a/src/perl/vdotest/vdotests.suites
+++ b/src/perl/vdotest/vdotests.suites
@@ -812,9 +812,10 @@ $suiteNames{debugKernelTests}
     ];
 
 ###########################################################################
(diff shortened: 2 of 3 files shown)


src/perl/vdotest/vdotests.suites
```diff
@@ -811,6 +811,23 @@ $suiteNames{debugKernelTests}
...
+     "DiscardPerf",     	# Test that normal discards work. Longer.
+     "Discard512",		# Test that 512 byte discards work
+     "Dmsetup",			# Test that dmsetup interface works
+     "Full01",			# Test filling a device works. Longer.
+     "Logical512",		# Test that 512 byte emulation works
+     "Suspend",			# Test that suspending works
+     "VDOStats",		# Test that vdostats works
+    ];
+
 ###########################################################################
```
lorelei-sakai, line 831: Do we realistically expect that the tests for mainline-next will be different than the ones we want to run on rawhide? If not, we should just have one suite. (The name of the suite could be updated to be more generic, maybe.)

src/perl/vdotest/VDOTest.pm
```diff
@@ -253,15 +253,17 @@ sub _minimumMemory {
...
     = (
-       FEDORA39 => 400 * $MB,
-       FEDORA40 => 400 * $MB,
-       RAWHIDE  => 400 * $MB,
-       RHEL8    => 400 * $MB,
-       RHEL9    => 400 * $MB,
+       FEDORA39   => 400 * $MB,
+       FEDORA40   => 400 * $MB,
+       FEDORANEXT => 400 * $MB,
+       RAWHIDE    => 400 * $MB,
```
lorelei-sakai, line 264: I would slightly prefer that FEDORANEXT is listed after RAWHIDE because technically it will always be newer than RAWHIDE. But it's not a big deal.

lorelei-sakai (changes requested):
This generally seems fine. Do we really need separate suits for linux-next and for rawhide? I would expect we can use the ame suite for both of them.

src/perl/vdotest/VDOTest.pm
C2Redhat: I can change it if you like but I put the list in alphabetical order.  Let me know,  Thanks!

src/perl/vdotest/vdotests.suites
C2Redhat: OK.  Let me update it.

src/perl/vdotest/VDOTest.pm
lorelei-sakai: Either is plausible. You used alphabetical order elsewhere (e.g., in //eng/main) so it's fine here, too.

src/perl/vdotest/vdotests.suites
```diff
@@ -812,9 +812,10 @@ $suiteNames{debugKernelTests}
     ];
 
 ###########################################################################
-# Some tests that exercise various kernel interface areas, to run on rawhide.
+# Some tests that exercise various kernel interface areas, to run on rawhide
+# mainline next systems.
 ##
```
lorelei-sakai, line 817: "rawhide _and_ mainline next tests"

src/perl/vdotest/vdotests.suites
```diff
@@ -812,9 +812,10 @@ $suiteNames{debugKernelTests}
...
-$suiteNames{rawhideTests}
+$suiteNames{rawhideAndMainlineNextTests}
   = [
```
lorelei-sakai, line 819: uit name is a bit long, how about "upstreamTests"?

lorelei-sakai (changes requested):

lorelei-sakai (approved):
Looks good,

PR #182, opened 2024-07-31
[VDO-5749] Add tests for Thin Pools using VDO as data devices.

bjohnsto:
In order to support multiple devices on top on a VDO, LVM has designed a setup where it will use VDO as a data device underneath a Thin Pool device. This will allow the user to create multiple thin devices on top of the thin pool, thus allowing for sharing of a dedupe pool. 

To support this in our test environment, we have split apart our former Thin.pm test device (which was a combination of a thin pool and single thin device on top) into two devices; a normal ThinPool test device (BlockDevice/LVM/ThinPool) and a thin volume (BlockDevice/LVM/Thin).  We then created a VDO backed thin pool test device that provides VDO and LVM support, similar to what LVMManaged (BlockDevice/VDO/LVMManged) does now.  

To make things cleaner, we have created an LVMVDO (BlockDevice/VDO/LVMVDO) base test device that inherits from VDO (BlockDevice/VDO) and LVM (BlockDevice/LVM). The old LVMManaged (now renamed to Managed) and the new thin vdo device (also named ThinPool) have been moved to a new directory called LVMVDO (BlockDevice/VDO/LVMVDO/Managed and BlockDevice/VDO/LVMVDO/ThinPool). BlockDevice/VDO/LVMVDO/ThinPool inherits from BlockDevice/LVM/ThinPool)

Changes were made to VolumeGroup.pm to match the new devices and modified old Thin device. We now have two new functions in VolumeGeroup.pm; createThinVolume and createThinPoolVolume, which cleans up some of the code but also makes it a bit more complicated since we have to add it some additional assertions in createThinVolume to make sure it can only be called when the storage device is a ThinPool device.

We have modified StorageStack to handle the new test devices and updated any old tests that used thin to now create a thinpool and then a thin device on top. This will make all old tests continue to work. Finally we have added one new test called ThinVDO.pm to test Thin with VDO. This currently requires RAWHIDE devices since it has the LVM version we need. There are two subtests in it; one basic test and one with multiple thin volumes on top of a single vdo backed thin pool. Both test writing data and verifying dedupe.


C2Redhat (approved):

lorelei-sakai (changes requested):
I discussed these commits with Bruce on Slack. I had meant to add a comment here, but apparently forgot.

My main points, I believe, were
a) It makes more sense to me to split the commits as one refactoring the old thin device, and one adding the new thinvdo device. Probably the test changes associated with each of those could be in the same commit since they're small.
b) The new ThinVDO.pm file share a lot of code with LVMManaged.pm and we discussed factoring out a common base class to allow sharing code between those two devices.
Also c), (not mentioned before) we should probably add at least one or two BlockDevice01 cases that use the new device(s).

src/perl/vdotest/VDOTest/RebuildStress03.pm
```diff
@@ -193,8 +193,8 @@ sub doThreadCountChange {
 
   # XXX when LVM adds support for changing thread counts (BZ 2070777), remove
   # this and implement properly at the bottom.
-  if ($device->isa("Permabit::BlockDevice::VDO::LVMManaged")) {
-    $log->info("Skipping thread count change on LVMManaged");
+  if ($device->isa("Permabit::BlockDevice::VDO::LVMVDO::Managed")) {
+    $log->info("Skipping thread count change on LVM Managed");
     return;
```
lorelei-sakai, line 198: Hey, does LVM support changing thread counts now? We should probably fix this so the tests can do that again. I'd make a separate ticket for that.

src/perl/Permabit/BlockDevice/VDO/LVMVDO.pm
```diff
@@ -443,4 +303,15 @@ sub enableDeduplication {
...
+# Rename a VDO device. Currently LVM managed devices are the only
+# ones that support rename.
+#
+# @param newName The new name for the VDO device.
+##
+sub renameVDO {
+  my ($self, $newName) = assertNumArgs(2, @_);
+  confess("Failed to override the renameVDO method");
+}
+
```
lorelei-sakai, line 316: It feels weird to move this method for no obvious reason. Why not leave it where it was? (around line 197)

src/perl/Permabit/BlockDevice/VDO/LVMVDO.pm
```diff
@@ -1,9 +1,9 @@
 ##
-# Perl object that represents a VDO Volume managed by LVM
+# Base Perl object that represents a VDO Volume managed by LVM
 #
```
lorelei-sakai, line 3: Did github auto-detect this file motion? It's weird to think of LVMManaged.pm turning into the base class.

src/perl/Permabit/BlockDevice/VDO/LVMVDO/ThinPool.pm
```diff
@@ -0,0 +1,176 @@
...
+  my ($self, $logicalSize) = assertNumArgs(2, @_);
+
+  # resize the vdo logical volume
+  my $name = "$self->{deviceName}";
+
+  my $newSize = $self->{volumeGroup}->alignToExtent($logicalSize);
+  $self->{volumeGroup}->resizeLogicalVolume($name, $newSize);
+  $self->{logicalSize} = $newSize;
+}
+
```
lorelei-sakai, line 61: growPhysical() and growLogical() both look like they have the same implementation in both subclasses of LVMVDO. If this is actaully treu (and not just me misreading), then they should stay in the superclass.

src/perl/Permabit/BlockDevice/VDO/LVMVDO/ThinPool.pm
```diff
@@ -0,0 +1,176 @@
...
+##
+sub disableWritableStorage {
+  my ($self) = assertNumArgs(1, @_);
+  my $deviceName = "$self->{deviceName}pool_vdata";
+  my $vgName   = $self->{volumeGroup}->getName();
+  my $fullName = "$vgName-$deviceName";
+
+  $self->runOnHost("sudo dmsetup remove $fullName");
+}
+
```
lorelei-sakai, line 101: The four readableStorage/writeableStorage methods are almost identical in the two implementations; the only difference is that Managed.pm calls the superclass method, while ThinPool doesn't. But looking at it, the superclass methods only set (or unset) a flag, so that tests can clean up properly if they get interrupted while in one of the disabled states.

So I think both versions should call the superclass method. From that, it also follows that these methods should migrate up into LVMVDO.pm.

src/perl/Permabit/BlockDevice/VDO/LVMVDO/ThinPool.pm
```diff
@@ -0,0 +1,176 @@
...
+  if (defined($self->{volumeGroup})) {
+    my $vgName      = $self->{volumeGroup}->getName();
+    my $storageName = "$vgName-$self->{deviceName}_vpool0_vdata";
+    my $path        = makeFullPath($self->{deviceRootDir}, $storageName);
+
+    return $self->_resolveSymlink($path);
+  }
+  return $self->getStoragePath();
+}
+
```
lorelei-sakai, line 150: These next three methods, (getVDOStoragePath(), setDeviceName() and setSymbolicPath()) look identical betweeo the two implementations except for the $storageName. It seems like it woudl be less repetitive to create a helper method to get the proper storage pool name, and then the rest of the functions could stay in the superclass.

src/perl/vdotest/VDOTest/ThinVDO.pm
```diff
@@ -0,0 +1,168 @@
...
+    "dedupe advice valid"    => 0,
+    "dedupe advice stale"    => 0,
+    "dedupe advice timeouts" => 0,
+    "entries indexed"        => 0,
+  };
+  $self->assertVDOStats($expectedStats);
+
+  # Write some blocks, read them back and verify the data have not changed.
+  my $slice1 = $self->createSlice(blockCount => $self->{blockCount});
+  $slice1->write(tag => "Direct1", direct => 1,);
```
lorelei-sakai, line 80: Trailing comma on this line feel awkward to me. (I know it doesn't do anything functionally.)
Similar comment for lines 88 and 89.
Also at lines 145 and 154.

lorelei-sakai (changes requested):
Okay, I think I generally followed tihs, and it looks like it makes sense.

The things. First, for commit messages, Each commit message should describe what thr commit does (which is mostly true), and the commit title should describe just that one commit (this needs changing). Also, the ticket number should not be in the commit messages, only in the PR title and description. (That doesn't matter too much since this is a test change, but better to build good habits.)

The second commit is really quite dense, so I would prefer to split it up a bit more. Maybe one commit for moving the Managed device to a new location, one commit for adding the new LVMVDO::ThinPool class, and maybe a separate one for adding the test. (Just separating the code motion would make this a bit more tractable, I think.)

src/perl/Permabit/BlockDevice/LVM/Thin.pm
```diff
@@ -15,12 +15,40 @@ use base qw(Permabit::BlockDevice::LVM);
...
+}
+
+########################################################################
+# @inherit
+##
+sub configure {
+  my ($self, $arguments) = assertNumArgs(2, @_);
+  $self->SUPER::configure($arguments);
+  $self->{lvmSize} = $self->{storageDevice}->getSize();
+  $self->{lvmSize} = $self->{volumeGroup}->alignToExtent($self->{lvmSize});
```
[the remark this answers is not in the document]
bjohnsto: > This spelling seems odd to me. The SUPER method also tries to set lvmSize so you're actually doing it twice by copying these lines. Also, since we have the optional $arguments hash, couldn't you just add lvmSize to the hash of arguments, especially since you're setting it unilaterally? That might be cleaner.

True. The wording was bad and not consistent with the code. I've reworked both to make it more understandable.

src/perl/Permabit/BlockDevice/LVM/Thin.pm
```diff
@@ -15,12 +15,40 @@ use base qw(Permabit::BlockDevice::LVM);
...
+  my ($self) = assertNumArgs(1, @_);
+  # This will create the bottom device if needed.
+  $self->SUPER::checkStorageDevice();
+  # If our storage device is a thin pool (this includes vdo thin pools),
+  # don't do anything, just use it.
+  if ($self->{storageDevice}->isa("Permabit::BlockDevice::LVM::ThinPool")) {
+    $self->{volumeGroup} = $self->{storageDevice}->{volumeGroup};
+    return;
+  }
+  # If we don't have a thin pool already, create a generic one.
```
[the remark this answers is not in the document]
bjohnsto: added

src/perl/Permabit/BlockDevice/LVM/Thin.pm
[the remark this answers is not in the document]
bjohnsto: The problem here is that the thinpool (if it doesn't already exist) is created in the configure method of thin.pm in the BlockDevice::configure method when it calls checkStorageDevice. So we are waiting till the super::configure is done before we can get the size of the thin pool which is what we want to use as the size for the thin volumes. I'm not sure what the best solution would be here. We could assert the thinpool is already created before and don't auto create it, but I put that there so that all our old tests which just use thin as the device type would just "work" with the new setup.

src/perl/Permabit/VolumeGroup.pm
```diff
@@ -177,17 +179,57 @@ sub createThinVolume {
...
+  # To create the pmspare ("pool metadata spare") LV, lvm first creates an LV with a
+  # default name, e.g. lvol0, and then converts this LV to a hidden LV with the
+  # _pmspare suffix, e.g. lvol0_pmspare.
+  #
+  # If you try to not create a spare using --poolmetadataspare n, the output will
+  # still generate a WARNING to stderr, thus causing our perl code to fail. Therefore,
+  # we will attempt to shrink the overall size of the pool by the default 30 extents
+  # the spare requires.
+  my $spareMetaSize = $self->{physicalExtentSize} / $KB * 60;
+  $ksize = $ksize - $spareMetaSize;
```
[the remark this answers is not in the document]
bjohnsto: You're right. The comments were not accurate. I've reworked the code and comments to make things clearer (I hope).

src/perl/Permabit/VolumeGroup.pm
```diff
@@ -177,17 +179,57 @@ sub createThinVolume {
...
+                         . " --size ${ksize}K --yes $config"
                          . " $self->{volumeGroup} </dev/null");
 
+  # Make sure to wait for the udev event recording the new event has
+  # been processed. Otherwise, the lv may be open due to blkid when
+  # we try to disable it below.
+  $machine->sendCommand("sudo udevadm settle");
+  # XXX As soon as we get rid of Squeeze, on which lvcreate doesn't have the
+  # -a flag, we can stop doing two separate commands and use lvcreate -an.
+  $self->disableLogicalVolume($name);
```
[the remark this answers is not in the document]
bjohnsto: Updated to remove the old code and use -an and -kn. Seems to work

src/perl/Permabit/BlockDevice/VDO/LVMVDO/ThinPool.pm
bjohnsto: growphysical and growlogical have been stubbed out for now for thinpoolvdo. Managed still grows the devices as before.

src/perl/Permabit/BlockDevice/LVM/Thin.pm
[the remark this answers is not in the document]
lorelei-sakai: I don't know why I didn't mention this before. Couldn't you add a parameter that basically says "Use the full size of the pool"? That's really what you want here, right? I get that you don't have a size you can pass in, but can't we just tell the superclass to Do The Right Thing?

src/perl/Permabit/BlockDevice/VDO/LVMVDO.pm
lorelei-sakai: This comment now applies to a lot more functions. For example, enable/disable ReadableStorage. I get that you need to change the methods, but putting them in a different place in the file makes the diff hard to read, and it's also hard to parse out what you really changed about those functions.

Basically, if a function existed in LVMManaged,pm, and it now exists in LVMVDO.pm, could you make sure the function order lines up so the diffs are clearer to read?
rename VDO() is an exmple of how I think this should be done. Make sure the function is in the same place in the file, so the diffs are dlear about what parts of the implementation changed and which didn't.

src/perl/vdotest/VDOTest/RebuildStress03.pm
lorelei-sakai: I made a tech task for this.

src/perl/vdotest/VDOTest/BlockDevice01.pm
```diff
@@ -46,22 +46,22 @@ my @deviceTypes = qw(
...
   vdo-loop
   vdo-raw
   vdo-stripfua
-  lvmvdo-crypt-raw
-  lvmvdo-dory-raw
-  lvmvdo-iscsi-linear
-  lvmvdo-tracer-raw
-  thin-lvmvdo-thin
-  thin-vdo-thin
   tracer-corruptor-raw
```
lorelei-sakai, line 65: The logic of the code motion here evades me.  Two things:
* The old list was two-device stacks, followed by three-devices stacks, each section alphabetized. We could preserve that. Or, it looks like you were trying to go for strictly alphabetical, in which case "tracer-corrupter-raw" and several of the vdo-foo stack are misalphabetized. I could see an argument for either choice, just please be consistent.
* Some combinations for device stacks have gone away (such as thin-vdo-thin). Is here a logic to which ones you removed? And can you describe that logic somewhere, either in the PR description or in the commit message?

src/perl/Permabit/VolumeGroup.pm
```diff
@@ -145,29 +146,24 @@ sub createLogicalVolume {
...
-# @param size  Thin volume size in bytes
+# @param  name    Thin pool name
+# @param  size    Thin pool size in bytes
+# @oparam config  String of create time lvm.conf overrides
+# @oparam vdo     Whether the thin pool uses VDO or not
 ##
-sub createThinVolume {
-  my ($self, $name, $size) = assertNumArgs(3, @_);
+sub createThinPool {
+  my ($self, $name, $size, $config, $vdo) = assertMinMaxArgs(["", 0], 3, 5, @_);
```
lorelei-sakai, line 162: $vdo here is a boolean, yes? Could you name it some thing like "useVdo" so it's clear that it's a boolean, and what the flag means? Just calling it $vdo makes me expect a reference to an object representing a VDO device or something.

lorelei-sakai (changes requested):
Hm, okay. I resolved the previous coments athat seem to have been dealt with. In general the structure and the comments feel cleaner. I do have some confusion on a few points, though.

Aside from the things noted separately, it seems we're back to one commit? That's probably okay; the complicateed part fo he change doesn't really split up. But The commit message still says "The first part..." when there's no second part.

I would prefer having a couple of commits so you can have targeted descriptions of each logical bit of change (e.g., the changes in the scenario list in BlockDevice01), but if you do want to stick with one commit, can you at least fix the commit message so it accurately describes the entire change?

Also I didn't read LVMVDO.pm very carefully, due the issue with the diffs I mentioned. I think it mostly looks fine but there might still be an issue hiding in there.

src/perl/Permabit/BlockDevice/LVM/Thin.pm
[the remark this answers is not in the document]
bjohnsto: I think I figured out a way to do this without the need for a configure override. I've put the setting of lvmsize in checkStorageDevice so that its set to the pool device size (not the volume group size as LVM.pm does).

src/perl/Permabit/VolumeGroup.pm
bjohnsto: renamed

src/perl/vdotest/VDOTest/BlockDevice01.pm
bjohnsto: redone

src/perl/Permabit/BlockDevice/VDO/LVMVDO.pm
bjohnsto: I've changed the locations to hopefully be correct now. There are some functions that aren't there in the new file, but hopefully diff will deal.

src/perl/Permabit/VolumeGroup.pm
```diff
@@ -177,27 +172,58 @@ sub createThinVolume {
...
+  # --poolmetadataspare n, the output will still generate a WARNING to stderr, thus
+  # causing our perl code to fail.
+  #
+  # As such, we have to take in account the size of both the metadata device and the
+  # spare metadata device when calculating how much space the pool takes as it doesn't
+  # take that into account itself. So we take the overall size asked and add the size
+  # needed for both the metadata and spare metadata devices. Then we see if the VG
+  # has sufficient space. Currently that space is 30 extents for each device.
+  my $metaDataSize = 60 * ($self->{physicalExtentSize} / $KB);
+  my $fullSize = $ksize + $metaDataSize;
```
lorelei-sakai, line 188: The comment makes this much clearer now.
(60 * $self->{physicalExtentSize}) / $KB as a spelling communicate the idea of "60 extents" better, to me, but it's a minor thing.

src/perl/vdotest/VDOTest/BlockDevice01.pm
lorelei-sakai: Okay, looks mostly reasonable. But the now thin-thinpool cases have more devices in the stack. So thin-thinpool is as two-devices stack, in should be listed with the other two device stacks. Similarly, the other new cases are three-device stacks and should go with the three-device stacks.

src/perl/Permabit/BlockDevice/VDO/LVMVDO.pm
```diff
@@ -183,40 +162,53 @@ sub installModule {
 }
 
 ########################################################################
-# @inherit
+# Get the lvm name of the device on top of VDO. Implemented in inheriting
+# classes.
 ##
-sub growLogical {
-  my ($self, $logicalSize) = assertNumArgs(2, @_);
```
lorelei-sakai, line 189: It seem to me that, much like renameVDO(), growPlogical() and growPhysical() should still be defined in the parent class, with an implmentation thta errors out if a subclass fails to override it. (That would also make these diffs cleaner, I think.)

src/perl/Permabit/BlockDevice/VDO/LVMVDO.pm
```diff
@@ -137,27 +137,6 @@ sub makeLVMConfigString {
   return join(" ", @values);
 }
 
-########################################################################
-# @inherit
-##
-sub setup {
-  my ($self) = assertNumArgs(1, @_);
```
lorelei-sakai, line 144: Potentially a stub for setup() should still exist for common bits like finding the vdoformat executable. I don't know how hard that would be to do.

lorelei-sakai (changes requested):
Mostly minor cleanup things, and a couple small typos. The diffs for LVMVDO.pm are much better now, thank you. The new classes and the new test all seem to make sense.

This series could be split into more commits to make it clearer, but I won't insist. (The -an change is logically separate and could be its own commit; similarly the nex test is logically separate and could be described in a separate commit.)

src/perl/Permabit/VolumeGroup.pm
```diff
@@ -145,29 +146,23 @@ sub createLogicalVolume {
...
-# @param size  Thin volume size in bytes
+# @param  name    Thin pool name
+# @param  size    Thin pool size in bytes
+# @oparam config  String of create time lvm.conf overrides
 ##
-sub createThinVolume {
-  my ($self, $name, $size) = assertNumArgs(3, @_);
+sub createThinPool {
+  my ($self, $name, $size, $config) = assertMinMaxArgs(["", 0], 3, 4, @_);
   my $machine = $self->{storageDevice}->getMachine();
```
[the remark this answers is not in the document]
bjohnsto: fixed

src/perl/Permabit/VolumeGroup.pm
bjohnsto: I've changed this to do this ordering

src/perl/Permabit/VolumeGroup.pm
```diff
@@ -250,17 +276,8 @@ sub createVDOVolume {
   }
 
   my $args = join(' ', map { "--$_ $args{$_}" } keys(%args));
-  $machine->runSystemCmd("sudo lvcreate $args --yes -ay $config"
+  $machine->runSystemCmd("sudo lvcreate $args --yes -an -kn $config"
```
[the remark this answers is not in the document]
bjohnsto: made it separate

src/perl/Permabit/BlockDevice/VDO/LVMVDO.pm
bjohnsto: i've added stubs for growlogical and growphysical into LVMVDO.pm

src/perl/vdotest/VDOTest/BlockDevice01.pm
bjohnsto: reordered back to using device number as sections

src/perl/Permabit/BlockDevice/VDO/LVMVDO.pm
lorelei-sakai: Sure, we can fix that up later if we end up deciding it needs to get done.

lorelei-sakai (approved):
Some very minor comments, but nothing to block this, especially since it's a test change. This should still wait to merge until it passes CI again though (infrastructure issues with jfarms made it fail). Have you run these testcases recently?

* The commit title should not end with a period (or any punctuation at all, really).
* I think it would be cleaner to shuffle the -an -kn change to the front instead of adding the old handling in VolumeGroup.pm in commit 1 just to rewrite it in commit 3. I'd probably insist on that if this were going upstream.

lorelei-sakai:
Man I wish the "Show unresolved conversations" link was more useful. It tries to take me to the conversation in code, but for outdated conversations, there's no code to link to so it's useless. I think I've resolved all the outstanding conversations though (by just scrolling through this conversations list).

PR #184, opened 2024-08-07
[VDO-5804] Skip corruption check for invalid recovery journal blocks

slegendr:
I was not able to reproduce the issue we've been seeing in nightly in the TornWrite tests due to its spurious nature. However, I believe the issue may be due to us checking for invalid blocks after we check for corruption. As such, I have moved around those checks so that we first verify we have a valid recovery journal block before we check if it is corrupted.

These modifications made it possible to simplify the parse_journal_for_recovery() function and eliminate the first loop over the recovery journal blocks that had been added in PR180.

### DIFF
--- a/src/c++/vdo/tests/RecoveryJournal_t2.c
+++ b/src/c++/vdo/tests/RecoveryJournal_t2.c
@@ -113,6 +113,21 @@ const SectorPattern noSectors[VDO_SECTORS_PER_BLOCK] = {
 };
 
 /**
+ * A non-wrapped journal that has not been corrupted and has no holes.
+ * The reap head is 1 and the highest sequence number is 5.
+ **/
+static BlockPattern basicFullBlocksPattern[JOURNAL_BLOCKS] = {
+  { 0, 0, GOOD_COUNT, USE_NONCE, FULL_BLOCK, false, normalSectors },
+  { 1, 1, GOOD_COUNT, USE_NONCE, FULL_BLOCK, true,  normalSectors },
+  { 1, 2, GOOD_COUNT, USE_NONCE, FULL_BLOCK, true,  normalSectors },
+  { 1, 3, GOOD_COUNT, USE_NONCE, FULL_BLOCK, true,  normalSectors },
+  { 1, 4, GOOD_COUNT, USE_NONCE, FULL_BLOCK, true,  normalSectors },
+  { 1, 5, GOOD_COUNT, USE_NONCE, FULL_BLOCK, true,  normalSectors },
+  { 0, 0, GOOD_COUNT, USE_NONCE, FULL_BLOCK, false, normalSectors },
+  { 0, 0, GOOD_COUNT, USE_NONCE, FULL_BLOCK, false, normalSectors },
+};
+
+/**
  * A wrapped journal with a reap head at block 6 and the tail at a partial
  * block1. The reap head is 14 and the highest sequence number is 17.
  **/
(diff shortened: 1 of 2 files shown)


lorelei-sakai (changes requested):
The code changes looks good. I'll probably squash this commit with the second one from PR 180 for going upstream. It would be nice if we had a testcase that could tell the difference between this code and the prior version, though.

slegendr:
> The code changes looks good. I'll probably squash this commit with the second one from PR 180 for going upstream. It would be nice if we had a testcase that could tell the difference between this code and the prior version, though.

I added a commit with tests of basic recovery, as well as of a journal with a block that has a corrupt header and a hole due to a bad sequence number. The head and sequence numbers are not realistic for the non-wrapped journal patterns I added, as there is not sufficient documentation on how to create these patterns.

src/c++/vdo/tests/RecoveryJournal_t2.c
```diff
@@ -112,6 +112,21 @@ const SectorPattern noSectors[VDO_SECTORS_PER_BLOCK] = {
   { TEAR_OLD, FULL_SECTOR,  GOOD_COUNT, APPLY_NONE },
 };
 
+/**
+ * A non-wrapped journal that has not been corrupted and has no holes.
+ * The reap head is 1 and the highest sequence number is 5.
+ **/
+static BlockPattern basicRecoveryPattern[JOURNAL_BLOCKS] = {
+  { 0, 0, GOOD_COUNT, USE_NONCE, FULL_BLOCK, false, normalSectors },
```
lorelei-sakai, line 120: I'm probably call this "fullBlocksPattern" or basicFullBlocksPattern" for this. "basic" is a bit generic as a description, and "recovery" is implicit in the test. The pattern itself looks good.

src/c++/vdo/tests/RecoveryJournal_t2.c
```diff
@@ -321,6 +336,23 @@ static BlockPattern corruptHeaderPattern[JOURNAL_BLOCKS] = {
...
+ * to a bad sequence number. The reap head is 1 and the highest sequence
+ * number is 5.
+ **/
+static BlockPattern corruptHeaderBadSequencePattern[JOURNAL_BLOCKS] = {
+  { 0, 0, GOOD_COUNT, USE_NONCE,    FULL_BLOCK, false, normalSectors },
+  { 1, 1, GOOD_COUNT, USE_NONCE,    FULL_BLOCK, true,  normalSectors },
+  { 1, 2, GOOD_COUNT, USE_NONCE,    FULL_BLOCK, true,  normalSectors },
+  { 1, 3, GOOD_COUNT, USE_NONCE,    FULL_BLOCK, true,  normalSectors },
+  { 1, 8, GOOD_COUNT, BAD_METADATA, FULL_BLOCK, false, normalSectors },
+  { 1, 5, GOOD_COUNT, USE_NONCE,    FULL_BLOCK, false, normalSectors },
```
lorelei-sakai, line 351: Instead of "1, 8", use "9, 12".

src/c++/vdo/tests/RecoveryJournal_t2.c
```diff
@@ -453,6 +485,12 @@ static void attemptRebuild(CorruptionType  corruption,
   }
 }
 
+/**********************************************************************/
+static void testBasicRecovery(void)
+{
```
lorelei-sakai, line 490: Similar to the above, maybe "testRebuildFullBlocks"?

lorelei-sakai (changes requested):
just some very minor things, new testcases look good.

src/c++/vdo/tests/RecoveryJournal_t2.c
slegendr: Fixed

src/c++/vdo/tests/RecoveryJournal_t2.c
slegendr: Fixed.

src/c++/vdo/tests/RecoveryJournal_t2.c
slegendr: Fixed.

src/c++/vdo/tests/RecoveryJournal_t2.c
```diff
@@ -321,6 +336,23 @@ static BlockPattern corruptHeaderPattern[JOURNAL_BLOCKS] = {
   { 0, 0, GOOD_COUNT, USE_NONCE,    FULL_BLOCK, false, normalSectors },
 };
 
+/**
+ * A non-wrapped journal that has been corrupted by the old header
+ * version being written in block 4, along with a hole in block 4 due
```
lorelei-sakai, line 341: Test comment is technically wrong (sorry), it's a wrapped journal now  (Could just say "A journal")
I'd also dro reference s to "block 4" specfically,  just say "... corrupted by a hole with the old header version. (Being a hole implies the older sequence number.)

lorelei-sakai (changes requested):

src/c++/vdo/tests/RecoveryJournal_t2.c
slegendr: Fixed.

lorelei-sakai (approved):
Great. Looks good.

PR #185, opened 2024-08-08
Update regexp for catching rpm -i errors, and check stderr too.

raeburn:
It missed a failure because stdout said "Failed" and stderr said "failed" but we only checked for "failed" on stdout.


lorelei-sakai (approved):
This change makes sense. I'm not sure why you put a line break in the concatenation, though. And it might be nice for an internal change to actaulyl have some detail in the commit message, but not a big deal.

The test failure looks like Volume_n4 again (MT 4 stress).

raeburn:
Without the line break the concatenation expression goes to 88 columns; was just trying to keep to 80 like we mostly still do for perl.

I'll update the commit message.

PR #187, opened 2024-08-21
dm vdo: rename useUpstreamKernel , add useUpstreamModule flag and useUpstream test suite

C2Redhat:
This PR contains 3 commit:

. Update useUpstreamKernel flag to useDmVdoModule
. Add useUpstreamModule option to load upstream dm-vdo for testing instead of kvdo. We check if 
useLinuxNextModule is 1, and set the appropriate modName when modprobe is called in KernelModule.
We sill need to load the kvdo module because the user module vdo still requires kvdo to be loaded.
. Add useUpstream test suite

### DIFF
--- a/src/perl/Permabit/BlockDevice/VDO.pm
+++ b/src/perl/Permabit/BlockDevice/VDO.pm
@@ -58,6 +58,7 @@ my $VDO_WARMUP               = "src/c++/vdo/bin/vdoWarmup";
 my $VDO_INITIALIZE_BLOCK_MAP = "src/c++/vdo/bin/vdoInitializeBlockMap";
 
 my $VDO_USER_MODNAME = "vdo";
+my $VDO_UPSTREAM_MODNAME = "dm-vdo";
 
 # The keys of this hash are the names of the histogram that records the
 # statistic, and the values are used in the error message if any request
@@ -177,6 +178,8 @@ our %BLOCKDEVICE_INHERITED_PROPERTIES
      useDistribution           => 0,
      # The directory to put the user tool binaries in
      userBinaryDir             => undef,
+     # Whether to use dm-vdo
+     useUpstreamModule         => 0,
      # UUID for VDO volume.
      uuid                      => undef,
      # Bitmask of CPUs on which to run VDO threads
@@ -252,7 +255,7 @@ sub configure {
   $self->{moduleVersion} //= $self->{vdoModuleVersion};
   $self->{moduleVersion} //= $VDO_MARKETING_VERSION;
 
-  if ($self->{useDistribution}) {
+  if ($self->{useDistribution} || $self->{useUpstreamModule}) {
     $self->getMachine()->{userBinaryDir} = "/usr/bin";
   }
 }
@@ -539,6 +542,13 @@ sub suspend {
 ##
 sub getModuleVersion {
   my ($self) = assertNumArgs(1, @_);
+  if ($self->{useUpstreamModule}) {
+    my $getVerCmd = "yum list $VDO_USER_MODNAME.`uname -m` | " .
+                    "awk '/^$VDO_USER_MODNAME/ {print \$2}'";
+    my @ver = split(/\./,
+		    runCommand($self->getMachineName(), $getVerCmd)->{stdout});
+    $self->setModuleVersion("$ver[0].$ver[1]");
+  }
   return $self->{moduleVersion};
 }
 
@@ -550,6 +560,20 @@ sub getModuleVersion {
 sub setModuleVersion {
   my ($self, $versionString) = assertNumArgs(2, @_);
   $self->{moduleVersion} = $versionString;
+}
+
+########################################################################
+# @inherit
+##
+sub getModuleName {
+  my ($self) = assertNumArgs(1, @_);
+
+  # If useUpstreamModule is set, we need to use $VDO_UPSTREAM_MODNAME(dm-vdo)
+  # module otherwise we load the $VDO_MODNAME(kvdo)
+  if ($self->{useUpstreamModule}) {
+    return $VDO_UPSTREAM_MODNAME;
+  }
+  return $self->SUPER::getModuleName();
 }
 
 ########################################################################
@@ -578,7 +602,9 @@ sub installModule {
   my ($self) = assertNumArgs(1, @_);
   my $machineName = $self->getMachineName();
   my $moduleName = $self->getModuleName();
+
   my $version = $self->getModuleVersion();
+
   if ($self->{_modules}{$machineName}) {
     if ($self->{_modules}{$machineName}{$moduleName}{modVersion} eq $version) {
       $log->info("Module $moduleName $version already installed"
@@ -600,6 +626,7 @@ sub installModule {
                                   modName         => $moduleName,
                                   modVersion      => $version,
                                   useDistribution => $self->{useDistribution},
+                                  useUpstream     => $self->{useUpstreamModule},
                                  );
   $module->load();
   $self->{_modules}{$machineName}{$moduleName} = $module;
@@ -613,6 +640,7 @@ sub installModule {
                                 modName         => $VDO_USER_MODNAME,
                                 modVersion      => $version,
                                 useDistribution => $self->{useDistribution},
+                                useUpstream     => $self->{useUpstreamModule},
                                );
   $userModule->load();
   $self->{_modules}{$machineName}{$VDO_USER_MODNAME} = $userModule;
--- a/src/perl/Permabit/KernelModule.pm
+++ b/src/perl/Permabit/KernelModule.pm
@@ -63,6 +63,7 @@ sub load {
         $log->debug("kernel log additions:\n$additions");
       }
     };
+    $log->info("Using kernel module: " . $modName);
     $self->_step(command => "sudo modprobe $modName $self->{modprobeOption}",
                  cleaner => "sudo modprobe -r $modName",
                  diagnostic => $uponModprobeError);
@@ -78,6 +79,12 @@ sub loadFromFiles {
   my ($self) = assertNumArgs(1, @_);
   my $modName = $self->{modName};
   my $modVer = $self->{modVersion};
+
+  # With useUpstream test, we don't need to load upstream
+  # module(dm-vdo), since it is in the kernel already.
+  if ($self->{useUpstream}) {
+    return 1;
+  }
 
   my $loaded = $self->SUPER::loadFromFiles();
   if ($loaded == 0) {
--- a/src/perl/Permabit/ModuleBase.pm
+++ b/src/perl/Permabit/ModuleBase.pm
@@ -42,6 +42,8 @@ my %PROPERTIES
      modVersion      => undef,
      # @ple whether the test is loading the module from a released RPM
      useDistribution => undef,
+     # @ple whether the test is loading the dm-vdo module from kernel
+     useUpstream     => 0,
     );
 ##
 
--- a/src/perl/Permabit/UserModule.pm
+++ b/src/perl/Permabit/UserModule.pm
@@ -40,6 +40,30 @@ sub load {
 ###############################################################################
 # @inherit
 ##
+sub loadFromFiles {
+  my ($self) = assertNumArgs(1, @_);
+  my $machine = $self->{machine};
+  my $modFileName = $self->{modFileName};
+
+  if ($self->{useUpstream}) {
+    $log->debug("Using upstream version VDO: $self->{modVersion}");
+    my $topdir = makeFullPath($machine->{workDir}, $self->{modVersion});
+    $self->_step(command => "mkdir -p $topdir");
+    my $getFromDnf = join(' ',
+			  "dnf", "download", "--destdir",
+			  "$self->{modDir}", "$modFileName");
+    $self->_step(command => $getFromDnf);
+    $getFromDnf = join(' ',
+                       "dnf", "download", "--destdir", "$topdir",
+                       "$modFileName-support");
+    $self->_step(command => $getFromDnf);
+  }
+  $self->SUPER::loadFromFiles();
+}
+
+###############################################################################
+# @inherit
+##
 sub loadFromBinaryRPM {
   my ($self, $filename, $modFileName) = assertMinMaxArgs([undef], 2, 3, @_);
   $modFileName //= $self->{modFileName};
--- a/src/perl/nightly/NightlyBuildType/VDO.pm
+++ b/src/perl/nightly/NightlyBuildType/VDO.pm
@@ -70,6 +70,14 @@ my $SUITE_PROPERTIES = {
     extraArgs    => "--clientClass=PFARM",
     osClasses    => ["FEDORANEXT"],
   },
+  vdoUpstreamTests => {
+    displayName    => "VDO_Upstream_Tests",
+    suiteName      => "upstreamTests",
+    scale          => "PFARM",
+    extraArgs      => "--clientClass=PFARM"
+                      . " --useUpstreamModule",
+    osClasses      => ["FEDORANEXT"],
+  },
   vdoPerfTests => {
     displayName => "VDO_Perf_Tests",
     suiteName   => "nightlyVDOPerfTests",
--- a/src/perl/vdotest/VDOTest.pm
+++ b/src/perl/vdotest/VDOTest.pm
@@ -147,7 +147,7 @@ our %PROPERTIES
      # @ple The directory to put the user tool binaries in
      userBinaryDir           => undef,
      # @ple Use the dmlinux src rpm for testing.
-     useUpstreamKernel       => 0,
+     useDmLinuxModule        => 0,
     );
 
 my @SRPM_NAMES
@@ -489,7 +489,7 @@ sub getTGZNameForVersion {
 sub listSharedFiles {
   my ($self) = assertNumArgs(1, @_);
   my @files = ($self->SUPER::listSharedFiles(), @SHARED_FILES);
-  if ($self->{useUpstreamKernel}) {
+  if ($self->{useDmLinuxModule}) {
     return (@files, @UPSTREAM_NAMES);
   } elsif ($self->{useDistribution}) {
     return (@files, @RPM_NAMES);
--- a/src/perl/vdotest/VDOTest.pm
+++ b/src/perl/vdotest/VDOTest.pm
@@ -148,6 +148,8 @@ our %PROPERTIES
      userBinaryDir           => undef,
      # @ple Use the dmlinux src rpm for testing.
      useDmLinuxModule        => 0,
+     # @ple Use the installed kernel module, dm-vdo
+     useUpstreamModule       => 0,
     );
 
 my @SRPM_NAMES
--- a/src/perl/dmtest/DMTest.pm
+++ b/src/perl/dmtest/DMTest.pm
@@ -101,7 +101,7 @@ our %PROPERTIES
      # @ple The directory to put the user tool binaries in
      userBinaryDir          => undef,
      # @ple Use the dmlinux src rpm for testing.
-     useUpstreamKernel      => 0,
+     useDmLinuxModule       => 0,
     );
 
 my @SRPM_NAMES
@@ -388,7 +388,7 @@ sub installModules {
 sub listSharedFiles {
   my ($self) = assertNumArgs(1, @_);
   my @files = ($self->SUPER::listSharedFiles(), @SHARED_FILES);
-  if ($self->{useUpstreamKernel}) {
+  if ($self->{useDmLinuxModule}) {
     return (@files, @UPSTREAM_NAMES);
   } elsif ($self->{useDistribution}) {
     return (@files, @RPM_NAMES);
--- a/src/perl/dmtest/Makefile
+++ b/src/perl/dmtest/Makefile
@@ -13,7 +13,7 @@ ifdef LOGDIR
 endif
 
 ifeq ($(DMLINUX), 1)
-  TEST_ARGS += --useUpstreamKernel
+  TEST_ARGS += --useDmLinuxModule
 endif
 
 ifeq ($(USER),continuous)
--- a/src/perl/vdotest/Makefile
+++ b/src/perl/vdotest/Makefile
@@ -33,7 +33,7 @@ else
 endif
 
 ifeq ($(DMLINUX), 1)
-  TEST_ARGS += --useUpstreamKernel
+  TEST_ARGS += --useDmLinuxModule
 endif
 
 .PHONY: jenkins


src/perl/vdotest/VDOTest.pm
```diff
@@ -148,6 +148,8 @@ our %PROPERTIES
      userBinaryDir           => undef,
      # @ple Use the dmlinux src rpm for testing.
      useUpstreamKernel       => 0,
+     # @ple Use the installed kernel module, dm-vdo
+     useLinuxNextModule      => 0,
     );
```
lorelei-sakai, line 153: useLinuxNextModule is a misleading name; the in-kernel dm-vdo module does not have to come from linux-next. Instead it will generally be the normal upstream module.
I really think useUpstreamModule is a more correct name for this property. You will probably also want to rename the existing useUpstreamKernel property to avoid confusion; it should properly be called something like useDmLinuxModule.

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -252,7 +256,9 @@ sub configure {
   $self->{moduleVersion} //= $self->{vdoModuleVersion};
   $self->{moduleVersion} //= $VDO_MARKETING_VERSION;
 
-  if ($self->{useDistribution}) {
+  # If useLinuxNextModule, vdo utilities are installed
+  # in /usr/bin/
+  if ($self->{useDistribution} || $self->{useLinuxNextModule}) {
```
lorelei-sakai, line 261: This comment seems unnecessary.

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -600,20 +609,26 @@ sub installModule {
                                   modName         => $moduleName,
                                   modVersion      => $version,
                                   useDistribution => $self->{useDistribution},
+                                  useDnf          => $self->{useLinuxNextModule},
                                  );
```
lorelei-sakai, line 613: There is no reason to change the flag name here.
The new flag should have a name something like useUpstream or useBuiltin. The kernel module side does not use dnf so it's not a good name for the generic case.

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -600,20 +609,26 @@ sub installModule {
...
   $module->load();
   $self->{_modules}{$machineName}{$moduleName} = $module;
   $self->_setModuleParameters();
 
-  $log->info("Installing module $VDO_USER_MODNAME $version on $machineName");
+  my $modVersion = $self->{useLinuxNextModule} ? $VDO_LINUXNEXT_MODVER
+                                               : $self->getModuleVersion();
+
   my $userModule
```
lorelei-sakai, line 621: In the upsteam case you're trying to implement, there is no module version number separate from the ernelversion. The argument is entirely meaningless in that case.

This means there is no reason to define VDO_LINUXNEXT_MODVER because the value should never be used for anything.

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -600,20 +609,26 @@ sub installModule {
...
                                 modName         => $VDO_USER_MODNAME,
-                                modVersion      => $version,
+                                modVersion      => $self->{useLinuxNextModule}
+                                                    ? $VDO_LINUXNEXT_MODVER
+                                                    : $self->getModuleVersion(),
                                 useDistribution => $self->{useDistribution},
-                               );
+                                useDnf          => $self->{useLinuxNextModule},
+                                );
   $userModule->load();
```
lorelei-sakai, line 632: Same comments as above, just pointing you'll need to adjust these arguments also.

src/perl/Permabit/KernelModule.pm
```diff
@@ -51,7 +51,7 @@ sub load {
   my $machine = $self->{machine};
   my $modName = $self->{modName};
 
-  if ($self->SUPER::load()) {
+  if ($self->{useDnf} || $self->SUPER::load()) {
     # modprobe will tell us whether we managed to load successfully.
```
lorelei-sakai, line 55: What is this line even for?  SUPER::load() is there to tell us whether the requeseted kernel module has already been loaded or not, so we don't try to re-laod a module that is already loaded. This doesn't make sense to me.

src/perl/Permabit/UserModule.pm
```diff
@@ -51,6 +51,14 @@ sub loadFromBinaryRPM {
   my $supportName = "$modFileName-support";
   my $supportFile = ($self->{useDistribution}) ? makeFullPath($self->{modDir}, "$supportName-*.rpm")
                                                : makeFullPath($topdir, "$supportName-*.rpm");
+   # if useDnf(useLinuxNextModule), download vdo package from distribution repo
+   if ($self->{useDnf}) {
+    my $getFromDnf = join(' ',
+                          "dnf", "download", "--destdir", "$topdir",
+		          "$modFileName-support"
+                          );
```
lorelei-sakai, line 59: $modFileName-support is alredy defined as $supprotName, just use that. Alos, don't have ); on a line by itself for a function call.

src/perl/Permabit/KernelModule.pm
```diff
@@ -63,11 +63,11 @@ sub load {
         $log->debug("kernel log additions:\n$additions");
       }
     };
+    $log->info("Using kernel module: " . $modName);
     $self->_step(command => "sudo modprobe $modName $self->{modprobeOption}",
                  cleaner => "sudo modprobe -r $modName",
                  diagnostic => $uponModprobeError);
   }
-
   return (!$machine->runSystemCmd("sudo modinfo $modName"));
```
lorelei-sakai, line 71: Don't remove this line, I like the extra space. (Generally a blank line is good after the end of a block.)

src/perl/Permabit/ModuleBase.pm
```diff
@@ -104,11 +106,22 @@ sub loadFromFiles {
...
+  # if useDnf(useLinuxNextModule), download vdo package from distribution repo
+  if ($useDnf) {
+    my $topdir = makeFullPath($machine->{workDir}, $self->{modVersion});
+    $self->_step(command => "mkdir -p $topdir");
+    my $getFromDnf = join(' ',
+                          "dnf", "download", "--destdir", "$self->{modDir}",
+                          "$modFileName"
+	                 );
+    $self->_step(command => $getFromDnf);
+  }
```
lorelei-sakai, line 124: This block seems specific to the user package. You're not getting the kernel module with dnf, are you? So this probably belongs in UserModule.pm instead.

src/tools/installers/CURRENT_VERSION
```diff
@@ -5,3 +5,5 @@ VDO_VERSION               = 8.4.0.0
 VDO_MARKETING_VERSION     = 8.4
 VDO_PROJECT_CODENAME      = "Potassium"
 VDO_MODNAME               = kvdo
+VDO_LINUXNEXT_MODNAME     = dm-vdo
+VDO_LINUXNEXT_MODVER      = 8.3
```
lorelei-sakai, line 9: Upstream, there is no meaningful version for dm-vdo (apart for the kernel version) so don't define it here.For the moudle name, it should clearly be VDO_UPSTREAM_MODNAME. I'd rather not define it at all in this file,  but I guess we need to put it somewhere.

lorelei-sakai (changes requested):
I am really not following the logic of some of the choices made here. You may need to talk about this directly. I've made comments where I think I see what you're doing.

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -58,6 +58,7 @@ my $VDO_WARMUP               = "src/c++/vdo/bin/vdoWarmup";
 my $VDO_INITIALIZE_BLOCK_MAP = "src/c++/vdo/bin/vdoInitializeBlockMap";
 
 my $VDO_USER_MODNAME = "vdo";
+my $VDO_LINUXNEXT_MODNAME = "dm-vdo";
```
lorelei-sakai, line 61: Do not use 'linuxnext' in any name. It is not accurate.
Use 'upstream'.

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -177,6 +178,8 @@ our %BLOCKDEVICE_INHERITED_PROPERTIES
      useDistribution           => 0,
      # The directory to put the user tool binaries in
      userBinaryDir             => undef,
+     # Whether to use dm-vdo
+     useUpstreamModule            => 0,
```
lorelei-sakai, line 182: => alignment is off.

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -589,30 +592,37 @@ sub installModule {
       $self->uninstallModule($machineName);
     }
   }
-
+  # If useUpstreamModule is set, we need to use $VDO_LINUXNEXT_MODNAME(dm-vdo)
+  # module otherwise we load the $VDO_MODNAME(kvdo)
+  if ($self->{useUpstreamModule}) {
+    $moduleName = $VDO_LINUXNEXT_MODNAME;
+  }
```
lorelei-sakai, line 599: There is a getModuleName() function. This logic should be in that function.

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -589,30 +592,37 @@ sub installModule {
...
   # Build the kernel and user tool modules from source.
-  $log->info("Installing module $moduleName $version on $machineName");
```
lorelei-sakai, line 594: Why did you remove this log line?

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -589,30 +592,37 @@ sub installModule {
...
   my $module
     = Permabit::KernelModule->new(
                                   machine         => $self->getMachine(),
                                   modDir          => $self->getModuleSourceDir(),
                                   modFileName     => "kmod-$moduleName",
                                   modName         => $moduleName,
                                   modVersion      => $version,
+                                  upstreamModName => $VDO_LINUXNEXT_MODNAME,
```
lorelei-sakai, line 608: I don't understand what this upstreaModName argument is for. We already have modName...

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -589,30 +592,37 @@ sub installModule {
...
                                   useDistribution => $self->{useDistribution},
+                                  useUpstream     => $self->{useUpstreamModule},
                                  );
   $module->load();
   $self->{_modules}{$machineName}{$moduleName} = $module;
   $self->_setModuleParameters();
 
-  $log->info("Installing module $VDO_USER_MODNAME $version on $machineName");
```
lorelei-sakai, line 608: Again, why remove this line?

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -589,30 +592,37 @@ sub installModule {
...
   my $userModule
     = Permabit::UserModule->new(
                                 machine         => $self->getMachine(),
                                 modDir          => $self->getModuleSourceDir(),
                                 modName         => $VDO_USER_MODNAME,
-                                modVersion      => $version,
+                                modVersion      => $self->{useUpstreamModule}
+                                                     ? ""
+                                                     : $self->getModuleVersion(),
```
lorelei-sakai, line 623: Again, this logic should be put in getModuleVersion().
Hm, looking at how we use the version number in the hash, the upstream case should probably have a special tag like "upstream" to differentiate it from the module we build ourselves.

Actually, for the user module, don't you need to pass a real version number?

src/perl/Permabit/KernelModule.pm
```diff
@@ -78,8 +79,16 @@ sub loadFromFiles {
...
   my $modVer = $self->{modVersion};
-
-  my $loaded = $self->SUPER::loadFromFiles();
+  my $loaded = 0;
+
+  # With useUpstream test, we don't need to load upstream
+  # module(dm-vdo), since it is in the kernel already.
+  if ($self->{useUpstream} &&
+      $modName eq $self->{upstreamModName}) {
+    $loaded = 1;
```
lorelei-sakai, line 88: In this case, you could just "return 1;" here. I think it simplifies the logic a little.  Also, I don't think you need to check the modName. We should only be setting useUpsream for the dm-vdo module.

src/perl/Permabit/ModuleBase.pm
```diff
@@ -42,6 +42,8 @@ my %PROPERTIES
      modVersion      => undef,
      # @ple whether the test is loading the module from a released RPM
      useDistribution => undef,
+     # @ple whether the test is loading the dm-vdo module from a released RPM
```
lorelei-sakai, line 45: This comment is not accurate.

src/perl/Permabit/ModuleBase.pm
```diff
@@ -177,6 +179,9 @@ sub buildModule {
   $self->_step(command => "mkdir -p $topdir");
   $self->_step(command => "cp -up $modulePath $topdir");
 
+  if ($self->{useDnf}) {
+    return
+  }
```
lorelei-sakai, line 184: Did you forget to remove this? I don't see useDnf being defined any more.

src/perl/Permabit/UserModule.pm
```diff
@@ -37,6 +37,27 @@ sub load {
...
+  my $modFileName = $self->{modFileName};
+
+  # if useUpstream, download vdo package from distribution repo
+  if ($self->{useUpstream}) {
+    my $topdir = makeFullPath($machine->{workDir}, $self->{modVersion});
+    $self->_step(command => "mkdir -p $topdir");
+    my $getFromDnf = join(' ',
+                          "dnf", "download", "--destdir", "$self->{modDir}",
+                          "$modFileName"
+	                 );
```
lorelei-sakai, line 55: The alignment is weird in theis function call. Wrap the lines at a reasonable length, and don't put ); on its own line.

src/perl/Permabit/UserModule.pm
lorelei-sakai: Here also, format the join call like a normal function.

src/perl/vdotest/VDOTest.pm
lorelei-sakai: This is still true.

Renaming useUpstreamKernel is probably best done as a separate commit, but it should still be in this PR.

lorelei-sakai (changes requested):
Generally much better. The logic mostly makes sense.

Still several minor style and naming issues.  Also I thought renaming useUpstreamKernel was going to be part of this PR.

src/perl/vdotest/VDOTest.pm
```diff
@@ -489,7 +491,7 @@ sub getTGZNameForVersion {
 sub listSharedFiles {
   my ($self) = assertNumArgs(1, @_);
   my @files = ($self->SUPER::listSharedFiles(), @SHARED_FILES);
-  if ($self->{useUpstreamKernel}) {
+  if ($self->{useDmLinixModule}) {
```
lorelei-sakai, line 494: This is a typo: the new name is useDmLinuxModule.  You can test this by running the dmtest perl tests; they only take a minute or two right now and they use this option so you can know it's working.

src/perl/Permabit/UserModule.pm
```diff
@@ -51,6 +72,13 @@ sub loadFromBinaryRPM {
   my $supportName = "$modFileName-support";
   my $supportFile = ($self->{useDistribution}) ? makeFullPath($self->{modDir}, "$supportName-*.rpm")
                                                : makeFullPath($topdir, "$supportName-*.rpm");
+   # if useUpstreamModule, download vdo package from distribution repo
+   if ($self->{useUpstream}) {
+     my $getFromDnf = join(' ',
+                           "dnf", "download", "--destdir", "$topdir",
+                           "$supportName");
+     $self->_step(command => $getFromDnf);
+  }
```
lorelei-sakai, line 81: Thinking about this now, I think the download of the support module probably belongs in loadFromFiles() next to the vdo package download. The main advantage to doing that is that if someone needs to change this later, it'll be easier to see that they need to change two things, since they'll be next to each other.

src/perl/Permabit/ModuleBase.pm
```diff
@@ -177,6 +179,9 @@ sub buildModule {
   $self->_step(command => "mkdir -p $topdir");
   $self->_step(command => "cp -up $modulePath $topdir");
 
+  if ($self->{useUpstream}) {
+    return
+  }
```
lorelei-sakai, line 184: Wait, if useUpstream is set, do we even call this function? Do we need this check here?  (If we do call into this, shouldn't be break out before you do any build steps? It should be close t othe top of the function.)

Also, if for some reason we do need this, please add a semicolon after "return".

EDIT: It looks like buildModule() only gets called from loadForSourceRPM(), so the useUpstream flag should prevent us from ever getting here.

src/perl/Permabit/KernelModule.pm
```diff
@@ -78,8 +79,15 @@ sub loadFromFiles {
...
-  my $loaded = $self->SUPER::loadFromFiles();
+  my $loaded = 0;
+
+  # With useUpstream test, we don't need to load upstream
+  # module(dm-vdo), since it is in the kernel already.
+  if ($self->{useUpstream}) {
+    return 1;
+  } else {
+    $loaded = $self->SUPER::loadFromFiles();
+  }
```
lorelei-sakai, line 90: The point of use return 1; here is that you don't need the "else" branch. $loaded can be defined just like it used to be.

lorelei-sakai (changes requested):
Aside from the separate comments, there's one more big thing:
I kind of wonder of we should have a completely separate method, called something like loadFromUpstream(), and can keep this new load pathway completely separate from the rpm pathways. (The new method would be parallel to, say, loadFromSourceRPM()

I'm not actually sure if that would make things clearer, or if it means awkward duplication or overlap.  So i think I'm fine if you decide not to do that at the moment.

src/perl/Permabit/ModuleBase.pm
```diff
@@ -119,6 +121,13 @@ sub loadFromFiles {
     return 1;
   }
 
+  # If useUpstream module is set, we should have loaded from
+  # loadFromBinaryRPM.
+  if ($self->{useUpstream}) {
+    $log->debug("useUpstream flag set, loadFromBinaryRPM should have succeed");
+    return 0;
+  }
+
```
lorelei-sakai, line 130: Why did you add this block? It seems like it should be unnecessary. (i.e., if the load didn't work, we also won't find an SRPM for dm-vdo so the next step will also fail.)

src/perl/Permabit/KernelModule.pm
```diff
@@ -78,8 +79,16 @@ sub loadFromFiles {
...
+
+  # With useUpstream test, we don't need to load upstream
+  # module(dm-vdo), since it is in the kernel already.
+  if ($self->{useUpstream}) {
+    return 1;
+  }
+
+  $loaded = $self->SUPER::loadFromFiles();
 
-  my $loaded = $self->SUPER::loadFromFiles();
```
lorelei-sakai, line 82: This line can be reverted instead of changing where $loaded is defined.

src/perl/nightly/NightlyBuildType/VDO.pm
```diff
@@ -70,6 +70,14 @@ my $SUITE_PROPERTIES = {
     extraArgs    => "--clientClass=PFARM",
     osClasses    => ["FEDORANEXT"],
   },
+  vdoUpStreamTests => {
+    displayName    => "VDO_dm_vdo_module_Tests",
```
lorelei-sakai, line 74: A minor thing, I'd spell this DM_VDO_Module_Tests. the capitalization looks odd the way it is.

I assume you're tried running tests this way and they work?

lorelei-sakai (changes requested):
Generally looks really good. here are still a couple of oddities that  could be a bit cleaner, or that I don't understand.

src/perl/Permabit/ModuleBase.pm
C2Redhat: This block has been removed

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -578,7 +581,16 @@ sub installModule {
...
   my $machineName = $self->getMachineName();
   my $moduleName = $self->getModuleName();
+
+  if ($self->{useUpstreamModule}) {
+    my $getVerCmd = "yum list $VDO_USER_MODNAME.`uname -m` | " .
+                    "awk '/^$VDO_USER_MODNAME/ {print \$2}'";
+    my @ver = split(/\./,
+		    runCommand($self->getMachineName(), $getVerCmd)->{stdout});
+    $self->setModuleVersion("$ver[0].$ver[1]");
+  }
```
lorelei-sakai, line 591: This code looks basically reasonable, but I think it ends up looking cleaner if this moves inside of getModuleVersion().

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -1366,6 +1380,20 @@ sub check {
...
+  my ($self) = assertNumArgs(1, @_);
+
+  # If useUpstreamModule is set, we need to use $VDO_UPSTREAM_MODNAME(dm-vdo)
+  # module otherwise we load the $VDO_MODNAME(kvdo)
+  if ($self->{useUpstreamModule}) {
+    return $VDO_UPSTREAM_MODNAME;
+  }
+  return $self->SUPER::getModuleName();
+}
+
```
lorelei-sakai, line 1396: This looks right, but I would move this function up the file so that it is next to getModuleVersion() (around line 537).

src/perl/Permabit/KernelModule.pm
```diff
@@ -79,7 +80,14 @@ sub loadFromFiles {
...
   my $modVer = $self->{modVersion};
 
+  # With useUpstream test, we don't need to load upstream
+  # module(dm-vdo), since it is in the kernel already.
+  if ($self->{useUpstream}) {
+    return 1;
+  }
+
   my $loaded = $self->SUPER::loadFromFiles();
+
```
lorelei-sakai, line 90: I'd remove this blank line but it's not a big deal.

src/perl/Permabit/UserModule.pm
```diff
@@ -37,6 +37,27 @@ sub load {
...
 
+###############################################################################
+# @inherit
+##
+sub loadFromFiles {
+  my ($self) = assertNumArgs(1, @_);
+  my $machine = $self->{machine};
+  my $modFileName = $self->{modFileName};
+
+  # if useUpstream, download vdo package from distribution repo
```
lorelei-sakai, line 48: A minor point, but you don't really need a comment that just says what the code does. It's only really useful there's something no obvious from the code, such as why we need to do a thing.

src/perl/Permabit/UserModule.pm
lorelei-sakai: This comment still applies; can you move this download step next to the other one?

lorelei-sakai (changes requested):
One small note about the commit messages: The title are not sentences. They shoudl be lowercase, no punctuation (besides the colon). Also I would be a bit more specific and start them with "dm vdo test:" instead of just "dm vdo:".

Other comments are mostly minor, just some polish. This basically looks right now.

lorelei-sakai (approved):
Sorry for the delays. This version looks good.

PR #189, opened 2024-09-18
[VDO-5820][RHEL-57824] Fix unaligned discard handling

lorelei-sakai:
Due to RHEL-57824, we noticed that unaligned discard bios are not always handled properly, causing VDO to throw assertions and to not properly discard data. Two general types of fixes are required:

First, some splitting logic was incorrect, leading VDO to incorrectly try to acknowledge and clean up the discard vio without doing processing the final segment. This would cause problem for discards that are 4K or less, but which crossed a VDO block boundary.

Second, large discard bios recycle the same data_vio for each subsequent 4K segment, but the data_vio was not being completely reset before processing the segments after the first one. By luck, this only caused problems for a final partial segment, but it was technically incorrect for all segments.

Due to the subtlety of some of these scenarios, this PR also adds a new unit test for various unaligned discard operations.

### DIFF
--- a/src/c++/vdo/base/data-vio.c
+++ b/src/c++/vdo/base/data-vio.c
@@ -532,6 +532,7 @@ static void launch_data_vio(struct data_vio *data_vio, logical_block_number_t lb
 
 	memset(&data_vio->record_name, 0, sizeof(data_vio->record_name));
 	memset(&data_vio->duplicate, 0, sizeof(data_vio->duplicate));
+	vdo_reset_completion(&data_vio->decrement_completion);
 	vdo_reset_completion(completion);
 	completion->error_handler = handle_data_vio_error;
 	set_data_vio_logical_callback(data_vio, attempt_logical_block_lock);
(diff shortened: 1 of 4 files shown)


src/c++/vdo/base/data-vio.c
```diff
@@ -1320,12 +1321,14 @@ static void clean_hash_lock(struct vdo_completion *completion)
 static void finish_cleanup(struct data_vio *data_vio)
 {
 	struct vdo_completion *completion = &data_vio->vio.completion;
+	u32 discard_size = min_t(u32, data_vio->remaining_discard,
```
bjohnsto, line 1324: Looks like there are two uses of VDO_BLOCK_SIZE - data_vio->offset. Might be good to make one local variable that explains in the name its purpose.

bjohnsto (changes requested):
Looks good in this commit, apart from my one comment.

bjohnsto (approved):
I think we should make this change for sure, though I'm still not sure where it shows up in the queue limits in sysfs. It might just be an internal thing used in calculations and cannot be changed in sysfs or seen.

raeburn (approved):

PR #194, opened 2024-09-24
dm-vdo: remove u64 alignment requirement in murmurhash3 computations

slegendr:
This addresses dm-vdo/vdo [issue #69](https://github.com/dm-vdo/vdo/issues/69) by cleaning up the code to remove the alignment requirement, as there is no need for it.


lorelei-sakai (changes requested):
Just one minot comment: The title of the commit message is very long. I would change it to "dm-vdo murmurhash: remove u64 alignment requirement".

We don't need to include "in murmurhash3 computations" because that is apparent from the files in the diff.

lorelei-sakai (approved):
Thanks. Looks good.

PR #195, opened 2024-09-30
dm vdo: stop using blk_limits_io_{min,opt}

C2Redhat:
Implement change from author, Christoph Hellwig <hch@lst.de>:

  dm: stop using blk_limits_io_{min,opt}
  https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/commit/drivers/md/dm-vdo?id=0a94a469a4f02bdcc223517fd578810ffc21c548

  Remove use of the blk_limits_io_{min,opt} and assign the values directly
  to the queue_limits structure.  For the io_opt this is a completely
  mechanical change, for io_min it removes flooring the limit to the
  physical and logical block size in the particular caller.  But as
  blk_validate_limits will do the same later when actually applying the
  limits, there still is no change in overall behavior.

  Signed-off-by: Christoph Hellwig <hch@lst.de>
  Signed-off-by: Mikulas Patocka <mpatocka@redhat.com>

Also remove VDO_USE_ALTERNATE in dm-vdo-target.c to use
limits->max_hw_discard_sector by default since the change has been
integrated in all the latest releases.

### DIFF
--- a/src/c++/devices/tracer/dmTracer.c
+++ b/src/c++/devices/tracer/dmTracer.c
@@ -318,9 +318,9 @@ static void tracerIoHints(struct dm_target *ti, struct queue_limits *limits)
   limits->physical_block_size = SECTOR_SIZE;
 
   // The minimum io size for random io
-  blk_limits_io_min(limits, sectorCount * SECTOR_SIZE);
+  limits->io_min = VDO_BLOCK_SIZE;
   // The optimal io size for streamed/sequential io
-  blk_limits_io_opt(limits, VDO_BLOCK_SIZE);
+  limits->io_opt = VDO_BLOCK_SIZE;
 
   // Discard hints
   limits->max_discard_sectors = VDO_SECTORS_PER_BLOCK;
--- a/src/c++/vdo/base/dm-vdo-target.c
+++ b/src/c++/vdo/base/dm-vdo-target.c
@@ -1069,9 +1069,9 @@ static void vdo_io_hints(struct dm_target *ti, struct queue_limits *limits)
 	limits->physical_block_size = VDO_BLOCK_SIZE;
 
 	/* The minimum io size for random io */
-	blk_limits_io_min(limits, VDO_BLOCK_SIZE);
+	limits->io_min = VDO_BLOCK_SIZE;
 	/* The optimal io size for streamed/sequential io */
-	blk_limits_io_opt(limits, VDO_BLOCK_SIZE);
+	limits->io_opt = VDO_BLOCK_SIZE;
 
 	/*
 	 * Sets the maximum discard size that will be passed into VDO. This value comes from a


slegendr:
It would be helpful to have a working link to the original commit from upstream.

src/c++/devices/tracer/dmTracer.c
```diff
@@ -19,6 +19,18 @@
...
 #include "traceLoggerBlockTrace.h"
 
+#ifndef VDO_UPSTREAM
+#undef VDO_USE_ALTERNATE
+#if defined(RHEL_RELEASE_CODE) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
+#if (RHEL_RELEASE_CODE < RHEL_RELEASE_VERSION(10, 0))
+#define VDO_USE_ALTERNATE
+#endif
+#else /* !RHEL_RELEASE_CODE */
+#if (LINUX_VERSION_CODE < KERNEL_VERSION(6, 10, 0))
```
slegendr, line 29: I don't think this is correct. Functions blk_limits_io_{min,opt} exist for the 6.10 and 6.11 kernels. It's 6.12 that they do not exist for. Seems like this should be if LINUX_VERSION_CODE < KERNEL_VERSION(6, 12, 0).

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -1068,10 +1068,30 @@ static void vdo_io_hints(struct dm_target *ti, struct queue_limits *limits)
...
 	limits->physical_block_size = VDO_BLOCK_SIZE;
 
+#ifndef VDO_UPSTREAM
+#undef VDO_USE_ALTERNATE
+#if defined(RHEL_RELEASE_CODE) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
+#if (RHEL_RELEASE_CODE < RHEL_RELEASE_VERSION(10, 0))
+#define VDO_USE_ALTERNATE
+#endif
+#else /* !RHEL_RELEASE_CODE */
+#if (LINUX_VERSION_CODE < KERNEL_VERSION(6, 10, 0))
```
slegendr, line 1078: Same comment here as for dmTracer.c. I think this should specify kernel version 6.12 in the comparison.

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -1086,18 +1106,6 @@ static void vdo_io_hints(struct dm_target *ti, struct queue_limits *limits)
...
-#if (RHEL_RELEASE_CODE < RHEL_RELEASE_VERSION(10, 0))
-#define VDO_USE_ALTERNATE
-#endif
-#else /* !RHEL_RELEASE_CODE */
-#if (LINUX_VERSION_CODE < KERNEL_VERSION(6, 10, 0))
-#define VDO_USE_ALTERNATE
-#endif
-#endif /* !RHEL_RELEASE_CODE */
-#endif /* !VDO_UPSTREAM */
 #ifdef VDO_USE_ALTERNATE
```
slegendr, line 1109: This will need to be adjusted to align with the kernel version modification in how VDO_USE_ALTERNATE is defined. However, I do not believe that we build vdo-devel against anything with less than a 6.10 kernel, so this VDO_USE_ALTERNATE block may not be needed.

slegendr (dismissed):

slegendr:
One question I have as I was looking at the upstream commit  - do we need the VDO_USE_ALTERNATE for this at all? It looks like the changes to use limits->io_{min,opt} directly were applied for kernel 6.11. However, the io_min and io_opt fields of the queue_limits structure already existed in 6.10. Maybe we just directly set it and avoid needing USE_ALTERNATE entirely?

src/c++/devices/tracer/dmTracer.c
```diff
@@ -19,6 +19,18 @@
...
+#if defined(RHEL_RELEASE_CODE) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
+#if (RHEL_RELEASE_CODE < RHEL_RELEASE_VERSION(10, 0))
+#define VDO_USE_ALTERNATE
+#endif
+#else /* !RHEL_RELEASE_CODE */
+#if (LINUX_VERSION_CODE < KERNEL_VERSION(6, 12, 0))
+#define VDO_USE_ALTERNATE
+#endif
+#endif /* !RHEL_RELEASE_CODE */
+#endif /* !VDO_UPSTREAM */
```
lorelei-sakai, line 33: It only just occurred to me, but we normally put this block next to the code where VDO_USE_ALERNATE is used. This should move down to around line 330.

lorelei-sakai, line 33: Second comment, this is a test-only file so it doesn't need the VDO_UPSTREAM parts. This code will never go upstream.

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -1068,10 +1068,30 @@ static void vdo_io_hints(struct dm_target *ti, struct queue_limits *limits)
...
 	limits->physical_block_size = VDO_BLOCK_SIZE;
 
+#ifndef VDO_UPSTREAM
+#undef VDO_USE_ALTERNATE
+#if defined(RHEL_RELEASE_CODE) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
+#if (RHEL_RELEASE_CODE < RHEL_RELEASE_VERSION(10, 0))
+#define VDO_USE_ALTERNATE
+#endif
+#else /* !RHEL_RELEASE_CODE */
+#if (LINUX_VERSION_CODE < KERNEL_VERSION(6, 12, 0))
```
lorelei-sakai, line 1078: This code actually got pulled into 6.11 as far as I can tell. So this boundary is not quite right.

But Susan made a good point that the limits structure has been there for a while, and we can probably set this unilaterally, without using VDO_USE_ALTERNATE.

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -1086,25 +1106,8 @@ static void vdo_io_hints(struct dm_target *ti, struct queue_limits *limits)
...
-#else /* !RHEL_RELEASE_CODE */
-#if (LINUX_VERSION_CODE < KERNEL_VERSION(6, 10, 0))
-#define VDO_USE_ALTERNATE
-#endif
-#endif /* !RHEL_RELEASE_CODE */
-#endif /* !VDO_UPSTREAM */
-#ifdef VDO_USE_ALTERNATE
-	limits->max_discard_sectors =
-		(vdo->device_config->max_discard_blocks * VDO_SECTORS_PER_BLOCK);
-#else
```
lorelei-sakai, line 1104: Removing this is fine and good, but you should mention it in the commit message.

lorelei-sakai (changes requested):

src/c++/devices/tracer/dmTracer.c
C2Redhat: I have removed the VDO_USE_ALTERNATE block completely

src/c++/vdo/base/dm-vdo-target.c
C2Redhat: I removed the block completely since it is not necessary anymore

src/c++/vdo/base/dm-vdo-target.c
lorelei-sakai: Unfortunately I figured out a reason to keep this VDO_USE_ALTERNATE block. Although all on pfarms are 6.10 or higher, so might have some older F39 and F40 machines (like my build machine) which are not as caught up on the kernels. I don't really want to have to constantly update my working machines, so I'd prefer the VDO_USE_ALTERNATE block for 6.10 remains.

We can ask the group on Monday for other opinions, too.

lorelei-sakai (changes requested):
Changed my mind, please keep the VDO_USE_ALTERNATE block for max_hw_discard_sectors.

lorelei-sakai (approved):
This looks good now.

PR #199, opened 2024-10-08
[VDO-5816] dm vdo user: Fix errx calls to use 1 for error code.

bjohnsto:
Error codes used in calls to errx should only use specific values. Previously we used to return our own error codes in a number of places which was incorrect. Replace those values with 1, which is the generic exit error code.


lorelei-sakai (approved):
Looks good to me.

One very minor thing, the commit title would need to be lowercase and no punctuation if it were going upstream. (It's not a sentence, so no capital, no period.) But since this is userspace only, it doesn't matter so much.

PR #200, opened 2024-10-10
dm vdo: add support for unaligned.h location change

C2Redhat:
unaligned.h file has moved from generated/asm to tools/linux in the kernel include src tree. Add in VDO_USE_NEXT to reflect this change.

### DIFF
--- a/src/c++/uds/src/uds/murmurhash3.c
+++ b/src/c++/uds/src/uds/murmurhash3.c
@@ -8,7 +8,24 @@
 
 #include "murmurhash3.h"
 
+#ifndef VDO_UPSTREAM
+#include <linux/version.h>
+#undef VDO_USE_NEXT
+#if defined(RHEL_RELEASE_CODE) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
+#if (RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(10, 0))
+#define VDO_USE_NEXT
+#endif
+#else /* !RHEL_RELEASE_CODE */
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 12, 0))
+#define VDO_USE_NEXT
+#endif
+#endif /* !RHEL_RELEASE_CODE */
+#endif /* !VDO_UPSTREAM */
+#ifndef VDO_USE_NEXT
 #include <asm/unaligned.h>
+#else
+#include <linux/unaligned.h>
+#endif
 
 static inline u64 rotl64(u64 x, s8 r)
 {
--- a/src/c++/uds/src/uds/numeric.h
+++ b/src/c++/uds/src/uds/numeric.h
@@ -6,7 +6,24 @@
 #ifndef UDS_NUMERIC_H
 #define UDS_NUMERIC_H
 
+#ifndef VDO_UPSTREAM
+#include <linux/version.h>
+#undef VDO_USE_NEXT
+#if defined(RHEL_RELEASE_CODE) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
+#if (RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(10, 0))
+#define VDO_USE_NEXT
+#endif
+#else /* !RHEL_RELEASE_CODE */
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 12, 0))
+#define VDO_USE_NEXT
+#endif
+#endif /* !RHEL_RELEASE_CODE */
+#endif /* !VDO_UPSTREAM */
+#ifndef VDO_USE_NEXT
 #include <asm/unaligned.h>
+#else
+#include <linux/unaligned.h>
+#endif
 #ifdef __KERNEL__
 #include <linux/kernel.h>
 #endif
--- a/src/c++/uds/userLinux/uds/linux/unaligned.h
+++ b/src/c++/uds/userLinux/uds/linux/unaligned.h
@@ -1,0 +1,124 @@
+/*
+ * %COPYRIGHT%
+ *
+ * %LICENSE%
+ */
+
+#ifndef LINUX_UNALIGNED_H
+#define LINUX_UNALIGNED_H
+
+#include <asm/byteorder.h>
+#include <linux/types.h>
+
+/* Type safe comparison macros, similar to the ones in linux/minmax.h. */
+
+/*
+ * If pointers to types are comparable (without dereferencing them and
+ * potentially causing side effects) then types are the same.
+ */
+ #define __typecheck(x, y) \
+	(!!(sizeof((typeof(x) *)1 == (typeof(y) *)1)))
+
+/* 
+ * Hack for VDO to replace use of the kernel's __is_constexpr() in __cmp_ macros.
+ * VDO cannot use __is_constexpr() due to it relying on a GCC extension to allow sizeof(void).
+ */
+#define __constcheck(x, y) \
+	(__builtin_constant_p(x) && __builtin_constant_p(y))
+
+/* It takes two levels of macro expansion to compose the unique temp names. */
+#define ___PASTE(a,b) a##b
+#define __PASTE(a,b) ___PASTE(a,b)
+#define __UNIQUE_ID(prefix) __PASTE(__PASTE(__UNIQUE_ID_, prefix), __COUNTER__)
+
+/* Defined in linux/minmax.h */
+#define __cmp_op_min <
+#define __cmp_op_max >
+
+#define __cmp(op, x, y)	((x) __cmp_op_##op (y) ? (x) : (y))
+
+#define __cmp_once(op, x, y, unique_x, unique_y) \
+	__extension__({                          \
+		typeof(x) unique_x = (x);        \
+		typeof(y) unique_y = (y);        \
+		__cmp(op, unique_x, unique_y);   \
+	})
+
+#define __careful_cmp(op, x, y)                            \
+	__builtin_choose_expr(                             \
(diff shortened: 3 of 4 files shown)


src/c++/uds/src/uds/murmurhash3.c
```diff
@@ -8,7 +8,24 @@
...
+#ifndef VDO_UPSTREAM
+#include <linux/version.h>
+#undef VDO_USE_ALTERNATE
+#if defined(RHEL_RELEASE_CODE) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
+#if (RHEL_RELEASE_CODE > RHEL_RELEASE_VERSION(10, 0))
+#define VDO_USE_ALTERNATE
+#endif
+#else /* !RHEL_RELEASE_CODE */
+#if (LINUX_VERSION_CODE > KERNEL_VERSION(6, 11, 0))
+#define VDO_USE_ALTERNATE
```
raeburn, line 20: I'm not sure if it matters, but in other places the USE_ALTERNATE type macros are always used for the older-than-current versions of kernel APIs, and the code branch with no ALTERNATE macro defined is what is used for the latest kernel sources. It looks like you're seting up VDO_USE_ALTERNATE to be used for the newest kernels.

src/c++/uds/src/uds/numeric.h
```diff
@@ -6,7 +6,24 @@
 #ifndef UDS_NUMERIC_H
 #define UDS_NUMERIC_H
 
+#ifndef VDO_UPSTREAM
+#include <linux/version.h>
+#undef VDO_USE_ALTERNATE
+#if defined(RHEL_RELEASE_CODE) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
+#if (RHEL_RELEASE_CODE > RHEL_RELEASE_VERSION(9, 0))
```
raeburn, line 13: The RHEL_RELEASE_VERSION checks aren't consistent throughout your patch. There are checks for > 9.0 and for > 10.0. Is this correct?

src/c++/uds/userLinux/uds/linux/unaligned.h
```diff
@@ -0,0 +1,124 @@
+/*
+ * %COPYRIGHT%
```
raeburn, line 2: If this is a copy of asm/unaligned.h because the kernel moved the file, please make one or the other in our tree just #include the other, so we don't risk the two getting out of sync. I would lean towards making the new name have the real content and the old name #include it.

src/c++/uds/userLinux/uds/linux/unaligned.h
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * %LICENSE%
+ */
+
+#ifndef ASM_UNALIGNED_H
+#define ASM_UNALIGNED_H
```
raeburn, line 8: If the new name is linux/unaligned.h, the protective macro name should be adjusted.

src/c++/vdo/fake/linux/bio.h
```diff
@@ -11,7 +11,22 @@
 #define __LINUX_BIO_H
 
 /* struct bio, bio_vec and BIO_* flags are defined in blk_types.h */
+#include <linux/version.h>
+#undef VDO_USE_ALTERNATE
```
raeburn, line 15: fake/linux/bio.h is only used for user-mode builds, isn't it? Won't both versions from userLinux/uds be available? Maybe for this file we can just switch to the new name and not conditionalize it?

src/c++/uds/userLinux/uds/linux/unaligned.h
lorelei-sakai: The file could see code changes in the future, meaning it would have different content for different versions. The old content should not change, but we could have the new file simply include the old one.  Then the new file could be changed to incorporate new content without messing up older builds. (In practice, I think this is unlikely to matter.)

raeburn (dismissed):
see above comments

src/c++/uds/userLinux/uds/linux/unaligned.h
raeburn: "The file could see code changes in the future..." That's true with or without the renaming anyway.

"The old content should not change..." We need to stay compatible, but if it simplifies our driver code, why not?

src/c++/uds/src/uds/murmurhash3.c
```diff
@@ -8,7 +8,24 @@
...
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 11, 0))
+#define VDO_USE_NEXT
+#endif
+#endif /* !RHEL_RELEASE_CODE */
+#endif /* !VDO_UPSTREAM */
+#ifdef VDO_USE_NEXT
+#include <linux/unaligned.h>
+#else
 #include <asm/unaligned.h>
+#endif
```
lorelei-sakai, line 28: Please follow the model we used before for VDO_USE_NEXT code: place the newer code before the older code, which also means the condition will be "#ifndef VDO_USE_NEXT"

src/c++/uds/src/uds/murmurhash3.c
```diff
@@ -8,7 +8,24 @@
...
 
+#ifndef VDO_UPSTREAM
+#include <linux/version.h>
+#undef VDO_USE_NEXT
+#if defined(RHEL_RELEASE_CODE) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
+#if (RHEL_RELEASE_CODE > RHEL_RELEASE_VERSION(10, 0))
+#define VDO_USE_NEXT
+#endif
+#else /* !RHEL_RELEASE_CODE */
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 11, 0))
```
lorelei-sakai, line 19: This version check is not right. This change actually shows up in 6.12-rc2. It should definitely not be used for any kernel version less than 6.12. (Hopefully we won't ever have to check the EXTRA_VERSION to support 6.12-rc1.)

src/c++/uds/src/uds/murmurhash3.c
```diff
@@ -8,7 +8,24 @@
 
 #include "murmurhash3.h"
 
+#ifndef VDO_UPSTREAM
+#include <linux/version.h>
+#undef VDO_USE_NEXT
+#if defined(RHEL_RELEASE_CODE) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
+#if (RHEL_RELEASE_CODE > RHEL_RELEASE_VERSION(10, 0))
```
lorelei-sakai, line 15: I suspect the RHEL_RELEASE_CODE handling is also not correct, but we can address that later when (as I expect) this code gets pulled into RAWHIDE. (But it should probably be >= 10.0 since this will get pulled into the RHEL 10 streams once it's released.)

src/c++/uds/src/uds/numeric.h
lorelei-sakai: As Ken says, whatever checks we settle on should be used consistently in all the places you're touching.

src/c++/uds/src/uds/numeric.h
```diff
@@ -6,7 +6,24 @@
...
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 11, 0))
+#define VDO_USE_NEXT
+#endif
+#endif /* !RHEL_RELEASE_CODE */
+#endif /* !VDO_UPSTREAM */
+#ifdef VDO_USE_NEXT
+#include <linux/unaligned.h>
+#else
 #include <asm/unaligned.h>
+#endif
```
lorelei-sakai, line 26: I would like to alphabetize this properly, with linux/unaligned.h  appearing below linux/types.h. 
However we should probably fix that up separately because the upstream patch was script-generated and replaced the line without moving it. (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=5f60d5f6bbc12e782fac78110b0ee62698f3b576)

For clarity, this is a comment that requires no code change.

src/c++/uds/userLinux/uds/linux/unaligned.h
lorelei-sakai: I actually thought of something else. For manifest purposes, we'll want to include one copy of this file but not both. That's easier if the two files are not linked.

I mean, I'll admit we could just include both in the manifest but that seems like an awkward thing for our userspace package to do, with one file simply being a redirect to the other.  Though I suppose we could unilaterally use the new location (or the old one) in userspace, the kernel linkage isn't that important...

src/packaging/src-dist/MANIFEST.yaml
```diff
@@ -191,6 +191,7 @@ tarballs:
             - mutex.h
             - random.h
             - types.h
+            - unaligned.h
```
lorelei-sakai, line 194: MANIFEST changes depend on what we do per my previous comment. But I really do not want to be including both files in the manifest.

lorelei-sakai (changes requested):
Please compare to the previous use of VDO_USE_NEXT in repair.c and slab-depot.c, and be careful about details.

lorelei-sakai (changes requested):
The commit message still refers to VDO_USE_ALTERNATE.
(Also the title should be all lowercase.)

The actual changes look good.

src/c++/vdo/fake/linux/bio.h
lorelei-sakai: Turns out this doesn't work. We have to conditionalize so that we don't end up pulling both versions when a file includes both numeric.h and bio.h (which is a lot of files).

lorelei-sakai (approved):
Looks good now. Thanks!

PR #201, opened 2024-10-11
[VDO-5815] dm vdo user: Fix assertion failures in vdoformat

bjohnsto:
[VDO-5815] The libuds library that vdoformat was using was built so that it would abort on assertion failures. This should only be happening in uds and vdo unit tests. To fix this, we make sure the NDEBUG flag is always set on libuds. Then inside albtest.c and vdotest.c we set the abort on assert failure flag to true at the start of each test and set it back to its original value at the end.

In addition, we want to improve on the error handling in vdoformat when we get UDS_ASSERTION_FAILURE errors. So we change the error result to VDO_BAD_CONFIGURATION since it is more useful and we add an addition bit of error text to point the users at the actual assertion failure which is written into the logs.

### DIFF
--- a/src/c++/uds/userLinux/build/Makefile
+++ b/src/c++/uds/userLinux/build/Makefile
@@ -15,7 +15,7 @@ vpath %.c $(SRC_UDS_DIR)
 # Flags for linking the shared library itself.
 SHLIBFLAGS = -lm -ldl -pthread -lrt $(shell getconf LFS_LDFLAGS)
 
-CFLAGS 	  = $(PLATFORM_CFLAGS) -I$(SRC_UDS_DIR)
+CFLAGS 	  = $(PLATFORM_CFLAGS) -DNDEBUG -I$(SRC_UDS_DIR)
 LDFLAGS   = $(GLOBAL_LDFLAGS)
 LDSHFLAGS = -shared -Wl,--unresolved-symbols=ignore-in-shared-libs
 
@@ -76,7 +76,7 @@ $(PROFDIR)/libuds.a: $(addprefix $(PROFDIR)/,$(UDS_OBJECTS))
 
 $(PROFDIR)/%.o: %.c
 	@mkdir -p $(@D)
-	$(CC) -pg -DNDEBUG $(CFLAGS) $< -c -o $@
+	$(CC) -pg $(CFLAGS) $< -c -o $@
 
 ########################################################################
 # Dependency processing
--- a/src/c++/uds/userLinux/tests/albtest.c
+++ b/src/c++/uds/userLinux/tests/albtest.c
@@ -499,6 +499,8 @@ int main(int argc, char **argv)
   umask(0);
   open_vdo_logger();
 
+  set_exit_on_assertion_failure(true);
+
   testArgv = calloc(argc - optind, sizeof(char *));
 
   // load all the tests and copy their suite information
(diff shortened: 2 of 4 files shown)


src/c++/uds/userLinux/build/Makefile
```diff
@@ -15,7 +15,7 @@ vpath %.c $(SRC_UDS_DIR)
 # Flags for linking the shared library itself.
 SHLIBFLAGS = -lm -ldl -pthread -lrt $(shell getconf LFS_LDFLAGS)
 
-CFLAGS 	  = $(PLATFORM_CFLAGS) -I$(SRC_UDS_DIR)
+CFLAGS 	  = $(PLATFORM_CFLAGS) -DNDEBUG -I$(SRC_UDS_DIR)
```
lorelei-sakai, line 18: HM, if we're building with NDEBUG all the time now, maybe it makes sense to rename the token and switch the sense of ? Make the flag false be default in the code, and then let a dev set -DDEBUG if they want to use debug mode?  If you don't want to do that now, we can clean that up later.

src/c++/uds/userLinux/tests/albtest.c
```diff
@@ -300,6 +301,7 @@ TestResult runTest(const CU_SuiteInfo *suite, const CU_TestInfo *test)
       int status;
       int tmp = waitpid(pid, &status, 0);
       if (tmp == -1) {
+        set_exit_on_assertion_failure(previous);
```
lorelei-sakai, line 304: I think you are doing more than you need here. There's no need to unset the flag, unit tests expect it to be on all the time anyway.

src/c++/uds/userLinux/tests/albtest.c
```diff
@@ -289,6 +289,7 @@ TestResult runTest(const CU_SuiteInfo *suite, const CU_TestInfo *test)
     .tests = 1,
   };
 
+  bool previous = set_exit_on_assertion_failure(true);
```
lorelei-sakai, line 292: This works, but I think it would be more efficient to set it once near the start of the program rather than setting it for each test.Say, in main() right before figuring out what test modules to load.

(The flag is a global static, so it'll persist as long as the program is running, I think.)

lorelei-sakai (changes requested):
We don't need to do so much work in albtest, you can simplify that. Otherwise, generally looks good.

lorelei-sakai:
One more thing, I find it very helpful to have the Jira ticket ID in the PR title, not just in the description. (It makes it easier to find things when I go to see what's in a release, for instance.) I can do the edit myself but I also wanted to remind you.

src/c++/uds/userLinux/build/Makefile
bjohnsto: I think it would be better to clean this up later.

src/c++/uds/userLinux/tests/albtest.c
bjohnsto: ok i followed your idea of moving the call to main()

src/c++/uds/userLinux/tests/albtest.c
bjohnsto: moved to main()

lorelei-sakai (approved):
The changes look good now.

In the commit message, I would prefer we get in the habit of downcasing the commit message title, and not including ticket numbers in the commit messages (they should be in the PR description, instead). But since this change isn't going to the kernel, it's not vital here.

PR #203, opened 2024-10-15
Add changelog for 8.3.0.72 release

lorelei-sakai:
vdo 8.3.0.72:
* Improved vdoformat error handling.
* Added clang support.
* General code clean up and simplification.

### DIFF
--- a/src/packaging/github/changelog
+++ b/src/packaging/github/changelog
@@ -15,3 +15,9 @@ vdo:
 * Removed obsolete utilities.
 * Fixed version numbering.
 * Updated project documentation.
+
+Version 8.3.0.72
+vdo:
+* Improved vdoformat error handling.
+* Added clang support.
+* General code clean up and simplification.


slegendr (approved):
Do you think it's worth applying this to the stable Fedora releases (F40, F41), or just Rawhide? The general Fedora recommendation is that updates to stable releases should be focused on bug fixes, not features. F41 is entering final freeze today, so if you want to apply it to that, the update can't be done until after release in November. At the same time, F39 goes EOL on November 19th, so I don't think it's worth updating F39. Updates to Rawhide and F40 can happen now.

lorelei-sakai:
I think RAWHIDE is enough, we're not in a hurry. Though getting it into F41 would be nice. (I could argue that most of the changes are simple fixes, though the Makefile change is arguably a new feature, and has a very slight change of breaking someone's existing build pattern.)

PR #205, opened 2024-10-17
dm vdo: Update migration initialScenario to X86_FEDORA40_HEAD

C2Redhat:
X86_RHEL9_head as a scenario is no longer a valid choice since Chlorine. Add X86_FEDORA40_HEAD in SupportedScenarios.yaml so it can be used as a scenario for migration tests.


raeburn (approved):
Typo in the PR title ("FORDORA") but otherwise looks good to me

PR #210, opened 2024-11-08
Fix VDOFormat.pm regexp

bjohnsto:
In the tryIllegal function in VDOFormat.pm, we pass in an error string parameter that is used to search through the error output of calls to vdoformat. The calls to vdoformat are done through a function called assertExecuteCommand which calls the perl function confess when an error occurs. Confess attaches a stack trace to the error output which means the search would always succeed because part of the stack trace would contain the error string we're looking for in the tryIllegal parameter part of the stack trace.

The fix for this issue is to be more restrictive in the search criteria so that it only looks at the actual output from vdoformat.


lorelei-sakai (approved):
The commit message should have a proper title. Otherwise, looks good.

PR #211, opened 2024-11-08
dm vdo: Disable FEDORA39 testing and replace head with FEDORA40

C2Redhat:
[TECH 181] Disable FEDORA39 from testing and replace any build reference from FEDORA39 to FEDORA40.


lorelei-sakai (approved):
I'm tempted to say we should leave the aliases in until we add F41, but I guess there's no real reason to keep them.

Looks good.

PR #213, opened 2024-11-13
[VDO-5853] Alias kvdo to dm-vdo for most testing

lorelei-sakai:
Creating an alias means that anything attempting to modprobe the dm-vdo module will load he kvdo module instead. This should prevent tests from ever loading the built-in dm-vdo moduleinstead of our test module. The exception is useUpstreamModule tests, which are explicitly intended to test the built-in dm-vdo module.

This may prevent the problems described in VDO-5852 as well.

### DIFF
--- a/src/perl/vdotest/VDOTest.pm
+++ b/src/perl/vdotest/VDOTest.pm
@@ -47,6 +47,7 @@ use Permabit::SystemUtils qw(
   assertCommand
   assertSystem
   createRemoteFile
+  runCommand
 );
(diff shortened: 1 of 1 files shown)


raeburn:
Would it not be better to put the alias file in the package we make, so package removal (by perl test or by checkserver) will get rid of it automatically?

lorelei-sakai:
This problem was addressed in a a better way by PR #214, so this commit is unnecessary.

PR #214, opened 2024-11-14
[VDO-5852] [VDO-5853] load kvdo module after reboot

C2Redhat:
[VDO-5852] [VDO-5853] With latest lvm, 2.03.25, lvm will load upstream vdo module(dm_vdo) if our test module kvdo module is not loaded. Add code to load the desire module before we lvm is called so the proper module is loaded.


lorelei-sakai (approved):
Looks good.

PR #221, opened 2024-12-09
[VDO-5864] dm vdo test: exclude Create03 test

C2Redhat:
Create03 uses sysfs but the upstream kernel module does not load sysfs so this test will fail. Disable Create03 from upstreamTests suite. Once sysfs test is removed in general, we can restore this test.

### DIFF
--- a/src/perl/vdotest/vdotests.suites
+++ b/src/perl/vdotest/vdotests.suites
@@ -818,7 +818,6 @@ $suiteNames{debugKernelTests}
 $suiteNames{upstreamTests}
   = [
      @{$suiteNames{checkin}},	# A bunch of basic VDO operation tests.
-     "Create03",		# Test that many restarts works
      "DiscardPerf",     	# Test that normal discards work. Longer.
      "Discard512",		# Test that 512 byte discards work
      "Dmsetup",			# Test that dmsetup interface works


lorelei-sakai (changes requested):
I do not tthink this i the right approach. The sysfs operations will fail on any test run with an upstream kernel, including mainlin-next and rawhide.

Instead of this, we should update the listin the "pstreamTests" suire to exclude Create03. This is the reason we created a named suite in the first place: so that we had a list that could be changed as required.

C2Redhat:
> I do not tthink this i the right approach. The sysfs operations will fail on any test run with an upstream kernel, including mainlin-next and rawhide.
> 
> Instead of this, we should update the listin the "pstreamTests" suire to exclude Create03. This is the reason we created a named suite in the first place: so that we had a list that could be changed as required.

I choose this approach since Bruce is working on disable sysfs.  We are not seeing the same problem in RAWHIDE test.  This only happen in FEDORANEXT system using --useUpstreamModule.

lorelei-sakai:
I think it is worth clarifying that even if you only want to exclude this test from one suite, the right approach is to make a new named suite and use that. The --exclude option is intended for interactive use (i.e., developers running tests directly), and should never be used in our automation.  Define a vdotests suite that includes the tests we want to run, and use that instead.

C2Redhat:
> I think it is worth clarifying that even if you only want to exclude this test from one suite, the right approach is to make a new named suite and use that. The --exclude option is intended for interactive use (i.e., developers running tests directly), and should never be used in our automation. Define a vdotests suite that includes the tests we want to run, and use that instead.

OK.  Since rawhide and upstream test are using the same suite, I think the best way is to create a new suite for upstream kernel test.  Do you have objection on this approach?

lorelei-sakai (approved):
Looks good.

PR #224, opened 2025-01-08
dm vdo: fix build problems with latest Fedora 40 and 41, 6.12.X

C2Redhat:
With the latest Fedora 6.12.X, we run into two problems:
. make fails in packaging/github due to unaligned.h file location changed from asm to linux
. kvdo.spec file is incorrect to include BUILD_DEPEND statement with LZ4_COMPRESS and LZ4_DECOMPRESS

This pull request will address both problems.


rhawalsh:
> Looks like src/c++/vdo/kernel/Makefile tries to put BUILD_DEPENDS into its generated DKMS_CONF too. It's got a slightly different syntax without the '[0]' bit. Should that be removed too?

I believe it should be removed.

Our implementation of the BUILD_DEPENDS was erroneous.

From the man page:
```
       BUILD_DEPENDS[#]=
              This  optional  directive  is an array that allows you to specify other modules as dependencies for your module. Each array element should be
              the PACKAGE_NAME of another module that is managed by dkms. Do not specify a version or architecture in the dependency. Note that this direc‐
              tive is only advisory; missing or broken dependencies cause non-fatal warni
```

Notably, "Each array element should be the PACKAGE_NAME of another module that is *managed by dkms*."

In addition to that, the notion of the brackets on the BUILD_DEPENDS line was also wrong.  The dkms script sources the dkms.conf (`. dkms.conf`) as part of the process, so by setting BUILD_DEPENDS[0]=LZ4_COMPRESS, and then BUILD_DEPENDS[0]=LZ4_DECOMPRESS, the first one is actually getting overwritten.  So the same effect would be when using it without the [] spelling.

This change is because with dkms-3.1.4, the script *actually verifies* the BUILD_DEPENDS.  Previously, it was never used, at least as best I can tell upon inspection and execution with dkms-3.1.3.

lorelei-sakai (changes requested):
There is a typo in the commit message for the second commit ("BUILD_DEPNEDS")

Otherwise it all looks good to me.

raeburn:
If fixing the unaligned.h issue is going to be included here instead of a separate PR, then the title and description of this PR need fixing.

raeburn (approved):

lorelei-sakai (approved):

PR #225, opened 2025-01-10
dm vdo: Remove VDO_USE_NEXT macro in bio.h

C2Redhat:
This is related to the change as in pull request 224: https://github.com/dm-vdo/vdo-devel/pull/224

We missed removing VDO_USE_NEXT macro in bio.h in the pull request 224 since "make" finished without problem. However, we still need to remove the VDO_USE_NEXT macro and only include <linux/unaligned.h> in bio.h.


raeburn:
The change title has a double underscore between VDO and USE. Otherwise looks good.

src/c++/vdo/fake/linux/bio.h
```diff
@@ -11,22 +11,7 @@
...
-#endif
-#else /* !RHEL_RELEASE_CODE */
-#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 12, 0))
-#define VDO_USE_NEXT
-#endif
-#endif /* !RHEL_RELEASE_CODE */
-#ifndef VDO_USE_NEXT
-#include <asm/unaligned.h>
-#else
 #include <linux/unaligned.h>
```
lorelei-sakai, line 14: Please more this include downbetween minmax.h and version.h. (This is our internal file so we don't need to be concerned with upstream churn.)

lorelei-sakai (changes requested):
In addition to reordering the #include line, this PR should also remove the old fake asm/unaligned.h file.

lorelei-sakai (commented):
Looks good.

raeburn (approved):

PR #226, opened 2025-01-10
dm vdo: use a short static string for thread name prefix

lorelei-sakai:
This fix was inspired by the upstream conversation with John Garry. This eliminates a warning for snprintf possibly truncating the MODULE_NAME string (even though in practice it's never that long), but also there's no reason for us to be setting a local value for the MODULE_NAME, and a six-byte prefix string is longer than we want anyway.

### DIFF
--- a/src/c++/vdo/base/vdo.c
+++ b/src/c++/vdo/base/vdo.c
@@ -159,12 +159,6 @@ static void finish_vdo_request_queue(void *ptr)
 #define MAX_VDO_WORK_QUEUE_NAME_LEN 16
 #endif /* __KERNEL__ */
 
-#ifdef MODULE
-#define MODULE_NAME THIS_MODULE->name
-#else
-#define MODULE_NAME "dm-vdo"
-#endif  /* MODULE */
-
 static const struct vdo_work_queue_type default_queue_type = {
 #ifdef __KERNEL__
 	.start = start_vdo_request_queue,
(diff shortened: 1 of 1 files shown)


sweettea:
It would be nice to have a link to the upstream discussion in the commit message.

src/c++/vdo/base/vdo.c
```diff
@@ -578,7 +572,7 @@ int vdo_make(unsigned int instance, struct device_config *config, char **reason,
 	*vdo_ptr = vdo;
 
 	snprintf(vdo->thread_name_prefix, sizeof(vdo->thread_name_prefix),
-		 "%s%u", MODULE_NAME, instance);
+		 "%s%u", VDO_LOGGING_MODULE_NAME, instance);
```
raeburn, line 575: Isn't VDO_LOGGING_MODULE_NAME going to be "dm: vdo" in the kernel?

lorelei-sakai: You're right, I guess I wasn't reading carefully. I'll just go with the bare string, then, instead of trying to be fancy.

lorelei-sakai:
> It would be nice to have a link to the upstream discussion in the commit message. Might also drop a 'Suggested-by: John Garry' tag?

I was planing to add a Suggested-by tag; I might as well do it now instead of at the point where we send it upstream. I'll also add a description of the warning it fixes.

raeburn (approved):

PR #230, opened 2025-01-23
[VDO-5868,RHEL-12834] Speed up scanning of loaded slab refcount blocks

raeburn:


src/c++/vdo/tests/ptr-u32-map.h
```diff
@@ -0,0 +1,66 @@
+/*
+ * %COPYRIGHT%
+ *
+ * %LICENSE%
+ *
+ * $Id$
+ */
+
+/*
```
lorelei-sakai, line 9: Do we really need an entirely new type and file to accomplish this? It kind of feels like excess wrapping.

src/c++/vdo/tests/RefCounts_t3.c
```diff
@@ -0,0 +1,184 @@
+/*
+ * %COPYRIGHT%
```
lorelei-sakai, line 2: It feels a bit weird to add one test before the main code changes, and one test after. I suppose it doesn' matter much, but this is just a correctness tests, right? So it should be able to go in before the refcount changes, right? Anyway, just felt weird.

(Also weird that RefCount_t3 is added after RefCount_t4, but if you rename the latter that problem solves itself.)

src/c++/vdo/base/slab-depot.c
```diff
@@ -2225,11 +2281,8 @@ static void unpack_reference_block(struct packed_reference_block *packed,
...
 	}
 
-	block->allocated_count = 0;
-	for (index = 0; index < COUNTS_PER_BLOCK; index++) {
-		if (counters[index] != EMPTY_REFERENCE_COUNT)
-			block->allocated_count++;
-	}
+	empty_count = fix_provisional_and_count_empty_refcounts(counters, COUNTS_PER_BLOCK);
+	block->allocated_count = COUNTS_PER_BLOCK - empty_count;
 }
```
lorelei-sakai, line 2286: Why not just have the function return the number of allocated blocks which is what we actually want? Seems like a more intuitive interface.

src/c++/vdo/base/slab-depot.c
```diff
@@ -2167,28 +2168,83 @@ static void dirty_all_reference_blocks(struct vdo_slab *slab)
...
-		(first.entry_count == second.entry_count));
+	u8 *bytes = counters;
+	unsigned int empty_count = 0;
+
+	/* Sanity check assumption used for optimizing this code. */
+	BUILD_BUG_ON(sizeof(vdo_refcount_t) != 1);
+
+	/* Check any unaligned bytes at start. */
+	while (length > 0 && !IS_ALIGNED((unsigned long)bytes, sizeof(u64))) {
+		if (*bytes == PROVISIONAL_REFERENCE_COUNT)
```
lorelei-sakai, line 2200: Shouldn't our refcounts start aligned to begin with?

Also, even if you think this check is necessary at the start (maybe), is it also necessary at the end. We'll end at the end of the sector or block, which is going to already be aligned, yes?

src/c++/vdo/base/slab-depot.c
```diff
@@ -2167,28 +2168,83 @@ static void dirty_all_reference_blocks(struct vdo_slab *slab)
...
+		if (*bytes == EMPTY_REFERENCE_COUNT)
+			empty_count++;
+		bytes++;
+		length--;
+	}
+	/* Same thing but do 8 bytes at a time. */
+	while (length >= sizeof(u64)) {
+		unsigned int length_this_loop = min_t(unsigned int, 254, length);
+		unsigned int word_count = length_this_loop / sizeof(u64);
+		u64 split_count = 0;	/* 8 byte-size counters */
```
lorelei-sakai, line 2211: Why do 254 words per loop? Why not pick a number that divides evenly by 8 bytes, like 248? It's not hat many more loop ends and it saves on having to realign things.

src/c++/vdo/base/slab-depot.c
```diff
@@ -2167,28 +2168,83 @@ static void dirty_all_reference_blocks(struct vdo_slab *slab)
...
+
+		while (word_count--) {
+			u64 word = *(u64 *)bytes;
```
lorelei-sakai, line 2214: So in general this math needs more commenting. I tried to read it for a while but I'm still not sure I understand why (or if) it does what you claim it does.

src/c++/vdo/base/slab-depot.c
```diff
@@ -2167,28 +2168,83 @@ static void dirty_all_reference_blocks(struct vdo_slab *slab)
...
 /**
- * clear_provisional_references() - Clear the provisional reference counts from a reference block.
- * @block: The block to clear.
+ * match_bytes() - Check an 8-byte word for bytes matching the value specified
+ *
+ * On return each byte of the word is 0x80 if the input byte matched, 0x00 otherwise.
  */
-static void clear_provisional_references(struct reference_block *block)
+static inline u64 match_bytes(u64 input, u8 match)
 {
```
lorelei-sakai, line 2184: It looks like everywhere we use match_bytes, we shift the results down by 7. Why not just do that before returning from the function?

src/c++/vdo/base/slab-depot.c
```diff
@@ -2167,28 +2168,83 @@ static void dirty_all_reference_blocks(struct vdo_slab *slab)
...
-	vdo_refcount_t *counters = get_reference_counters_for_block(block);
-	block_count_t j;
+	u64 temp = input ^ (match * 0x0101010101010101ULL);
```
lorelei-sakai, line 2185: This probably wants an explanation. Why does this work?
An explanation in the PR might be enough, maybe. But I'm worried this code is easy to break because it's hard to check tha it still works after changes.

src/c++/vdo/base/slab-depot.c
```diff
@@ -2167,28 +2168,83 @@ static void dirty_all_reference_blocks(struct vdo_slab *slab)
...
+		unsigned int word_count = length_this_loop / sizeof(u64);
+		u64 split_count = 0;	/* 8 byte-size counters */
+
+		while (word_count--) {
+			u64 word = *(u64 *)bytes;
+			u64 temp;
+
+			/* First, if we have any provisional refcount values, fix them. */
+			temp = match_bytes(word, PROVISIONAL_REFERENCE_COUNT);
+			if (temp) {
```
lorelei-sakai, line 2219: Might be worth adding a flag for whether to clear provisional reference counts; on a clean load there won't be any and we can save some work. But we can also leave that for a future optimization.

lorelei-sakai (changes requested):
Sort of a rough cut at a review but I called out some of the big things I saw.

In particular I did not look or think too hard about the content of the tests yet. In general I do not like randomness in tests (because it can lead to failure that are difficult to reproduce) but I haven't looked carefuly at how you're using it.

src/c++/vdo/base/slab-depot.h
```diff
@@ -627,6 +627,7 @@ void adjust_slab_journal_block_reference(struct slab_journal *journal,
 enum reference_status __must_check reference_count_to_status(vdo_refcount_t count);
 struct reference_block * __must_check get_reference_block(struct vdo_slab *slab,
 							  slab_block_number index);
+void finish_reference_block_load(struct vdo_completion *completion);
```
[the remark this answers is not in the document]
raeburn: It's consistent with the various things around it ... defined with STATIC, declared inside the INTERNAL block. If we need an ifdef for this one, I think we need it for a bunch of others too, if not the whole block.

src/c++/vdo/tests/RefCounts_t4.c
```diff
@@ -0,0 +1,271 @@
+/*
```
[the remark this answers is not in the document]
raeburn: Good idea, I'll rename it to _x and maybe bump the iteration count a little for hopefully more consistent timing.

src/c++/vdo/base/slab-depot.h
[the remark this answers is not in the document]
lorelei-sakai: Mea culpa. I did not expand the context enough to see the larger enclosing block.  So I agree it's right. (And INTERNAL is the correct parker, not __KERNEL__, I messed up that too.)

src/c++/vdo/tests/ptr-u32-map.h
raeburn: It was initially used in both new tests, but _t3 got stripped down to the point where it wasn't needed any more. Since it's only called in a couple places, yeah, I can throw out the new type and do the conversions inline.

src/c++/vdo/tests/RefCounts_t3.c
raeburn: The correctness test looks at what might be edge cases with the new code, because of the funny counter manipulation. The results should be the same with the old code, but the cases it exercises aren't interesting ones to look at for the old implementation.

I put the performance test in first because it made sense to have a point in the sequence with the performance test and the old implementation together.

src/c++/vdo/tests/RefCounts_t4.c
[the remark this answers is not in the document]
raeburn: The _p tests need their own main() and initialization and all that. The _x and _t tests can use the existing framework for unit tests. So for a test needing a VDO device, it was expedient to use _t (or _x).

src/c++/vdo/base/slab-depot.c
raeburn: It's a byte array inside a packed struct. Chasing through various type definitions, it looks like the byte array offset inside the allocated buffer should be a multiple of 8 bytes, and I expect the allocated buffer to start on an aligned address. But for a function that just processes an array of bytes, I'm reluctant to rely on those, at least without any sort of sanity check. (A BUILD_BUG_ON for the size would be easy, but the alignment would mean a run-time check -- either BUG_ON or an error path.)

src/c++/vdo/base/slab-depot.c
raeburn: The "% 255" trick imposes the limit of 254 for the inner loop, and the nature of the parallel processing of the bytes imposes the multiple-of-8 constraint. I thought "pick a number that divides evenly by 8 bytes" was expressed pretty well in the calculations there, but maybe "length_this_loop" should've been called "max_length_this_iteration", since it's word_count that actually controls the number of bytes to be processed.

src/c++/vdo/base/slab-depot.c
raeburn: Yeah, it took me a couple iterations of reading to understand it.

The idea is to get out an indicator in each returned byte of whether the corresponding input byte matches a specified value.

The match*0x0101... gives us a constant with the desired value in each byte, and the XOR converts "which bytes match X?" to "which bytes are zero?" (At a micro-optimization level, the multiplication will be done at compile time, and if we're searching for zero bytes in the first place, the XOR can be skipped.)

The check is sort of split into two pieces -- we check the top bit of each byte, and we check the lower 7 bits of each byte, and make sure both pieces are zero.

"~temp" gives us bytes with the top bit set iff the top bit of the temp value (being tested for zero) is clear; masking with 0x8080... throws away the other bits, leaving only those top-bit flags.

"temp & 0x7f7f..." gives us each byte with the low 7 bits of the temp value, 0 to 0x7f in each byte, and subtracting from 0x8080... gives us 0x80 in each result byte if the temp low-7-bits is 0, or something in the range 0x01-0x7f if low-7-bits is not zero; because of the limited range of each byte's values, there's no borrowing from the next byte over and it can still be looked at as 8 parallel calculations. So, in each byte computed here, the top bit will be set exactly when low-7-bits from temp is 0; the lower computed bits are of no further use.

A bitwise-and of that with the result of checking the top bits ("~temp & 0x8080...", which can only have the top bits of each byte possibly set) gives us 8 output bytes each with the value 0x80 if the corresponding temp byte was 0 and 0x00 otherwise.

So we've transformed "which bytes are X?" to "which bytes are 0?" and then generated 0x80/0x00 in each position depending on whether we found a zero byte, using a half dozen math/logic operations and no conditional branching.

src/c++/vdo/base/slab-depot.c
raeburn: Yeah, looking at it (or merging the memcpy into the loop) might be worth looking at down the road.

src/c++/vdo/base/slab-depot.c
raeburn: BTW I just did an experiment without the nested loop, doing the "empty_count += split_count % 255" after scanning each word, making the code a little easier to understand. That makes the perf test run around 10% slower on my laptop compared to the nested-loop version.

src/c++/vdo/base/slab-depot.c
raeburn: A quick test on my laptop indicates dropping the provisional->empty transformation speeds up the perf test by about half. That's hard-coded, not testing a passed-in flag. But I'm wary of making such a change without some thought into what happens if a provisional value is there when we think it shouldn't be.

raeburn:
Also, just noting here in case anyone asks later, I tried a version of the refcount scanning loop that checks each byte individually, dropping the complicated u64 processing, but merges the provisional->empty and counting aspects into one pass. It's slower than the u64 version by about 2.5x on my laptop -- 2.6us/block vs 1. (I expected it to be slower, but didn't have numbers to point at if asked.)

src/c++/vdo/base/slab-depot.c
```diff
@@ -2167,28 +2168,89 @@ static void dirty_all_reference_blocks(struct vdo_slab *slab)
...
+				/* 'temp' has 0x01 bytes where 'word' has PROVISIONAL */
+				word ^= temp * (PROVISIONAL_REFERENCE_COUNT ^ EMPTY_REFERENCE_COUNT);
+				*(u64 *)bytes = word;
+			}
+
+			/* Now count the EMPTY_REFERENCE_COUNT bytes, updating the 8 counters. */
+			temp = match_bytes(word, EMPTY_REFERENCE_COUNT);
+			split_count += temp;
+			bytes += sizeof(u64);
+			bytes_left -= sizeof(u64);
```
lorelei-sakai, line 2245: Why don't we update the byte_left count in the outer loop? The inner loop doesn't care about this.

src/c++/vdo/base/slab-depot.c
```diff
@@ -2167,28 +2168,89 @@ static void dirty_all_reference_blocks(struct vdo_slab *slab)
...
+		/*
+		 * The counter "% 255" trick used below to fold split_count into empty_count
+		 * imposes a limit of 254 bytes examined each iteration of the outer loop.
+		 */
+		unsigned int max_bytes_this_iteration = min_t(unsigned int, 254, bytes_left);
+		/* We process 8 at a time, so round down to at most 248 bytes/31 words. */
+		unsigned int word_count = max_bytes_this_iteration / sizeof(u64);
+
+		while (word_count--) {
+			u64 word = *(u64 *)bytes;
```
lorelei-sakai, line 2230: Can we do this cast at the top? We're really treating the whole refcount array as an array of 8-bytes words, not as individual bytes any more.)

src/c++/vdo/base/slab-depot.c
```diff
@@ -2167,28 +2168,89 @@ static void dirty_all_reference_blocks(struct vdo_slab *slab)
...
+			bytes += sizeof(u64);
+			bytes_left -= sizeof(u64);
+		}
+		/*
+		 * This use of modulus to add the bytes together only works if the sum is kept
+		 * below 255. Hence the 254 limit above.
+		 */
+		empty_count += split_count % 255;
+	}
+	return length - empty_count;
```
lorelei-sakai, line 2253: Length doesn't need to be a local here, I think it's always COUNTS_PER_BLOCK. Also I'd like a blank line before the return line.

src/c++/vdo/base/slab-depot.c
lorelei-sakai: I'll look at this later if we try to split regualr load processing from recovery processing. For now, it doesn't matter.

src/c++/vdo/tests/RefCounts_t3.c
```diff
@@ -0,0 +1,184 @@
...
+ *
+ * %LICENSE%
+ *
+ * $Id$
```
lorelei-sakai, line 6: The $Id$ tag was for perforce and we don't need to add them any more.

src/c++/vdo/tests/RefCounts_t3.c
```diff
@@ -0,0 +1,184 @@
...
+  int i;
+  
+  vio->completion.parent = block;
+  vio->io_size = VDO_BLOCK_SIZE;
+
+  for (i = 0; i < VDO_SECTORS_PER_BLOCK; i++) {
+    packed->sectors[i].commit_point.encoded_point = 0;
+  }
+  /*
+   * Some patterns to examine: Big runs (over 256) of EMPTY and of
```
lorelei-sakai, line 121: Can we move this comment maybe three lines up. The separation of the encoded point initialization from the rest of the initialization is a little confusing.

lorelei-sakai (changes requested):

lorelei-sakai (changes requested):
Okay. I left a lot of comments but it's mostly formatting and fretting about naming.

Generally looks good. In particular I want to say the commit message on the third commit (the one that will go upstream) is very good, succinct but informative.

src/c++/vdo/tests/RefCounts_x1.c
```diff
@@ -0,0 +1,257 @@
...
+  for (int block_number = 0; block_number < BLOCKS_PER_VIO; block_number++) {
+    packed = (struct packed_reference_block *) (vio->data + block_number * VDO_BLOCK_SIZE);
+    for (int sector_number = 0; sector_number < VDO_SECTORS_PER_BLOCK; sector_number++) {
+      struct packed_reference_sector *sector = &packed->sectors[sector_number];
+      for (int count_index = 0; count_index < COUNTS_PER_SECTOR; count_index++) {
+        vdo_refcount_t value;
+        if (mostly_empty && ((random() % 7) != 0)) {
+          value = EMPTY_REFERENCE_COUNT;
+        } else {
+          value = random();
```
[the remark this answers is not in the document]
raeburn: I think it's a couple seconds (on my laptop) to initialize the array. The timed part of the test takes about 15s in the current incarnation so I don't think the initialization is a big deal.

src/c++/vdo/base/slab-depot.c
```diff
@@ -2167,28 +2168,99 @@ static void dirty_all_reference_blocks(struct vdo_slab *slab)
...
+		 * had the target value found in byte 0, etc. We just have to avoid overflow.
+		 */
+		u64 split_count = 0;
+		/*
+		 * The counter "% 255" trick used below to fold split_count into empty_count
+		 * imposes a limit of 254 bytes examined each iteration of the outer loop. We
+		 * process a word at a time, so that limit gets rounded down to 31 u64 words.
+		 */
+		const unsigned int max_words_per_iteration = 254 / sizeof(u64);
+		unsigned int words_left_this_iteration = min_t(unsigned int, words_left,
```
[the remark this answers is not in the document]
raeburn: I'm wary of overloading "count" too much. It's already used for reference count values and for the number of "empty" values found, so I intentionally avoided it.

"iter_words_left"? We've generally tried too stay away from abbreviating things... "inner_words_left"? That sort of asks for "max_words_per_iteration" to be renamed I think.

src/c++/vdo/tests/RefCounts_x1.c
```diff
@@ -0,0 +1,257 @@
...
+  memset(wrapper, 0, sizeof(*wrapper));
+  vdo_initialize_completion(&wrapper->completion, vdo, VDO_TEST_COMPLETION);
+  wrapper->waiter.callback = didAcquireVIO;
+}
+
+/**********************************************************************/
+static void initVIO(struct reference_block *block, struct vio *vio,
+                    unsigned int *allocated_count_ptr)
+{
+  struct packed_reference_block *packed = (struct packed_reference_block *) vio->data;
```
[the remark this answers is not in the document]
raeburn: yep, missed that in rearranging the code, thanks

src/c++/vdo/tests/RefCounts_x1.c
```diff
@@ -0,0 +1,257 @@
...
+
+  // do vio_pool stuff on the right worker thread
+  vdo_initialize_completion(&test_completion, vdo, VDO_TEST_COMPLETION);
+  performAction(doIngest, &test_completion);
+
+  for (i = 0; i < vio_count; i++) {
+    returnVIO(wrappers[i].entry);
+  }
+  pool = NULL;
+  free(wrappers);
```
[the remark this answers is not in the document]
raeburn: whoops, yes we do, fixing

src/c++/vdo/base/slab-depot.c
```diff
@@ -2167,28 +2168,99 @@ static void dirty_all_reference_blocks(struct vdo_slab *slab)
...
+			split_count += temp;
+			words++;
+		}
+		/*
+		 * This use of modulus to add the bytes together only works if the sum is kept
+		 * below 255. Hence the 254 limit above.
+		 */
+		empty_count += split_count % 255;
+	}
+	return COUNTS_PER_BLOCK - empty_count;
```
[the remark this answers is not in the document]
raeburn: fixing

src/c++/vdo/base/slab-depot.c
```diff
@@ -2167,28 +2168,99 @@ static void dirty_all_reference_blocks(struct vdo_slab *slab)
...
+			}
+
+			/* Now count the EMPTY_REFERENCE_COUNT bytes, updating the 8 counters. */
+			temp = match_bytes(word, EMPTY_REFERENCE_COUNT);
+			split_count += temp;
+			words++;
+		}
+		/*
+		 * This use of modulus to add the bytes together only works if the sum is kept
+		 * below 255. Hence the 254 limit above.
```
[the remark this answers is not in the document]
raeburn: Okay, I'm dropping the second comment.

src/c++/vdo/fake/linux/align.h
```diff
@@ -0,0 +1,16 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+/* Copied from kernel include/linux/align.h. */
+#ifndef _LINUX_ALIGN_H
+#define _LINUX_ALIGN_H
+
+#include <linux/const.h>
+
+/* @a is a power of 2 value */
```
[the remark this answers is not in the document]
raeburn: not any more, dropping

src/c++/vdo/tests/RefCounts_x1.c
```diff
@@ -0,0 +1,257 @@
...
+  unsigned long long usage_per_block = cpu_usage * 1000 / blocks_processed;
+  fprintf(stderr,
+          "%u iterations, %u vios of %u blocks: cpu time = %llu.%06llu s, %llu ns per block\n",
+          iteration_count, vio_count, BLOCKS_PER_VIO, cpu_usage / 1000000, cpu_usage % 1000000,
+          usage_per_block);
+  vdo_finish_completion(completion);
+}
+
+/**
+ * Most basic refCounts test.
```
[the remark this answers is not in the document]
raeburn: updated

src/c++/vdo/tests/RefCounts_x1.c
```diff
@@ -0,0 +1,257 @@
...
+  vio->completion.parent = block;
+  vio->io_size = VDO_BLOCK_SIZE * BLOCKS_PER_VIO;
+
+  for (int block_number = 0; block_number < BLOCKS_PER_VIO; block_number++) {
+    packed = (struct packed_reference_block *) (vio->data + block_number * VDO_BLOCK_SIZE);
+    for (int sector_number = 0; sector_number < VDO_SECTORS_PER_BLOCK; sector_number++) {
+      struct packed_reference_sector *sector = &packed->sectors[sector_number];
+      for (int count_index = 0; count_index < COUNTS_PER_SECTOR; count_index++) {
+        vdo_refcount_t value;
+        if (mostly_empty && ((random() % 7) != 0)) {
```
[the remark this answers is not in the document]
raeburn: blank lines added

src/c++/vdo/base/slab-depot.c
```diff
@@ -2167,28 +2168,99 @@ static void dirty_all_reference_blocks(struct vdo_slab *slab)
...
+				 * 'temp' has 0x01 bytes where 'word' has PROVISIONAL; this xor
+				 * will alter just those bytes, changing PROVISIONAL to EMPTY.
+				 */
+				word ^= temp * (PROVISIONAL_REFERENCE_COUNT ^ EMPTY_REFERENCE_COUNT);
+				*words = word;
+			}
+
+			/* Now count the EMPTY_REFERENCE_COUNT bytes, updating the 8 counters. */
+			temp = match_bytes(word, EMPTY_REFERENCE_COUNT);
+			split_count += temp;
```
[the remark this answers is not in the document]
raeburn: combined

src/c++/vdo/tests/RefCounts_x1.c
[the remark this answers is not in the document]
lorelei-sakai: 20 seconds seems like a long time but guess for an _x test it's not much.

lorelei-sakai (approved):
Thanks for your patience going through all these iterations.
This all looks good to me now.

PR #232, opened 2025-01-24
dm vdo: fix latest rawhide build problem

C2Redhat:
This MR fixed two build problems:

. update array initialize to prevent the latest gcc 15.0.1 warning.
     error: initializer-string for array of ‘unsigned char’ is too long
     [-Werror=unterminated-string-initialization]
   
. update USE_VDO_NEXT comparison marco to include
  us_to_ktime() online definition in latest RAWHIDE, 6.13.0 release.

### DIFF
--- a/src/c++/uds/src/uds/time-utils.h
+++ b/src/c++/uds/src/uds/time-utils.h
@@ -21,7 +21,7 @@
 #define VDO_USE_NEXT
 #endif
 #else /* !RHEL_RELEASE_CODE */
-#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 13, 0))
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 14, 0))
 #define VDO_USE_NEXT
 #endif
 #endif /* !RHEL_RELEASE_CODE */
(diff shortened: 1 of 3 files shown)


src/c++/uds/src/uds/time-utils.h
```diff
@@ -53,6 +53,7 @@ static inline ktime_t current_time_us(void)
 {
 	return current_time_ns(CLOCK_REALTIME) / NSEC_PER_USEC;
 }
+
 #ifndef VDO_USE_NEXT
```
lorelei-sakai, line 57: There's no reason to add an extra line here, please don't.

lorelei-sakai (changes requested):
The second commit here (at least the part to index-layout.c) is something we will want to push upstream.  So the commit message should be written with upstream in mind.

I suppose the main thing is that you can't "fix" a "warning". I would say "prevent a warning". (We treat it as an error, but other people won't.)

src/c++/uds/src/uds/time-utils.h
C2Redhat: OK.  Do you want me to separate those two commits?   I can do that if it is cleaner.

lorelei-sakai (approved):
Looks good.

src/c++/uds/src/uds/time-utils.h
C2Redhat: This is fixed.

PR #233, opened 2025-01-24
dm vdo: simplify format and includes

lorelei-sakai:
This is more upstream feedback on the thread name patch, will be squashed upstream into the existing patch.
See PRs 226 and 229.

Simplify the format string, and remove some unnecessary #includes. Actually a good number of the linux #includes seem to not be necessary, but I only removed the ones where I could make a plausible argument to relate them to the module handling.


bjohnsto (approved):
Looks fine to me

PR #236, opened 2025-01-29
[VDO-5594] Allow more iterating to converge on desired memory size.

raeburn:


lorelei-sakai (changes requested):
The change description could stand to use a bit more description about _why_ this change is helpful or appropriate. The current description only repeats what the code change does.

(Also, please use kernel style descriptions , e.g. "dm vdo test: allow more iterations...", even though for a test-only change it's not strictly necessary.

bjohnsto (approved):

PR #216, opened 2024-11-19
[VDO-5752] Add support for in kernel formatting

bjohnsto:
This PR moves the functionality of vdoformat into the kernel. It includes three commits. The first is all the code that will go upstream. The second is support for unit tests and the third is support for perl tests. 

The key parts of this patch include:

1. Adding three new optional parameters to the vdo table line (slab bits, index memory size, and index sparse) from the vdoformat parameter list, along with validation code and default values for each one.

2. Adding a check at dmsetup create time to check the geometry block to see if it is all zeroes. This will inform the formatting code whether it should run or not. Basically if the geometry block is all zeroes, we format the device.

3. The actual formatting code. This is the most detailed part. It will initialize structures with appropriate values so that both the geometry block and super block can be written out with the correct format. The formatting code will also zero out several areas of the device (uds super block, component areas of the device) so no old info is picked up when it shouldn't. The zero'ing code uses a kernel function called blkdev_issue_zeroout.

Along with the upstream changes are additional commits to add testing and user mode support. Those commits come directly after the code changes. The unit test support includes getting blkdev_issue_zeroout to run in user mode. Currently that is a no-op function since the filelayer for the FormatVDO_t* tests will create the storage with all zeroes.

PR #237, opened 2025-01-30
[VDO-STORY-293] dm vdo: add user tool vdoCalculateSize

C2Redhat:
This user tool is used to calculate the following:

Given physical storage size X, and slab size Y, how much space will be available for data storage (plus block map)? How much space will be unused? (Caller can easily compute the minimum amount that needs to be added to allocate another slab.)

For logical size X, how much space will be consumed by a fully allocated block map tree? And how much RAM? (An approximate ratio is easy to give for VDO generally but for a specific logical size we can compute it precisely. The block map size might also be used to choose a block map cache size when starting the device.)

For physical storage size X, slab size Y, and logical size Z, how much space will be available for data storage once the block map tree is fully allocated? For a given index size and sparseness, what’s the size of the dedupe window (or dense and sparse windows both) the index can remember?


rhawalsh (changes requested):
This needs to have a test written for it before we can merge it.

rhawalsh (changes requested):
The description refers to "VDO-STORY-293", but it should be "VDOSTORY-293"

src/perl/vdotest/VDOTest/VDOCalculateSize.pm
```diff
@@ -0,0 +1,48 @@
...
+######################################################################
+# Test the basic execution of vdoCalculatSize.
+##
+sub testBasicExecution {
+  my ($self) = assertNumArgs(1, @_);
+  my $device = $self->getDevice();
+  my $machine = $device->getMachine();
+  my $cmd = $self->findBinary("vdoCalculateSize");
+  my $args = "--slabBits 20 --physicalSize=300G " .
+             "--logicalSize=1T --indexMemorySize=2";
```
rhawalsh, line 36: If time is a problem, then please say so, but I would kind of like to see a couple of different configurations tested.  In particular dense index vs sparse.

If you wouldn't be able to get these additional configurations handled as part of this commit, then I'd like to have a todo ticket created so that we can do it later.

rhawalsh (commented):
We also need to establish a man page for this utility.  Probably easiest to copy another utility and just adapt it.

src/perl/vdotest/VDOTest/VDOCalculateSize.pm
C2Redhat: I can add more conditions to it.  It should be simple to add.

rhawalsh:
I was discussing this with Chung yesterday where he showed me the current version of the output from the tool (displayed below).  I have some thoughts about this that I'd like to get posted so we can discuss them.

When this story was put together, there were several questions that it set out to ensure this tool can answer:

**Required**
- Given physical storage size X, and slab size Y, how much space will be available for data storage (plus block map)? How much space will be unused? (Caller can easily compute the minimum amount that needs to be added to allocate another slab.)
I believe this is answered by looking at [VDO Usage][Usable Size]
- For logical size X, how much space will be consumed by a fully allocated block map tree? And how much RAM? (An approximate ratio is easy to give for VDO generally but for a specific logical size we can compute it precisely. The block map size might also be used to choose a block map cache size when starting the device.)
I don't believe this is directly answered
- For physical storage size X, slab size Y, and logical size Z, how much space will be available for data storage once the block map tree is fully allocated?
I believe this is answered by looking at [VDO meta data][Total BlockMap Pages]
- For a given index size and sparseness, what’s the size of the dedupe window (or dense and sparse windows both) the index can remember? (Okay to just document? I think that’s a simple fixed ratio. But how do duplication and copied index entries affect how much unique data can be remembered?) The dedupe window may necessarily be approximate; round down. May need to account for thread counts.
I believe this is answered by looking at [VDO Usage][Dedup Window]
- For given format parameters X, Y, Z, how much physical space will be required? And how much RAM?
I don't believe the RAM question is answered.
I believe the Physical Size is answered in some way based on arguments provided.  The example I was provided used an argument `--physicalSize=300G`.  It's not clear to me that if you leave out that argument what will happen.

**Optional:**
- For physical storage size X, slab size Y, and expected space savings rate Z, how much logical space should we allocate to correspond to the physical space that will be available? (Computable from above?)
I do not believe this is answered.
- For slab size X, what’re the smallest and largest physical sizes that can be supported? For physical size X, what’re the smallest and largest slab sizes that can be supported? (Must factor in index and other non-slab space. Maybe just generate a table of physical size ranges for all slab sizes?)
I do not believe this is answered.

Current output:
```
Blocksize: 4096
VDO:
  Physical Block: 78643200
  Logical Block: 268435456
  Sparse: 0
  Slab Size: 1048576
  Slab Count: 69
  Index Memory: 2
VDO Constants(4K block):
  Superblock Size: 1
  Geometery Block Size: 1
  VDO Blockmap Tree Root Count: 60
  Slab Summary Size: 64
  Recovery Journal Size: 32768
VDO meta data:
  UDS Index Size: 21469MB
  Total BlockMap Pages: 1292MB
  Slab Reference Count Usage: 70MB
  Slab Jouranl Usage: 60MB
VDO Usage:
  Physical Size: 300GB
  Logical Size: 1024GB
  Usable Size: 277GB
  Dedup Window: 2048GB
Memory Usage:
  Block map cache: 128MB
```

Now that we've gotten the goals posted and whether they've been addressed, I believe one of the purposes of this tool is to allow a user to run the tool and gather some information, but also allow for a program to run the tool and use some of that information to build the stack (e.g. LVM creating the backing devices).

If I were in the position of implementing the usage of this tool for LVM (or something else), I'd much prefer that I can look at one sub-section and get all the numbers I need.  I'd also prefer to be able to have a consistent unit size rather than being provided si units in one place and 4k blocks in another.

Maybe something along these lines (which is the same input for what I was told generated the above output):
```
./vdoCalculateSize --physicalSize=300G --logicalSize=1T --slabBits=20 --indexMemorySize=2
Input Parameters:
  Physical Size: 300G
  Logical Size: 1T
  slabBits: 20
  indexMemorySize: 2
  indexSparse: 0
  <other-input-parameters>
Storage Usage:
  Blockmap: 1292 MB
  UDS Index: 21469 MB
  Slab Reference Count: 70 MB
  Slab Journal: 60 MB
Memory Usage:
  Blockmap Cache: 128 MB
  <other parameters>
Volume Characteristics:
  Slab Count: 69
  Block Size: 4096
  Physical Blocks: 78643200
  Logical Blocks: 268435456
  Slab Size: 1048576 MB
  Slab Count: 69
  Index Memory: 2 GB
  Sparse: 0
```

With this type of output:
- A machine can run this tool and look at the output for "Storage Usage" and get everything it needs.
- The rest of the data is intended for human consumption.
- The input parameters would list out all of the possible input parameters regardless of whether they were specified. 
 Either it'll print the specified value, or it'll give the default that is used for calculations.
- Memory usage, unless we expand to have more than just the size of the Blockmap Cache, then I don't see a lot of value in retaining it.

I would like to see the units be a bit more consistent, even in my suggested output, though.  Though, I'm not sure what I'm asking for is the right thing, other people's opinions would be good to hear.

src/perl/vdotest/VDOTest/VDOCalculateSize.pm
```diff
@@ -0,0 +1,48 @@
...
+  $log->info("cmd: $cmd $args\n");
+  foreach my $key1 (keys %expectedResult) {
+    foreach my $key2 (keys %{$expectedResult{$key1} }) {
+      $log->info("Expected: $expectedResult{$key1}{$key2}\n");
+      $log->info("Cmd: $sizeYaml->{$key1}->{$key2}\n");
+      assertEq($expectedResult{$key1}{$key2}, $sizeYaml->{$key1}->{$key2});
+    }
+  }
+}
+1;
```
lorelei-sakai, line 48: Add a blank line before the final "1;", like our other perl files.

src/perl/vdotest/VDOTest/VDOCalculateSize.pm
```diff
@@ -0,0 +1,48 @@
...
+  my ($self) = assertNumArgs(1, @_);
+  my $device = $self->getDevice();
+  my $machine = $device->getMachine();
+  my $cmd = $self->findBinary("vdoCalculateSize");
+  my $args = "--slabBits 20 --physicalSize=300G " .
+             "--logicalSize=1T --indexMemorySize=2";
+  $machine->runSystemCmd("$cmd $args");
+  my $sizeYaml = YAML::Load($machine->getStdout());
+  $log->info("cmd: $cmd $args\n");
+  foreach my $key1 (keys %expectedResult) {
```
lorelei-sakai, line 40: This hash comparison is pretty generic. Don't we already have a utility method for doing this? (If we don't, it might be worth adding it to, say, Permabit::Utils rather than just inlining it here.)

src/perl/vdotest/VDOTest/VDOCalculateSize.pm
```diff
@@ -0,0 +1,48 @@
...
+  assertEq
+  assertNumArgs
+);
+use Permabit::Constants;
+use YAML;
+use base qw(VDOTest);
+
+my $log = Log::Log4perl->get_logger(__PACKAGE__);
+my %expectedResult;
+$expectedResult{'VDO'}{'Slab Size'}="1048576";
```
lorelei-sakai, line 22: I think the expected results would be clearer if they are spelled as a singe hash structure instead of several separate assignments like this. (Also, the expectations will vary depending on the parameters so this should be incorporated into the testcase itself.)

src/perl/vdotest/VDOTest/VDOCalculateSize.pm
```diff
@@ -0,0 +1,48 @@
+##
+# Basic vdoCalculateSize tests
+#
+# $Id$
```
lorelei-sakai, line 4: I know you just copied this, but the $id tag is for Perforce so we don't need to keep including them for new code.

src/packaging/src-dist/MANIFEST.yaml
```diff
@@ -84,6 +84,7 @@ tarballs:
             - slabSummaryReader.[hc]
             - userVDO.[hc]
             - vdoAudit.c->vdoaudit.c
+            - vdoCalculateSize.c
```
lorelei-sakai, line 87: This should be downcased like our other user utilities. Compare to vdoDebugMetadata, below.

src/c++/vdo/user/Makefile
```diff
@@ -26,12 +26,14 @@ USER_OBJS = blockMapUtils.o      \
             parseUtils.o         \
             slabSummaryReader.o  \
             userVDO.o            \
+            vdoCalculateSize.o   \
```
lorelei-sakai, line 29: I don't think you need this here. The USER_OBJS group is for object files that do not turn into executables by themselves.

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,358 @@
...
+#include "constants.h"
+#include "status-codes.h"
+#include "types.h"
+#include "vdoConfig.h"
+
+static const u64 KB = 1024;
+static const u64 MB = 1024 * KB;
+static const u64 GB = 1024 * MB;
+
+#define CHAPTER_PER_GIG             1024
```
lorelei-sakai, line 30: Do not hardcode these values. They all exist in VDO or in UDS so you should include the canonical values so they don't get out of sync.

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,358 @@
...
+ */
+
+#include <err.h>
+#include <getopt.h>
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+/* .h file in src/c++/vdo/user */
```
lorelei-sakai, line 16: You don't need hese comments. Also, the /user files generally come after the /base files. (Compare with other user utilities.)

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,358 @@
...
+static const u64 GB = 1024 * MB;
+
+#define CHAPTER_PER_GIG             1024
+#define DEFAULT_BLOCKMAP_CACHE_SIZE 32768
+#define DEFAULT_GEOMETRY_BLOCK      1
+#define DEFAULT_SLAB_SUMMARY_SIZE   64
+#define DEFAULT_SUPERBLOCK_SIZE     1
+
+enum {
+  MIN_SLAB_BITS        =  4,
```
lorelei-sakai, line 37: This value has changed. (But also, it will soon be defined in constants.h, too.)

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,358 @@
...
+
+/**********************************************************************/
+static void usage(void)
+{
+  fprintf(stderr,
+          "Usage: vdoCalculateSize [--blockMapCacheSize=size]\n"
+          "                        [--indexMemorySize=GB]\n"
+          "                        [--logicalSize=B]\n"
+          "                        [--physicalSize=B]\n"
+          "                        [--slabBits=GB]\n"
```
lorelei-sakai, line 219: Bruce's new formatting code switches to specifying slab size instead of slab bits, and we should do the same here.

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,358 @@
...
+      usage();
+      break;
+    }
+  }
+
+  if (optind < argc) {
+    usage();
+  }
+
+  u64 dedupWindowSize = getDedupWindowSize(indexMemorySize);
```
lorelei-sakai, line 329: I would find it slightly clearer if this calculation and printing part is in a differnt function than the option parsing part.

Also, i probably makes sens to create a C structure to sore all he numbers related to the configuration and simply use that structure instead of passing around individual numbers.

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,358 @@
...
+{
+  char *indexMemorySize = NULL;
+  int result = 0, sparseIndex = 0, threadCount = 0;
+  unsigned int slabBits = DEFAULT_SLAB_BITS;
+
+  u64 blockMapCacheSize = DEFAULT_BLOCKMAP_CACHE_SIZE;
+  u64 logicalSize = 0, physicalSize = 0;
+
+  enum { OPT_BMCS = 'A', OPT_IMS,
+         OPT_LS, OPT_PS, OPT_SB, OPT_SI, OPT_TC};
```
lorelei-sakai, line 268: I'm not sure what I think about this enumeration. Our other utilities just use letters here.

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,358 @@
...
+  while ((opt = getopt_long_only(argc, argv, "", options, NULL)) != -1) {
+    switch (opt) {
+    case OPT_BMCS:
+      result = parseSize(optarg, true, &blockMapCacheSize);
+      if (result != VDO_SUCCESS) {
+        usage();
+      }
+      break;
+    case OPT_IMS:
+      indexMemorySize = optarg;
```
lorelei-sakai, line 291: I think you'll need to do parsing on the memory size to reflect the fact that we can have small index sizes, too. (E.g. 0.25)

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,358 @@
...
+         DEFAULT_SLAB_SUMMARY_SIZE +
+         DEFAULT_VDO_RECOVERY_JOURNAL_SIZE;
+}
+
+/*
+  Each block map page contains 812 entries, each logical block is one entry.
+  First, calculate the total leaves pages by dividing 812. We then calculate
+  additional block map pages at the upper level.
+*/
+static u64 getBlockMapPages(u64 logicalBlocks)
```
lorelei-sakai, line 80: These block map-related functions might be better off in blockMapUtils.h, since they could be useful to other utilities, too.

lorelei-sakai (changes requested):
Largely organizational issues. Also several points where this utility should be more similar to our existing user tools.

The test should also be more thorough (more testcases).

And the commit message should describe more speciifcally what this PR does rather than just copying text from the ticket.

src/perl/vdotest/VDOTest/VDOCalculateSize.pm
C2Redhat: Done

src/perl/vdotest/VDOTest/VDOCalculateSize.pm
C2Redhat: Done

src/packaging/src-dist/MANIFEST.yaml
C2Redhat: Done

src/c++/vdo/user/Makefile
C2Redhat: Done

src/c++/vdo/user/vdoCalculateSize.c
C2Redhat: Done

src/c++/vdo/user/vdoCalculateSize.c
C2Redhat: Update to 13.  Will update later after https://github.com/dm-vdo/vdo-devel/pull/216 merged

src/c++/vdo/user/vdoCalculateSize.c
C2Redhat: We will update this after https://github.com/dm-vdo/vdo-devel/pull/216

rhawalsh:
When I try to run the currently posted version of the tool...

1. Using `--version` works.  Thank you for implementing it.
2. The tool prints help text when there are insufficient arguments provided (including no arguments at all)

I have bumped into a couple of issues, though.

1. The help text when providing insufficient arguments should just print the error and usage.
This is what I got
```
$ ./vdoCalculateSize                                  
vdoCalculateSize: logicalSize and physicalSize are required                                                                                                   
Usage: vdoCalculateSize [--blockMapCacheSize=size]                             
                        [--humanReadable]                                                                                                                     
                        [--indexMemorySize=GB]                                                                                                                
                        [--logicalSize=B]                                                                                                                                                                                                                                                                                    
                        [--physicalSize=B]                                                                                                                    
                        [--slabBits=GB]                                                                                                                                                                                                                                                                                      
                        [--sparseIndex]                                                                                                                       
                        [--version]                                                                                                                                                                                                                                                                                          
                                                                                                                                                              
DESCRIPTION                                                                                                                                                                                                                                                                                                                  
  Calculate VDO space and memory usage.                                        

OPTIONS:                                                                       
  --blockMapCacheSize=size     block map cache size                            

  --help                       display this help and exit                                                                                                     

  --humanReadable            print sizes in human readable format                                                                                             

  --indexMemorySize=<gigabytes>                                                
    Specify the amount of memory, in gigabytes, to devote to the                                                                                              
    index. Accepted options are .25, .5, .75, and all positive                                                                                                
    integers.                                                                  

  --logicalSize=GB             VDO logical size                                

  --physicalSize=GB            VDO physical size                               

  --slabBits                   slabBits                                        
    Set the free space allocator's slab size to 2^<bits> 4 KB blocks.                                                                                         
    <bits> must be a value between 4 and 23 (inclusive), corresponding                                                                                        
    to a slab size between 128 KB and 32 GB. The default value is 19                                                                                          
    which results in a slab size of 2 GB. This allocator manages the                                                                                          
    space VDO uses to store user data.                                         
    The maximum number of slabs in the system is 8192, so this value                                                                                          
    determines the maximum physical size of a VDO volume. One slab is                                                                                         
    the minimum amount by which a VDO volume can be grown. Smaller                                                                                            
    slabs also increase the potential for parallelism if the device                                                                                           
    has multiple physical threads. Therefore, this value should be set                                                                                        
    as small as possible, given the eventual maximal size of the                                                                                              
    volume.                                                                    

  --sparseIndex                Default false                                   

  --version                                                                    
                  output version information and exit                                                                                                         

```
My desired output:
```
$ ./vdoCalculateSize                                  
vdoCalculateSize: logicalSize and physicalSize are required                                                                                                   
Usage: vdoCalculateSize [--blockMapCacheSize=size]                             
                        [--humanReadable]                                                                                                                     
                        [--indexMemorySize=GB]                                                                                                                
                        [--logicalSize=B]                                                                                                                                                                                                                                                                                    
                        [--physicalSize=B]                                                                                                                    
                        [--slabBits=GB]                                                                                                                                                                                                                                                                                      
                        [--sparseIndex]                                                                                                                       
                        [--version]                                                                                                                                                                                                                                                                                          
```
3. It appears that indexMemorySize is also required, despite not being called out.
Example:
```
$ ./vdoCalculateSize --blockMapCacheSize=128 --logicalSize=1T --physicalSize=1G
vdoCalculateSize: parseInt failed: getDedupeWindowSize
```
When I add `--indexMemorySize=1`, it provides me with an error tied to physical size (which is correct):
```
$ ./vdoCalculateSize --blockMapCacheSize=128M --indexMemorySize=1 --logicalSize=1T --physicalSize=1G --humanReadable
vdoCalculateSize: Physical Size too small
```
4. The word "Journal" is still spelled incorrectly.  (s/Jouranl/Journal/)
```
$ ./vdoCalculateSize --blockMapCacheSize=128M --indexMemorySize=1 --logicalSize=1T --physicalSize=15G
Input Parameters:
  Physical Size: 16106127360
  Logical Size: 1099511627776
  Slab Bits: 19
  Sparse: 0
  Index Memory: 1
Storage Usage:
  Total BlockMap Pages: 1355763712
  UDS Index Size: 11193331712
  Slab Reference Count Usage: 1187840
  Slab Jouranl Usage: 1835008
Total In Memory Usage: 560952016896
  Block Map Cache: 549755813888
  Forest Memory Usage: 1683456
  Slab Reference Count Usage: 1187840
  UDS Index Size: 11193331712
Volume Characteristics in blocks:
  Blocksize: 4096
  Physical Blocks: 3932160
  Logical Blocks: 268435456
  Slab Size: 524288
  Slab Count: 2
  Index Memory: 1
  Sparse: 0
VDO Meta Data:
  Superblock Size: 4096
  Geometery Block Size: 4096
  VDO Blockmap Tree Root Count: 60
  Slab Summary Size: 262144
  Recovery Journal Size: 134217728
  UDS Index Size: 11193331712
  Total BlockMap Pages Usage: 1355763712
  Slab Reference Count Usage: 1187840
  Slab Jouranl Usage: 1835008
```
5. The usage and help wrapping the arguments in square brackets conveys to me that they are optional.  Some of them are, but clearly based on the errors, --indexMemorySize, --logicalSize, and --physicalSize are required.  Therefore, they should not be wrapped in square brackets.
Current output:
```
$ ./vdoCalculateSize --help
Usage: vdoCalculateSize [--blockMapCacheSize=size]
                        [--humanReadable]                                                                                                                                                                          
                        [--indexMemorySize=GB]
                        [--logicalSize=B]
                        [--physicalSize=B]
                        [--slabBits=GB]
                        [--sparseIndex]
                        [--version]
...
```
Preferred output:
```
$ ./vdoCalculateSize --help
Usage: vdoCalculateSize --indexMemorySize=GB
                        --logicalSize=B
                        --physicalSize=B
                        [--blockMapCacheSize=size]
                        [--humanReadable]
                        [--slabBits=GB]
                        [--sparseIndex]
                        [--version]
...
```
6. The units printed in the help message are not consistent for `--blockMapCacheSize` (uses "size" vs "B" or "GB").  Also, the usage output for `--logicalSize=B` and `--physicalSize=B` is not consistent with the expanded text below where it states `--logicalSize=GB` and `--physicalSize=GB`.
Current output:
```
$ ./vdoCalculateSize --help                                                                                                                                                               
Usage: vdoCalculateSize [--blockMapCacheSize=size]                                                                                                                                                                 
                        [--humanReadable]
                        [--indexMemorySize=GB]
                        [--logicalSize=B]
                        [--physicalSize=B]
                        [--slabBits=GB]
                        [--sparseIndex]
                        [--version]

DESCRIPTION
  Calculate VDO space and memory usage.

OPTIONS:
  --blockMapCacheSize=size     block map cache size

  --help                       display this help and exit

  --humanReadable            print sizes in human readable format

  --indexMemorySize=<gigabytes>
    Specify the amount of memory, in gigabytes, to devote to the
    index. Accepted options are .25, .5, .75, and all positive
    integers.

  --logicalSize=GB             VDO logical size

  --physicalSize=GB            VDO physical size

```
Preferred output:
```
$ ./vdoCalculateSize --help                                                                                                                                                               
Usage: vdoCalculateSize [--blockMapCacheSize=MB]
                        [--humanReadable]
                        [--indexMemorySize=GB]
                        [--logicalSize=GB]
                        [--physicalSize=GB]
                        [--slabBits=GB]
                        [--sparseIndex]
                        [--version]

DESCRIPTION
  Calculate VDO space and memory usage.

OPTIONS:
  --blockMapCacheSize=MB     block map cache size

  --help                       display this help and exit

  --humanReadable            print sizes in human readable format

  --indexMemorySize=<gigabytes>
    Specify the amount of memory, in gigabytes, to devote to the
    index. Accepted options are .25, .5, .75, and all positive
    integers.

  --logicalSize=GB             VDO logical size

  --physicalSize=GB            VDO physical size

```

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,514 @@
...
+    return 512;
+  } else if (strcmp(indexMemorySize, "0.75") == 0) {
+    return 768;
+  } else {
+    int number;
+    int result = parseInt(indexMemorySize, &number);
+    if (result != VDO_SUCCESS) {
+      errx(result, "parseInt failed: getDedupeWindowSize");
+    }
+    return number * CHAPTER_PER_GIG;
```
lorelei-sakai, line 176: It is true that the deduplication window for a dense index is about 1024 times as large as the amount of memory the index uses. However it is misleading to use CHAPTER_PER_GIG because the deduplication window is a function of how many entries the index can store and is not directly related to the chapter count in any way.
(More precisely, it is 4KB * the number of entries the index can store.)

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,514 @@
...
+         uds_string_error(result, errorBuffer, sizeof(errorBuffer)));
+  }
+  return indexBlocks;
+}
+
+/*
+  For deduplication window calculation, please refer to geometry.c in
+  utils/uds/geometry.c for detail on how deduplication window is calculated.
+ */
+static u64 getDedupeWindowSize(char *indexMemorySize)
```
lorelei-sakai, line 161: This function also calculated the wrong value for a sparse index; a sparse index has a dedupe window 10x as large as a dense index with the same size. This function either needs to take the sparse flag as a parameter (and use it), or be rewritten in term of the number of index entries instead. (The latter suggestion would require calculating the number of index entries somewhere, so I don't know which of these two approaches is simpler.)

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,514 @@
...
+  result = computeIndexBlocks(&indexConfig, &indexBlocks);
+  if (result != VDO_SUCCESS) {
+    errx(result, "computeIndexBlocks failed: %s",
+         uds_string_error(result, errorBuffer, sizeof(errorBuffer)));
+  }
+  return indexBlocks;
+}
+
+/*
+  For deduplication window calculation, please refer to geometry.c in
```
lorelei-sakai, line 158: This comment says many things twice, could you clean it up?

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,514 @@
...
+  vdoInfo->totalReferenceCount =
+    vdo_get_saved_reference_count_size(vdoInfo->userDataBlocks);
+}
+
+static void calculateBlockMapMetaInfo(struct vdoInfo *vdoInfo)
+{
+  vdoInfo->totalBlockMapLeaves =
+    vdo_compute_block_map_page_count(vdoInfo->logicalBlocks);
+  vdoInfo->totalBlockMapPages = getBlockMapPages(vdoInfo->logicalBlocks);
+  vdoInfo->totalForestMemoryUsage = vdoInfo->totalBlockMapPages -
```
lorelei-sakai, line 202: This line wrapping is awkward, could you split it after the '=' instead? hat way the X - Y expression would could appear on a single line.

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,514 @@
...
+                                    vdoInfo->totalBlockMapLeaves;
+  vdoInfo->totalUsableSpace = (vdoInfo->userDataBlocks -
+			       vdoInfo->totalBlockMapPages -
```
lorelei-sakai, line 205: The indentation here is very weird, I don't know what's going on. Could you fix it?

src/c++/vdo/user/man/vdocalculatesize.8
```diff
@@ -0,0 +1,54 @@
...
+.B \-\-blockMapCacheSize=\fIsize\fP
+block map cache size
+.TP
+.B \-\-logicalSize=\fIsize\fP
+VDO logical size
+.TP
+.B \-\-physicalSize=\fIsize\fP
+VDO physical size
+.TP
+.B \-\-slab\-bits=\fIbits\fP
```
lorelei-sakai, line 25: Note also that the format of this does not match the other arguments, or the code in the actual utility; the option should be camel-cased: --slabBits (or --slabSize).

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,514 @@
...
+  Recovery Journal 32K
+  UDS index - Created during VDO format.
+
+  The rest of the data are divided into slabs. Within each slab, it contains
+  user data and three types of metadata, block map page, reference count
+  and slab journal. These meta data usages depend on the size of the VDO and
+  the size of each slab.
+
+  Slab size is configurable during VDO format. Default slab size is 2G.
+  Slab size can be adjusted using slab bit, default is 19. That is 2 to
```
lorelei-sakai, line 98: The explanation here of why 19 slab bits (== 0.5M) should produce a 2G slab size is a bit confusing. This would be simplified by using slab size as the the argument, becuase this explanation wouldn't be needed at all.

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,514 @@
...
+  /* VDO block info */
+  u64 physicalBlocks;
+  u64 logicalBlocks;
+  u64 userDataBlocks;
+  int totalSystemBlock;
+
+  /* Slab info */
+  int slabBlockCount;
+  u64 slabCount;
+  u64 totalSlabJouranl;
```
lorelei-sakai, line 67: "Journal" is misspelled here, so it will need to fixed where this parameter is used. Also "Journal" is misspelled in a couple of string constants so it's probably worth doing a find-and-replace. (All misspellings I saw were the same, "Jouranl.")

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,514 @@
...
+  Each block map page contains 812 entries, each logical block is one entry.
+  First, calculate the total leaves pages by dividing 812. We then calculate
+  additional block map pages at the upper level.
+*/
+static u64 getBlockMapPages(u64 logicalBlocks)
+{
+  u64 totalBlockMapLeaves = vdo_compute_block_map_page_count(logicalBlocks);
+  u64 parentBlockUsage = totalBlockMapLeaves;
+  int additionalPage = 0;
+  for(int i=0; i < VDO_BLOCK_MAP_TREE_HEIGHT - 1; i++) {
```
lorelei-sakai, line 115: Another "spacs around operators" issue, around '=' here.

src/c++/vdo/user/vdoCalculateSize.c
lorelei-sakai: Given that Bruce is working on a branch now, let's not gate this on the formatting code. Also see discussion elsewhere on whether we should use --slab-size or --slab-bits.

src/perl/vdotest/VDOTest/VDOCalculateSize.pm
```diff
@@ -0,0 +1,731 @@
...
+      "physicalSize"      => "300G",
+      "slabBits"          => "20",
+      );
+  my %expectedResult = (
+       'VDO' => { 'Slab Size'=> "1048576" },
+       'VDO meta data' => {
+          'UDS Index Size' => "10674MB",
+          'Total BlockMap Pages' => "1292MB"},
+       'VDO Usage' => {
+         'Usable Size'  => "288GB",
```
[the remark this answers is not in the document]
C2Redhat: I was lining up with the line below, but I will remove them

src/perl/vdotest/VDOTest/VDOCalculateSize.pm
```diff
@@ -0,0 +1,731 @@
...
+	 'Block Map Cache'  => "128MB",
+	 'Forest Memory Usage' => "1MB",
+         'Slab Reference Count Usage' => "73MB",
+       },
+      );
+  my $cmdArgs = "";
+  foreach my $arg (keys %vdoCalculateSizeArgs) {
+    $cmdArgs .= "--" . "$arg" . " $vdoCalculateSizeArgs{$arg} ";
+  }
+  $log->info("CMD ARGS $cmdArgs\n");
```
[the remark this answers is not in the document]
C2Redhat: I think it is good to see all the parameter, but it is easy to removed.

src/c++/vdo/user/vdoCalculateSize.c
C2Redhat: OK.  I will add in slab-size as argument

src/c++/vdo/user/printUtils.c
```diff
@@ -0,0 +1,45 @@
...
+#include "printUtils.h"
+
+static const u64 KB = 1024;
+static const u64 MB = 1024 * KB;
+static const u64 GB = 1024 * MB;
+static const u64 TB = 1024 * GB;
+static const u64 PB = 1024 * TB;
+
+#define PRINTSIZE 20
+static char printString[PRINTSIZE];
```
lorelei-sakai, line 16: Please don't use a static string here; please pass in a buffer or a piece of memory to write to

This does not scale well and it could lead to strange bugs if more than one process tries to use the print functions at the same time. (For a utility like this, that's unlikely, but it's not a good design.)

src/c++/vdo/user/printUtils.c
```diff
@@ -0,0 +1,45 @@
...
+  } else {
+    sprintf(printString, "%ld", size);
+  }
+  return printString;
+}
+
+void setReadablePrintString(u64 size)
+{
+  long double readableSize = 0;
+  if ((readableSize = (long double)size / PB) > 1) {
```
lorelei-sakai, line 32: Putting an assignment in the if() clause like this is usually confusing. (In flag, compilers often produce a warning for this because it can easily be a typo.)
It woul be clearer to spell these check something like this:
if (size >PB) {
  sprintf(printString, "%.2LfP", (long double) size / PB);

src/c++/vdo/user/printUtils.c
```diff
@@ -0,0 +1,45 @@
...
+    sprintf(printString, "%.2LfP", readableSize);
+  } else if ((readableSize = (long double)size / TB) > 1) {
+    sprintf(printString, "%.2LfT", (long double)readableSize);
```
lorelei-sakai, line 35: I'm also a bit skeptical of casting the size to a long double for formatting; is that really necessary? You certainly don't need to cast the readableSize value, since that's already a long double. (But if you take my suggestion above then that local will go away anyway.)

src/c++/vdo/user/vdoCalculateSize.c
lorelei-sakai: I guess we decided these values are reasonable.
Except for one: CHAPTER_PER_GIG is no longer used.

lorelei-sakai (changes requested):
Nwe comments. Mostly formtating issues, plus a few concerns about memory string allocation and usage.

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,596 @@
...
+{
+  u64 slabBlock = vdoInfo->slabSize / VDO_BLOCK_SIZE;
+
+  if (slabBlock == 0 || (slabBlock & (slabBlock - 1)) != 0) {
+    warnx("Slabsize is not power of 2");
+    return(VDO_OUT_OF_RANGE);
+  }
+  int slabBits = 0;
+  while (slabBlock > 1) {
+    slabBlock >>= 1; // Right-shift n by 1 bit
```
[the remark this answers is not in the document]
C2Redhat: Removed

src/c++/vdo/user/man/vdocalculatesize.8
```diff
@@ -0,0 +1,54 @@
...
+.SH DESCRIPTION
+.B vdocalculatesize
+Calculate VDO space and memory usage
+.PP
+.SH OPTIONS
+.TP
+.B \-\-help
+Print help message and exit.
+.TP
+.B \-\-blockMapCacheSize=\fIsize\fP
```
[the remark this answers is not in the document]
C2Redhat: added

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,596 @@
...
+    int result = parseInt(indexMemorySize, &number);
+    if (result != VDO_SUCCESS) {
+      errx(EXIT_FAILURE, "parseInt failed: getDedupeWindowSize");
+    }
+    /* deduplication window size for dense index is about 1024 times
+       size of index memory */
+    dedupeWindowSize = number * 1024;
+  }
+  if (vdoInfo->sparseIndex) {
+    /* Dedupwindow with sparse index is 10 times larger than dense index */
```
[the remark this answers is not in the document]
C2Redhat: fixed

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,596 @@
...
+{
+  vdoInfo->dedupeWindowSize = getDedupeWindowSize(vdoInfo);
+  vdoInfo->udsIndexSize = getUDSIndexSize(vdoInfo->indexMemorySize,
+					  vdoInfo->sparseIndex);
+  calculateVDOBlockInfo(vdoInfo);
+  calculateSlabInfo(vdoInfo);
+  calculateBlockMapMetaInfo(vdoInfo);
+}
+
+static u64 minimunVDOSize(struct vdoInfo *vdoInfo)
```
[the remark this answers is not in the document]
C2Redhat: fixed

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,596 @@
...
+{
+  char *size = malloc(sizeof(optarg)+1);
+  if (isdigit(optarg[strlen(optarg)-1]))
+    sprintf(size,"%sG", optarg);
+  else
+    sprintf(size,"%s", optarg);
+  return size;
+}
+
+static u64 convertSlabSizeToBit(struct vdoInfo *vdoInfo)
```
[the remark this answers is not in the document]
C2Redhat: fixed

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,596 @@
...
+          "\n"
+          "  --version\n                  output version information and exit\n"
+          "\n");
+  }
+  exit(1);
+}
+
+static char *fixSizeArg(char *optarg)
+{
+  char *size = malloc(sizeof(optarg)+1);
```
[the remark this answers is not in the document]
C2Redhat: parseSize if being used by other functions, if I change parseSize, we will need to change it everywhere.
Change to MB I think is not user friendly since when we will have a hard drive that is in MB.
fixSizeArg is only use within this program and will not be use in any place else.  It is not designed to 
be use by other called.

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,596 @@
...
+  printVDOStorageUsage(vdoInfo);
+  printVDOMemoryUsage(vdoInfo);
+  printVDOVolumeCharacteristics(vdoInfo);
+  printVDOMetaData(vdoInfo);
+}
+
+static int checkSlabSize(u64 slabSize)
+{
+  u64 slabBlock = slabSize / VDO_BLOCK_SIZE;
+  if(slabBlock < MIN_VDO_SLAB_BLOCKS) {
```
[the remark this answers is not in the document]
C2Redhat: fixed

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,596 @@
...
+    warnx("slab-block too small, minumum size %s",
+	  getSizeString(MIN_VDO_SLAB_BLOCKS * VDO_BLOCK_SIZE, 1));
```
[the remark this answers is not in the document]
C2Redhat: fixed

src/c++/vdo/user/vdoCalculateSize.c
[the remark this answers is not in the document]
C2Redhat: After doing some thinking, I have changed the the default size to MB and remove the fixSizeArg to simplified the program

src/c++/vdo/user/vdoCalculateSize.c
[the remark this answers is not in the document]
C2Redhat: I have changed that.

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,611 @@
...
+  A VDO volume contains three types of data, system metadata, user metadata
+  and user data. All VDO data is stored in a 4K block.
+
+  System metadata are constant regardless the size of VDO volume.  They are:
+  Super Block 1
+  Geometry Block 1
+  Block map root 60
+  Slab Summary 64
+  Recovery Journal 32K
+  UDS index - Created during VDO format.
```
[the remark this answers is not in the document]
C2Redhat: I am not sure I understand.  At this point. I think we can do it in 2.0 if there are any structure change.

lorelei-sakai (changes requested):
Generally an improvement from the last version. Some questions about the way slab-bits / slab-size is handled, plus some left-over comments from previous rounds.

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,597 @@
...
+    dedupeWindowSize = 768;
+  } else {
+    int number;
+    int result = parseInt(indexMemorySize, &number);
+    if (result != VDO_SUCCESS) {
+      errx(EXIT_FAILURE, "parseInt failed: getDedupeWindowSize");
+    }
+    /* deduplication window size for dense index is about 1024 times
+       size of index memory */
+    dedupeWindowSize = (u64)number * (1024 * 1024 * 1024) * 1024;
```
lorelei-sakai, line 182: I do not understand this change. It appeasr that you are expressing the dedupe window  in terms of bytes (e.g. an dense index size of "1" has a dedupe window of 1T == 10^40 bytes). If that is your intent, you should change the calculation the smaller indexes. A dense index of size "0.25" has a dedupe window of 256G == 256 * 1024 * 1024 * 1024 bytes.

Right now the calculation is inconsistent and I'm not sure which way you meant to be correct.

src/perl/vdotest/VDOTest/VDOCalculateSize.pm
```diff
@@ -0,0 +1,185 @@
...
+  my $device  = $self->getDevice();
+  $runProperties{machine} = $device->getMachine();
+  $runProperties{binary} => $self->findBinary("vdoCalculateSize"),
+}
+
+sub testBasicExecution {
+  my ($self) = assertNumArgs(1, @_);
+  my $machine = $runProperties{machine};
+  my $args = {
+       binary => $runProperties{binary},
```
lorelei-sakai, line 94: Passing these two properties in a hash seems like overkill. The machine is readily accessible from the test directly. The binary could probably be incorporated in the Command String class directly? I'd have to check against the way we do this elsewhere. But at a minimum you can probably resolve it directly in executeTest() where you want to use it.

src/perl/vdotest/VDOTest/VDOCalculateSize.pm
```diff
@@ -0,0 +1,185 @@
...
+}
+
+sub test2TLogicalSize {
+  my ($self) = assertNumArgs(1, @_);
+  my $machine = $runProperties{machine};
+  my $args = {
+    "binary" => $runProperties{binary},
+    "logical-size" => "2T",
+  };
+  my $myExpectedResult = {
```
lorelei-sakai, line 107: I think I see what you're going for, but I think you need to have a way to combine the generic expectations with the ones that change specifically for this scenario. Right now you're only checking the output values you've explicitly listed in your test hash, and not all the rest of them like you were doing before.

You can create a full expectations hash by using the test-local $myExpectedResult to override generic values in %expectedResult, I'd probably do it that way.

On a side note, see if you can come up with better names for these hashes. They are very similar, and if the name are also similar it can make it confusing to read. But naming is tricky, we can discuss that in person some time if you want.

src/perl/Permabit/CommandString/VDOCalculateSize.pm
```diff
@@ -0,0 +1,84 @@
...
+     "blockmap-cache-size" => "32K",
+     "name"                => "vdoCalculateSize",
+     "index-memory-size"   => "1",
+     "logical-size"        => "1T",
+     "physical-size"       => "300G",
+     "slab-bits"           => "20",
+     "sparse-index"        => undef,
+    );
+
+our %EXPECTED_RESULT
```
lorelei-sakai, line 34: The CommandString class doesn't use the %EXPECTED_RESULT hash, I think you can just remove it.

lorelei-sakai (changes requested):
Just commenting on differences from the last draft, the new test structure is genreally very nice. I did not think if making a CommandString class but it is a good abstraction here. I think it can be simplified a little in a few place, and the expectation hash does not seem to be checking enough properties, but generally very good.

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,590 @@
...
+#include "constants.h"
+#include "status-codes.h"
+#include "types.h"
+#include "vdoConfig.h"
+
+#include "blockMapUtils.h"
+#include "parseUtils.h"
+#include "printUtils.h"
+
+#define CHAPTER_PER_GIG             1024
```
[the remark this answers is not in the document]
C2Redhat: Done

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,590 @@
...
+
+/* Should change MIN_SLAB_BITS */
+enum {
+  MIN_SLAB_BITS        = 13,
+  DEFAULT_SLAB_BITS    = 19,
+  MAX_SLAB_BITS        = 23,
+};
+
+u64 MIN_VDO_SLAB_SIZE = (1 << 4) * VDO_BLOCK_SIZE;
+u64 MAX_VDO_SLAB_SIZE = (u64)(1 << 23) * VDO_BLOCK_SIZE;
```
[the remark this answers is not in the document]
C2Redhat: 1U << 23 actually doesn't work, the result is 0.

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,590 @@
...
+#define DEFAULT_SUPERBLOCK          1
+
+/* Should change MIN_SLAB_BITS */
+enum {
+  MIN_SLAB_BITS        = 13,
+  DEFAULT_SLAB_BITS    = 19,
+  MAX_SLAB_BITS        = 23,
+};
+
+u64 MIN_VDO_SLAB_SIZE = (1 << 4) * VDO_BLOCK_SIZE;
```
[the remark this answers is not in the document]
C2Redhat: Fix

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,590 @@
...
+u64 MAX_VDO_SLAB_SIZE = (u64)(1 << 23) * VDO_BLOCK_SIZE;
+
+#define FIXED_METADATA_BLOCKS DEFAULT_SUPERBLOCK + \
+                              DEFAULT_GEOMETRY_BLOCK + \
+                              DEFAULT_VDO_BLOCK_MAP_TREE_ROOT_COUNT + \
+                              VDO_SLAB_SUMMARY_BLOCKS + \
+                              DEFAULT_VDO_RECOVERY_JOURNAL_SIZE
+
+struct vdoInfo{
```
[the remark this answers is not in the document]
C2Redhat: Fix

src/c++/vdo/user/printUtils.c
```diff
@@ -0,0 +1,40 @@
...
+static const u64 TB = 1024 * GB;
+static const u64 PB = 1024 * TB;
+
+void getSizeString(u64 size, int humanReadable, char *printString)
+{
+  memset(printString, 0, PRINTSTRINGSIZE);
+  if (humanReadable) {
+    setReadablePrintString(size, printString);
+  } else {
+    sprintf(printString, "%ld", size);
```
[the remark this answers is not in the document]
C2Redhat: I disagree with this.  Since we know know the max size of the string, there is no reason to complicate the code since the caller could have say to want to print 1000000 but put in size string to 2.

src/perl/vdotest/VDOTest/VDOCalculateSize.pm
C2Redhat: This is updated

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,596 @@
...
+          "                        [--slab-size=MB]\n"
+          "                        [--sparse-index]\n"
+          "                        [--version]\n\n");
+  if (printDetail) {
+    fprintf(stderr,
+          "DESCRIPTION\n"
+          "  Calculate VDO space and memory usage.\n"
+          "\n"
+          "\n"
+          "  --block-map-cache-size=B       Block map cache size\n"
```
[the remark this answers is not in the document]
lorelei-sakai: This line should get updated to match whatever change you make above.

src/c++/vdo/user/vdoCalculateSize.c
```diff
@@ -0,0 +1,596 @@
...
+  }
+  exit(1);
+}
+
+static int convertSlabSizeToBit(struct vdoInfo *vdoInfo)
+{
+  u64 slabBlock = vdoInfo->slabSize / VDO_BLOCK_SIZE;
+
+  if (slabBlock == 0 || (slabBlock & (slabBlock - 1)) != 0) {
+    warnx("Slabsize is not power of 2");
```
lorelei-sakai: "Slab size" should be two words. Also be sure to update the bad parameter test that looks for this message.

lorelei-sakai (changes requested):
Several things addressed, thanks. The test looks to be in good shape.
The block-map-cache-size issue appears to just be documentation, but we do need to clear that up.
For follow-up issues, I think we should actually make a couple of issues, so we can separate the tasks by which are important and which are just nice-to-have, or trivial. That should also mean the individual tasks are small enough that they can be handled within a sprint instead of trying to do everything in one go. We can discuss tomorrow.

src/c++/vdo/user/man/vdocalculatesize.8
```diff
@@ -0,0 +1,58 @@
...
+Specify the amount of memory, in gigabytes, to devote to the
+index. Accepted options are .25, .5, .75, and all positive
+integers.
+.TP
+.B \-\-sparse-index
+Specify whether or not to use a sparse index.
+.PP
+.SH EXAMPLE
+.nf
+vdoCalculateSize --physical-size 600G --slab-bits 22 --logical-size 2T --index-memory-size 1 --blockmap-cache-size 8M
```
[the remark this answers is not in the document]
C2Redhat: Fix

lorelei-sakai (changes requested):
Missed some of the block map documentation updates, otherwise looks good.

lorelei-sakai (approved):
A long road, but I agree that's everything.

Looks good to me, yay.

PR #242, opened 2025-02-03
ci: allow org members to approve outside submitter PRs

bjohnsto:
Change the approver ci for outside submissions to be anyone in the entire dm-vdo org, instead of the current czar. Switch to using the VDODEVEL secret since that should have the permissions we need to get everyone in dm-vdo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,8 +75,8 @@ jobs:
         uses: trstringer/manual-approval@v1
         timeout-minutes: 1440
         with:
-          secret: ${{ secrets.GITHUB_TOKEN }}
-          approvers: ${{ join(github.event.pull_request.requested_reviewers.*.login, ', ') }}
+          secret: ${{ secrets.VDODEVEL }}
+          approvers: dm-vdo
           minimum-approvals: 1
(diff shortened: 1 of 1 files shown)


raeburn (approved):

.github/workflows/ci.yml
```diff
@@ -75,8 +75,8 @@ jobs:
         uses: trstringer/manual-approval@v1
         timeout-minutes: 1440
         with:
-          secret: ${{ secrets.GITHUB_TOKEN }}
-          approvers: ${{ join(github.event.pull_request.requested_reviewers.*.login, ', ') }}
+          secret: ${{ secrets.VDODEVEL }}
```
raeburn, line 78: Was the secret supposed to change too? If so it should be mentioned in the description.

raeburn (changes requested):

.github/workflows/ci.yml
bjohnsto: I'm testing to see if it works. Based on the doc, I think we need a secret that has the right permissions on the org. But I need to actually test it first

bjohnsto: updated

raeburn (approved):

PR #244, opened 2025-02-03
dm vdo indexer: reorder uds_request to reduce padding

raeburn:
Saves 48 kB (about 12%) per struct hash_zone.

### DIFF
--- a/src/c++/uds/src/uds/index-session.c
+++ b/src/c++/uds/src/uds/index-session.c
@@ -100,7 +100,6 @@ static int get_index_session(struct uds_index_session *index_session)
 
 int uds_launch_request(struct uds_request *request)
 {
-	size_t internal_size;
 	int result;
 
 	if (request->callback == NULL) {
@@ -121,10 +120,7 @@ int uds_launch_request(struct uds_request *request)
 	}
 
 	/* Reset all internal fields before processing. */
-	internal_size =
-		sizeof(struct uds_request) - offsetof(struct uds_request, zone_number);
-	// FIXME should be using struct_group for this instead
-	memset((char *) request + sizeof(*request) - internal_size, 0, internal_size);
+	memset(&request->internal, 0, sizeof(request->internal));
 
 	result = get_index_session(request->session);
 	if (result != UDS_SUCCESS)
--- a/src/c++/uds/src/uds/indexer.h
+++ b/src/c++/uds/src/uds/indexer.h
@@ -8,6 +8,7 @@
 
 #include <linux/mutex.h>
 #include <linux/sched.h>
+#include <linux/stddef.h>
 #include <linux/types.h>
 #ifdef __KERNEL__
 #include <linux/wait.h>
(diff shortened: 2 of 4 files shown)


src/c++/uds/src/uds/index-session.c
```diff
@@ -122,7 +122,7 @@ int uds_launch_request(struct uds_request *request)
 
 	/* Reset all internal fields before processing. */
 	internal_size =
-		sizeof(struct uds_request) - offsetof(struct uds_request, zone_number);
+		sizeof(struct uds_request) - offsetof(struct uds_request, unbatched);
 	// FIXME should be using struct_group for this instead
```
lorelei-sakai, line 126: Can we address this FIXME in some fashion? Either add a struct group to the uds_request structure (ideally), or say we've thought about it, but we've decided not to do it for {reasons}.

I really don't like the kludge of having to name a specific field in the structure, resurfaced here because you had to change it.

src/c++/uds/src/uds/indexer.h
```diff
@@ -275,9 +275,15 @@ struct uds_request {
 
 	/*
 	 * The remaining fields are used internally and should not be altered by clients. The index
-	 * relies on zone_number being the first field in this section.
+	 * relies on unbatched being the first field in this section.
 	 */
```
lorelei-sakai, line 279: I'm not sure if putting "unbatched" first is better or if putting "requeued" first is better. They're both kind of ugly, rarely used flags that I don't really want to draw attention to.

Requeued is probably better; I think requeued is used rarely and only for the sparse index, but "unbatched" is only ever set to true in unit tests.

lorelei-sakai (changes requested):
The basic idea seems sounds. I would like a bit more specificity in the commit message about how many bytes we save in the uds_request structure. Calling out he larger saving is also good (highlighting that we allocate thousands of these structures) but I would stick with absolute numbers and drop the percent measurement.

src/c++/uds/src/uds/index-session.c
raeburn: using struct_group now

src/c++/uds/src/uds/indexer.h
raeburn: struct_group gets rid of this

raeburn:
New version pushed makes uds_zone_message packed and further shuffles things to get another 8 bytes out, as discussed offline. Also rebased to current vdo-devel.

lorelei-sakai (changes requested):
Code changes all look good. The commit message on the second (upstream) commit needs revising, though. The last line looks like notes that weren't supposed to be there? And please sign it, too. (Also, x86-64 with a hyphen? Isn't it x86_64?)

raeburn:
Whoops, careless commit message editing while rebasing. Fixed the extra line, and switched the architecture to use an underscore.

lorelei-sakai (approved):
Commit message is better, but it still isn't signed-off.  I can add that when pushing it upstream but do try to remember going forward. ("git commit -s" will do it automatically.)

PR #245, opened 2025-02-03
Update ci.yml to use usernames directly.

bjohnsto:
Replace dm-vdo with specific usernames to get this action to work


lorelei-sakai (approved):
i'm not super happy about encoding the membership list in the action, but if there isn't another way...

PR #247, opened 2025-02-04
Remove a couple of remaining references to rings.

sweettea:
Rings were a list-like abstraction replaced by lists, but there are still a couple of confusing references to them. Fix those.

### DIFF
--- a/src/c++/vdo/base/block-map.c
+++ b/src/c++/vdo/base/block-map.c
@@ -451,7 +451,7 @@ static struct page_info * __must_check find_page(struct vdo_page_cache *cache,
  * select_lru_page() - Determine which page is least recently used.
  *
  * Picks the least recently used from among the non-busy entries at the front of each of the lru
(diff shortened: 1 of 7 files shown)


lorelei-sakai (changes requested):
It looks like there's one reference to "ring" in priority-table.c that you missed. Could you add that one, too?

Also, I would not be averse to squashing these two commit together, especially since they both touch dedupe.c.  Keeping them separate is fine though.

Otherwise, looks pretty good. I'll appreove and merge once you makes the changes.

sweettea:
Sorry I missed this comment initially -- added priority-table.c edit, squashed commits.

lorelei-sakai (approved):
Change generally look good.

In the commit message, the title should not have a period. And I'd replace "some" with "all" to match what you say later on.

Neither of these are blockers; I can adjust the message when sending it upstream. However I am going to wait to merge until the CI passes again. (IT's infrastructure issues, nothing to do with this change.)

PR #250, opened 2025-02-04
[VDO-5886] dm vdo user: increase minimum slab size for users

lorelei-sakai:
This is a merge of vdo-devel/240 to the 8.3 branch for RHEL 10 (post-GA).

Prevent users from trying to use small slab sizes which were only intended for unit test scenarios. Also document the updated limits.


bjohnsto (approved):
Based on our discussions, this looks right to me

PR #258, opened 2025-02-14
[VDO-5889] Fix device name used in changeVDOSetting

bjohnsto:
The function changeVDOSetting in LVMVDO.pm was using the wrong name for the pool. It should have been using getLVMVDOName.

I've also updated a couple comments that were incorrect.


lorelei-sakai (approved):
In general, please doesn't refer to Jira tickets in commit messages. (That information should be in the PR description instead.)

For an internal test change, it doesn't really matter, though.

PR #261, opened 2025-02-18
deprecate compression message

sweettea:
As discussed, this change emits a warning when people use the compression message.

Happy to send this upstream myself if preferred once this is acceptable.

### DIFF
--- a/src/c++/vdo/base/dm-vdo-target.c
+++ b/src/c++/vdo/base/dm-vdo-target.c
@@ -1177,6 +1177,9 @@ static int __must_check process_vdo_message_locked(struct vdo *vdo, unsigned int
 						   char **argv)
 {
 	if ((argc == 2) && (strcasecmp(argv[0], "compression") == 0)) {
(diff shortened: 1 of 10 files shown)


lorelei-sakai (commented):
Deprecation is mostly a question of how and when we communicate it. We'll try to figure out what we want to do for this next week once the whole team is around.

The message isn't really hurting anything, but if we do deprecate it, it might make sense to deprecate all the run-time messages together.

sweettea:
I realized a missing part of this was updating unit tests to not (essentially) use the message. So +1 change.

lorelei-sakai:
The unit test changes aren't strictly necessary, but I agree that if we deprecate the message, the unit tests should switch to using the new preferred code pathway. So that commit is fine.

lorelei-sakai:
I shouldn't have let this issue linger for so long.

There's really no reason to deprecate the online compression enabling and disabling feature, and it's independent of the work in PR #241 so I'm closing this one.

PR #264, opened 2025-02-19
[VDO-5888] dm vdo test: update expected instance number for test changes

raeburn:
Commit b3872222 changes the VDO device creation sequence such that this test only starts the device once, as opposed to the start-stop-start sequence that used to happen. This changes the instance number assigned to the device, which most tests don’t care about but this one checks.


lorelei-sakai (approved):
Looks good.

PR #265, opened 2025-02-22
WIP: add compression level

sweettea:
# Add configurable LZ4 compression levels to VDO devices

This PR introduces a new `compressType` parameter that allows users to configure the LZ4 compression level used by VDO devices. This change adds the interface and parameter parsing, but does not yet implement the actual compression level configuration, hence WIP.

## Interface Changes

The PR adds a new optional `compressType` parameter to the device configuration. Values can be:

- `none` - Disable compression (default)
- `lz4` - Enable LZ4 compression with default acceleration factor
- `lz4:<level>` - Enable LZ4 compression with specific acceleration factor

Example usage:
```
# Default LZ4 compression
dmsetup create vdo0 --table "0 1048576 vdo /dev/sdb 4096 ... compressType=lz4"

# LZ4 with acceleration factor of 4
dmsetup create vdo0 --table "0 1048576 vdo /dev/sdb 4096 ... compressType=lz4:4"
```

The existing `compression=on/off` parameter remains supported but is now considered legacy in favor of the more flexible `compressType`.

## Questions for Review

Is this the right interface? Do the versions get bumped correctly? Does the doc capture the essential points?

### DIFF
--- a/doc/vdo.rst
+++ b/doc/vdo.rst
@@ -168,7 +168,16 @@ Miscellaneous parameters:
 
 	compression:
 		Whether compression is enabled. The default is 'off'; the
-		acceptable values are 'on' and 'off'.
+		acceptable values are 'on' and 'off'. It is preferred to
+                use the compressType parameter.
+
+        compressType:
+                The compression type, if any, to use. Acceptable values are
+                'none', the default; 'lz4', which uses the same behavior
+                as the 'compression=on' parameter; and 'lz4:<level>' for
+                integer levels. If a level is specified, this is used as the
+                acceleration factor for lz4. See lz4 docs for the specific
+                meanings of various acceleration factors. 
 
 Device modification
 -------------------
--- a/src/c++/vdo/base/dm-vdo-target.c
+++ b/src/c++/vdo/base/dm-vdo-target.c
@@ -9,6 +9,7 @@
 #include <linux/delay.h>
 #include <linux/device-mapper.h>
 #include <linux/err.h>
+#include <linux/lz4.h>
 #include <linux/module.h>
 #include <linux/mutex.h>
 #include <linux/spinlock.h>
@@ -649,6 +650,34 @@ static int process_one_key_value_pair(const char *key, unsigned int value,
 }
 
 /**
+ * parse_compression_level() - Process the compression level, if any.
+ * @value[in]: The string (possibly only a null byte) recording the level
+ * @level_ptr[out]: The pointer to return the level into, if one is provided.
+ *
+ * Does not log the error reason.
+ *
+ * Return: 0 or -EINVAL.
+ */
+static int parse_compression_level(const char *value, int *level_ptr)
+{
+	int ret;
+	int level;
+
+	if (value[0] == '\0')
+		return 0;
+
+	if (value[0] != ':')
+		return -EINVAL;
+
+	ret = kstrtoint(&value[1], 10, &level);
+	if (ret)
+		return -EINVAL;
+
+	*level_ptr = level;
+	return 0;
+}
+
+/**
  * parse_compression() - Process the compression type parameter.
  * @value: The parameter specifying the type
  * @config: The configuration data structure to update.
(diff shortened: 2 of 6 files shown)


src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -826,7 +834,7 @@ static int parse_device_config(int argc, char **argv, struct dm_target *ti,
 	};
 	config->max_discard_blocks = 1;
 	config->deduplication = true;
-	config->compression = false;
+	config->compression = VDO_NO_COMPRESSION;
```
lorelei-sakai, line 905: I think it's cleaner to keep the internal representation as an on/off boolean, and a separate parameter for the algorithm type.

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -648,6 +649,34 @@ static int process_one_key_value_pair(const char *key, unsigned int value,
...
+	int ret;
+	int level;
+
+	if (value[0] == '\0')
+		return 0;
+
+	if (value[0] != ':')
+		return -EINVAL;
+
+	ret = kstrtoint(&value[1], 10, &level);
```
lorelei-sakai, line 672: lz4 acceleration factors seem to only be well defined for positive integers, so doing a signed parsing might not be useful. But I know zstd will want negative levels, so maybe this is fine.

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -648,6 +649,34 @@ static int process_one_key_value_pair(const char *key, unsigned int value,
...
+ * @value[in]: The string (possibly only a null byte) recording the level
+ * @level_ptr[out]: The pointer to return the level into, if one is provided.
+ *
+ * Does not log the error reason.
+ *
+ * Return: 0 or -EINVAL.
+ */
+static int parse_compression_level(const char *value, int *level_ptr)
+{
+	int ret;
```
lorelei-sakai, line 663: A minor nit: We prefer "result" instead of "ret" for this kind of case.

doc/vdo.rst
```diff
@@ -168,7 +168,16 @@ Miscellaneous parameters:
...
+		acceptable values are 'on' and 'off'. It is preferred to
+                use the compressType parameter.
+
+        compressType:
+                The compression type, if any, to use. Acceptable values are
+                'none', the default; 'lz4', which uses the same behavior
+                as the 'compression=on' parameter; and 'lz4:<level>' for
+                integer levels. If a level is specified, this is used as the
+                acceleration factor for lz4. See lz4 docs for the specific
+                meanings of various acceleration factors.
```
lorelei-sakai, line 180: In form, this is a reasonable doc modification.

In content, it is not correct; the default is to use lz4 compression unless it is explicitly disabled.

lorelei-sakai (commented):
I'm actually working on a series to enable lz4 compresion levels, including what you're doing here. However, I've added some review contents to answer your qestions about whether this is the right way to change things. (In broad strokes this is basically right.)

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -172,10 +172,10 @@ static const char * const ADMIN_PHASE_NAMES[] = {
 };
 
 /* If we bump this, update the arrays below */
-#define TABLE_VERSION 4
+#define TABLE_VERSION 5
 
 /* arrays for handling different table versions */
-static const u8 REQUIRED_ARGC[] = { 10, 12, 9, 7, 6 };
+static const u8 REQUIRED_ARGC[] = { 10, 12, 9, 7, 6, 6};
```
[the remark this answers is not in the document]
sweettea: Ah, I meant to not be combining deprecating the old argument, clearly I didn't understand the new table line requirements.

src/c++/vdo/base/dm-vdo-target.c
sweettea: Shrug. We can constrain to LZ4_ACCELERATION_DEFAULT to LZ4_ACCELERATION_MAX if you want?

doc/vdo.rst
sweettea: I think it is correct as written, though the lvm default may be on?

 The doc at [line 170](https://github.com/dm-vdo/vdo-devel/blob/main/doc/vdo.rst?plain=1#L170) records the default as off; and [the code](https://github.com/dm-vdo/vdo-devel/blob/main/src/c%2B%2B/vdo/base/dm-vdo-target.c#L829) also seems to record the default as off.

src/c++/vdo/base/dm-vdo-target.c
sweettea: I had been thinking of vdo->compressing as the on/off boolean, while device_config->compression was the algorithm type -- but it's a good point that we do, at least for now, need to support `compression=off compressType=lz4:230` so that makes sense.

doc/vdo.rst
lorelei-sakai: ... Apologies. You are correct. I have apparently had a misunderstanding of what our default is, for years.

PR #268, opened 2025-02-26
[VDO-5890] Fix issue with auto activation

bjohnsto:
The addition of auto deactivation within the lvcreate calls create scenarios where vdoformat is not called until the device is activated. This means the timing between creation and wiping of old formatting through vdoformat calls is now changed, causing certain tests to fail.

This commit fixes the issue by calling wipefs on the VDO storage device in the configure() method of VDO.pm. This will wipe the geometry block; allowing calls like lvresize to pass because they do not find any old signature there.

We log the output of the wipefs call because it is very useful to see what is being wiped or not.


lorelei-sakai (approved):
I probably would have added a blank line of two (like after the SUPER call) to make it feel less cramped, but otherwise looks good.

PR #269, opened 2025-02-28
[VDO-5892] Add compressionType option to set lz4 acceleration factor

lorelei-sakai:
This PR is not quite complete. Notably, the testing is thin. I wanted to get this out for review though while I try to figure out how to test the new option. Suggestions on this point are welcome.

I also hope this framework will be useful for implementing zstd compression support (PR 241).
This PR supersedes PR 265, though I did borrow a few parts of that series here.

One of the bigger changes here is adding the acceleration parameter to the userspace lz4 implementation we use in tests. Although it works for now, in the long run it would probably make sense to bring it in line with the rest of our kernel fakes and port the existing kernel implementation to userspace instead of the current strategy. Is that worth doing here as part of this series?

### DIFF
--- a/doc/vdo.rst
+++ b/doc/vdo.rst
@@ -170,6 +170,17 @@ Miscellaneous parameters:
 		Whether compression is enabled. The default is 'off'; the
 		acceptable values are 'on' and 'off'.
 
+	compressionType:
+		The compression type to use. This value must begin with a
+		string indicating which compresion algorithm to use when
+		compression is enabled. It may also contain a ':' after the
+		algorithm, followed by options specific to the chosen
+		algorithm. The only supported algorithm is 'lz4'.
+
+		For lz4, there is one option, which is the acceleration factor
+		to use when compressing. For example, 'lz4:3' will use an
+		acceleration factor of 3.
+
 Device modification
 -------------------
 
@@ -242,11 +253,11 @@ All vdo devices accept messages in the form:
 
 ::
 
-        dmsetup message <target-name> 0 <message-name> <message-parameters>
+	dmsetup message <target-name> 0 <message-name> <message-parameters>
 
 The messages are:
 
-        stats:
+	stats:
 		Outputs the current view of the vdo statistics. Mostly used
 		by the vdostats userspace program to interpret the output
 		buffer.
@@ -270,7 +281,7 @@ The messages are:
 			default: Equivalent to 'queues vdo'
 			all: All of the above.
 
-        dump-on-shutdown:
+	dump-on-shutdown:
 		Perform a default dump next time vdo shuts down.
 
 
--- a/src/c++/vdo/base/dm-vdo-target.c
+++ b/src/c++/vdo/base/dm-vdo-target.c
@@ -9,6 +9,7 @@
 #include <linux/delay.h>
 #include <linux/device-mapper.h>
 #include <linux/err.h>
+#include <linux/lz4.h>
 #include <linux/module.h>
 #include <linux/mutex.h>
 #include <linux/spinlock.h>
@@ -178,6 +179,8 @@ static const char * const ADMIN_PHASE_NAMES[] = {
 static const u8 REQUIRED_ARGC[] = { 10, 12, 9, 7, 6 };
 /* pool name no longer used. only here for verification of older versions */
 static const u8 POOL_NAME_ARG_INDEX[] = { 8, 10, 8 };
+
+#define VDO_COMP_LZ4 "lz4"
 
 /*
  * Track in-use instance numbers using a flat bit array.
@@ -449,6 +452,43 @@ static inline int __must_check parse_bool(const char *bool_str, const char *true
 }
 
 /**
+ * parse_compression() - Parse a conpression configuration string.
+ * @string: The string value describing compression options.
+ * @config: The configuration data structure to update.
+ *
+ * Return: VDO_SUCCESS or an error.
+ */
+static int __must_check parse_compression(const char *string,
+					  struct device_config *config)
+{
+	char *delimiter = strchr(string, ':');
+	char *option_string;
+	int key_length = strlen(string) - 1;
+	int result;
+
+	if (delimiter) {
+		key_length = delimiter - string;
+		option_string = &delimiter[1];
+	}
+
+	if ((strncmp(string, VDO_COMP_LZ4, key_length) == 0) && delimiter) {
+		result = kstrtoint(&delimiter[1], 10,
+				   &config->compression_level);
+		if (result) {
+			vdo_log_error("optional config string error: integer needed, found \"%s\"",
+				      option_string);
+			return VDO_BAD_CONFIGURATION;
+		}
+	} else {
+		vdo_log_error("optional config string error: unknown compression type \"%s\"",
+			      string);
+		return VDO_BAD_CONFIGURATION;
+	}
+ 
+	return VDO_SUCCESS;
+}
+
+/**
  * process_one_thread_config_spec() - Process one component of a thread parameter configuration
  *				      string and update the configuration data structure.
  * @thread_param_type: The type of thread specified.
@@ -669,13 +709,17 @@ static int parse_one_key_value_pair(const char *key, const char *value,
 	if (strcmp(key, "compression") == 0)
 		return parse_bool(value, "on", "off", &config->compression);
 
-	/* The remaining arguments must have integral values. */
+	if (strcmp(key, "compressionType") == 0)
+		return parse_compression(value, config);
+
+	/* The remaining arguments must have positive integral values. */
 	result = kstrtouint(value, 10, &count);
 	if (result) {
-		vdo_log_error("optional config string error: integer value needed, found \"%s\"",
+		vdo_log_error("optional config string error: positive integer needed, found \"%s\"",
 			      value);
-		return result;
+		return VDO_BAD_CONFIGURATION;
 	}
+
 	return process_one_key_value_pair(key, count, config);
 }
 
@@ -827,6 +871,7 @@ static int parse_device_config(int argc, char **argv, struct dm_target *ti,
 	config->max_discard_blocks = 1;
 	config->deduplication = true;
 	config->compression = false;
+	config->compression_level = LZ4_ACCELERATION_DEFAULT;
 
 	arg_set.argc = argc;
 	arg_set.argv = argv;
@@ -1144,11 +1189,11 @@ static void vdo_status(struct dm_target *ti, status_type_t status_type,
 		vdo_fetch_statistics(vdo, &vdo->stats_buffer);
 		stats = &vdo->stats_buffer;
 
-		DMEMIT("/dev/%pg %s %s %s %s %llu %llu",
+		DMEMIT("/dev/%pg %s %s %s %s %s %llu %llu",
 		       vdo_get_backing_device(vdo), stats->mode,
 		       stats->in_recovery_mode ? "recovering" : "-",
 		       vdo_get_dedupe_index_state_name(vdo->hash_zones),
-		       vdo_get_compressing(vdo) ? "online" : "offline",
+		       vdo_get_compressing(vdo) ? "online" : "offline", VDO_COMP_LZ4,
 		       stats->data_blocks_used + stats->overhead_blocks_used,
 		       stats->physical_blocks);
 		mutex_unlock(&vdo->stats_mutex);
@@ -3104,7 +3149,7 @@ static void vdo_resume(struct dm_target *ti)
 static struct target_type vdo_target_bio = {
 	.features = DM_TARGET_SINGLETON,
 	.name = "vdo",
-	.version = { 9, 1, 0 },
+	.version = { 9, 2, 0 },
 #ifdef __KERNEL__
 	.module = THIS_MODULE,
 #endif /* __KERNEL__ */
--- a/src/c++/vdo/base/types.h
+++ b/src/c++/vdo/base/types.h
@@ -223,13 +223,14 @@ struct device_config {
 	 * of 4K blocks regardless of the value of the logical_block_size parameter below.
 	 */
 	block_count_t logical_blocks;
+	block_count_t max_discard_blocks;
 	unsigned int logical_block_size;
 	unsigned int cache_size;
 	unsigned int block_map_maximum_age;
-	bool deduplication;
-	bool compression;
 	struct thread_count_config thread_counts;
-	block_count_t max_discard_blocks;
+	int compression_level;
+	bool compression;
+	bool deduplication;
 };
 
 enum vdo_completion_type {
--- a/src/c++/vdo/fake/linux/kstrtox.h
+++ b/src/c++/vdo/fake/linux/kstrtox.h
@@ -13,9 +13,8 @@
 #include <errno.h>
 #include <stdlib.h>
 
-int
-kstrtouint(const char *string, unsigned int base, unsigned int *result);
-int
-kstrtoull(const char *s, unsigned int base, u64 *res);
+int kstrtoint(const char *s, unsigned int base, int *result);
+int kstrtouint(const char *string, unsigned int base, unsigned int *result);
+int kstrtoull(const char *s, unsigned int base, u64 *res);
 
 #endif // LINUX_KSTRTOX_H
--- a/src/c++/vdo/tests/vdoTestBase.c
+++ b/src/c++/vdo/tests/vdoTestBase.c
@@ -555,6 +555,12 @@ static void addUInt64(char **arg, uint64_t u)
 }
 
 /**********************************************************************/
+static void addCompressionType(char **arg, const char *s, int32_t d)
+{
+  CU_ASSERT(asprintf(arg, "%s:%d", s, d) != -1);
+}
+
+/**********************************************************************/
 static TestConfiguration fixThreadCounts(TestConfiguration configuration)
 {
   struct thread_count_config *threads = &configuration.deviceConfig.thread_counts;
@@ -623,6 +629,9 @@ static int makeTableLine(TestConfiguration configuration, char **argv)
   addString(&argv[argc++],
             (configuration.deviceConfig.compression ? "on" : "off"));
 
+  addString(&argv[argc++], "compressionType");
+  addCompressionType(&argv[argc++], "lz4", 
+                     configuration.deviceConfig.compression_level);
   return argc;
 }
 
--- a/src/perl/Permabit/BlockDevice/VDO/LVMVDO.pm
+++ b/src/perl/Permabit/BlockDevice/VDO/LVMVDO.pm
@@ -91,6 +91,12 @@ sub makeLVMConfigString {
       push(@values, ["vdo_use_compression", $self->{enableCompression}]);
     }
   }
+
+  # lvm does not support this feature yet (VDO-5892)
+  if (defined($self->{compressionType})) {
+    push(@optional, ["vdo_compression_type", "$self->{compressionType}"]);
+  }
+
   if (defined($self->{enableDeduplication})) {
     # magic value -1 suppresses the option completely
     if ($self->{enableDeduplication} != -1) {
--- a/src/c++/vdo/base/data-vio.c
+++ b/src/c++/vdo/base/data-vio.c
@@ -1839,6 +1839,7 @@ static void pack_compressed_data(struct vdo_completion *completion)
 static void compress_data_vio(struct vdo_completion *completion)
 {
 	struct data_vio *data_vio = as_data_vio(completion);
+	struct vdo *vdo = vdo_from_data_vio(data_vio);
 	int size;
 
 	assert_data_vio_on_cpu_thread(data_vio);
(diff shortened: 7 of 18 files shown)


src/c++/vdo/base/types.h
```diff
@@ -223,13 +223,14 @@ struct device_config {
 	 * of 4K blocks regardless of the value of the logical_block_size parameter below.
 	 */
 	block_count_t logical_blocks;
+	block_count_t max_discard_blocks;
```
sweettea, line 226: why is this reordered?

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -669,13 +709,17 @@ static int parse_one_key_value_pair(const char *key, const char *value,
...
 
-	/* The remaining arguments must have integral values. */
+	if (strcmp(key, "compressionType") == 0)
+		return parse_compression(value, config);
+
+	/* The remaining arguments must have positive integral values. */
 	result = kstrtouint(value, 10, &count);
 	if (result) {
-		vdo_log_error("optional config string error: integer value needed, found \"%s\"",
+		vdo_log_error("optional config string error: positive integer needed, found \"%s\"",
```
sweettea, line 718: nonnegative

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -1144,11 +1189,11 @@ static void vdo_status(struct dm_target *ti, status_type_t status_type,
 		vdo_fetch_statistics(vdo, &vdo->stats_buffer);
 		stats = &vdo->stats_buffer;
 
-		DMEMIT("/dev/%pg %s %s %s %s %llu %llu",
+		DMEMIT("/dev/%pg %s %s %s %s %s %llu %llu",
 		       vdo_get_backing_device(vdo), stats->mode,
 		       stats->in_recovery_mode ? "recovering" : "-",
 		       vdo_get_dedupe_index_state_name(vdo->hash_zones),
-		       vdo_get_compressing(vdo) ? "online" : "offline",
+		       vdo_get_compressing(vdo) ? "online" : "offline", VDO_COMP_LZ4,
```
sweettea, line 1196: maybe lz4:{config->compression_level}({vdo_get_compressing(vdo)?"on":"off"}) ?

(I admit changing online/offline is bigger than adding another thing -- but also it might make sense to change format for dedupe also -- dedupe online/offline made sense when there was a socket connection involved; compression just copied it; but it doesn't make sense for either anymore imho, especially for compression.)

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -448,6 +451,43 @@ static inline int __must_check parse_bool(const char *bool_str, const char *true
...
+	char *option_string;
+	int key_length = strlen(string) - 1;
+	int result;
+
+	if (delimiter) {
+		key_length = delimiter - string;
+		option_string = &delimiter[1];
+	}
+
+	if ((strncmp(string, VDO_COMP_LZ4, key_length) == 0) && delimiter) {
```
sweettea, line 474: This doesn't work for e.g. "dinosaursarenotpotatoes:elephant" -- it overruns the VDO_COMP_LZ4 string.

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -448,6 +451,43 @@ static inline int __must_check parse_bool(const char *bool_str, const char *true
...
+		result = kstrtoint(&delimiter[1], 10,
+				   &config->compression_level);
```
sweettea, line 476: if I'm reading the code right, "lz4:" will get parsed as compression level 0 -- kstrtoint seems to parse "\0" as 0, which is probably not what we want.

doc/vdo.rst
```diff
@@ -170,6 +170,17 @@ Miscellaneous parameters:
...
+	compressionType:
+		The compression type to use. This value must begin with a
+		string indicating which compresion algorithm to use when
+		compression is enabled. It may also contain a ':' after the
+		algorithm, followed by options specific to the chosen
+		algorithm. The only supported algorithm is 'lz4'.
+
+		For lz4, there is one option, which is the acceleration factor
+		to use when compressing. For example, 'lz4:3' will use an
+		acceleration factor of 3.
```
sweettea, line 182: I think this is unnecessarily roundabout and generalized. Even if we're moving to a universe with more than one option for an algorithm, or an option that isn't the level, we don't need to say what we'd do in that universe until we get there.

I still prefer the text from #265, which imho elaborates exactly what acceptable values there are at the present day only.

doc/vdo.rst
```diff
@@ -270,7 +281,7 @@ The messages are:
 			default: Equivalent to 'queues vdo'
 			all: All of the above.
 
-        dump-on-shutdown:
+	dump-on-shutdown:
```
sweettea, line 284: Unintentional whitespace changes?

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -669,13 +709,17 @@ static int parse_one_key_value_pair(const char *key, const char *value,
...
+		return parse_compression(value, config);
+
+	/* The remaining arguments must have positive integral values. */
 	result = kstrtouint(value, 10, &count);
 	if (result) {
-		vdo_log_error("optional config string error: integer value needed, found \"%s\"",
+		vdo_log_error("optional config string error: positive integer needed, found \"%s\"",
 			      value);
-		return result;
+		return VDO_BAD_CONFIGURATION;
```
sweettea, line 720: this is not connected to adding the option -- perhaps split this into a general cleanup change as it's orthogonal.

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -448,6 +451,43 @@ static inline int __must_check parse_bool(const char *bool_str, const char *true
...
+		}
+	} else {
+		vdo_log_error("optional config string error: unknown compression type \"%s\"",
+			      string);
+		return VDO_BAD_CONFIGURATION;
+	}
+ 
+	return VDO_SUCCESS;
+}
+
```
sweettea, line 490: I mostly prefer the version from #265 . I appreciate the elegance of searching for the delimiter and think that's a great addition.

I think strchrnul might be better than strchr since then there's an unconditional [string:delimiter) key string in that case -- can get rid of the if.

I think naming it option_string, when there's only one option right now, is confusingly general.

I think &delimiter[1] is not really kernel style, and that delimiter + 1 more clearly conveys the intent of advancing one past the delimiter.

src/perl/Permabit/BlockDevice/VDO/LVMVDO.pm
```diff
@@ -91,6 +91,12 @@ sub makeLVMConfigString {
       push(@values, ["vdo_use_compression", $self->{enableCompression}]);
     }
   }
+
+  # lvm does not support this feature yet (VDO-5892)
+  if (defined($self->{compressionType})) {
+    push(@optional, ["vdo_compression_type", "$self->{compressionType}"]);
```
sweettea, line 97: c/p error from Unmanaged? @values not @optional?

sweettea (approved):
It seems like LZ4_compress_default isn't actually used after the next change -- is it worth removing it?

sweettea:
Mostly looks good. A couple differences from how I envisioned it, but mostly in line with how I thought it would go.

As for testing: RebuildStress03 has a compression toggle -- my first thought is to expand that into adjusting compression levels randomly.

Compress01 with two specified non-standard levels might be nice.
Some invalid levels in VDOLoadFailure01.

I don't think there's a good way to test different accel levels giving different results. Nor do I think it's super important; as long as we unit test that the right level gets passed to LZ4_compress_fast, and that specifying a nonstandard level in a VDOTest works out, I think it's enough.

src/c++/vdo/base/types.h
lorelei-sakai: I was trying to pack device_config a little better. But we only make on of these per table, I should prioritize sensible groupings instead.

src/c++/vdo/base/dm-vdo-target.c
lorelei-sakai: I was thinking printing he algorithm was enough. If we print the options, too. I'll think about how to change his.

doc/vdo.rst
lorelei-sakai: I'm trying to provide a template so it's easy to see where/how to update for when adding zstd. I suppose we could defer that change until later. Any objections to this sort of format when we do add zstd?

doc/vdo.rst
lorelei-sakai: Intentional, cleaning up formatting nits while we're here. (spaces vs tabs for all of these.)

src/c++/vdo/base/dm-vdo-target.c
lorelei-sakai: It's ultimately irrelevant, I was just trying to be more consistent. I can drop this.

src/perl/Permabit/BlockDevice/VDO/LVMVDO.pm
lorelei-sakai: Yes, this is why CI tests failed. Teach me to push a commit without running checkin tests.

lorelei-sakai:
> It seems like LZ4_compress_default isn't actually used after the next change -- is it worth removing it?

The function isn't used in our code but it does exist in the kernel. I don't see a lot of benefit in removing it in our implementation.

src/c++/vdo/base/dm-vdo-target.c
lorelei-sakai: delimiter + 1: I somehow convinced myself this didn't work, but of course it does. I'll change it.

option_string: Like in the doc, I'm trying to set the framework for when the interpretation is dependent on the algorithm. In this case I'm not sure there is a better name, though.

strchrnul: I think this makes things more confusing when we go to parse the option string, since we have to check if we're already pointing at the final null. (Can't do delimiter + 1 in that case. At least not safely.) So we still have to branch for delimiter vs no delimiter. I'll think about this.

And neither way deals well with the degenerate "lz4:" case. Parsing the option as 0 works fine but you're right that we should maybe explicitly complain in that case.

src/c++/vdo/base/dm-vdo-target.c
lorelei-sakai: True enough. Though I'm fixing a bug; in the current implementation, it actually won't allow 0 as a valid result...

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -448,6 +451,43 @@ static inline int __must_check parse_bool(const char *bool_str, const char *true
...
+					  struct device_config *config)
+{
+	char *delimiter = strchr(string, ':');
+	char *option_string;
+	int key_length = strlen(string) - 1;
+	int result;
+
+	if (delimiter) {
+		key_length = delimiter - string;
+		option_string = &delimiter[1];
```
bjohnsto, line 471: If someone put in "lz4:" as the value, I'm trying to think what delimiter[1] would be? The space between "lz4:" and the next key/value pair?

src/c++/vdo/tests/vdoTestBase.c
```diff
@@ -623,6 +629,9 @@ static int makeTableLine(TestConfiguration configuration, char **argv)
   addString(&argv[argc++],
             (configuration.deviceConfig.compression ? "on" : "off"));
 
+  addString(&argv[argc++], "compressionType");
+  addCompressionType(&argv[argc++], "lz4",
```
bjohnsto, line 633: If you move the define of the string "lz4" to some place common, it could be used here.

src/c++/vdo/base/types.h
```diff
@@ -223,13 +223,14 @@ struct device_config {
...
 	unsigned int logical_block_size;
 	unsigned int cache_size;
 	unsigned int block_map_maximum_age;
-	bool deduplication;
-	bool compression;
 	struct thread_count_config thread_counts;
-	block_count_t max_discard_blocks;
+	int compression_level;
```
bjohnsto, line 231: I get that we don't technically need a compression_type field (since we still only support lz4) but having compressionType on the table line and not storing it anywhere just seems weird to me. Would it hurt if we had compression_type stored in the device_config now, rather than adding it later on? Are we not sure what type of field we want there?

bjohnsto (changes requested):

src/c++/vdo/base/dm-vdo-target.c
raeburn: The string being parsed here is null terminated, so delimiter[1] would be the \0 byte.

src/c++/vdo/fake/linux/kstrtox.h
```diff
@@ -13,9 +13,8 @@
 #include <errno.h>
 #include <stdlib.h>
 
-int
-kstrtouint(const char *string, unsigned int base, unsigned int *result);
-int
-kstrtoull(const char *s, unsigned int base, u64 *res);
+int kstrtoint(const char *s, unsigned int base, int *result);
```
raeburn, line 16: The __must_check annotation on the function definition won't be useful outside that one C file unless you also put the attribute on the declaration in the header file too.

src/c++/vdo/base/dm-vdo-target.c
lorelei-sakai: This is not true, delimiter + 1 is the original string starting after the delimiter. "delimiter" is a pointer into the original string if the delimiter is found.

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -179,6 +180,8 @@ static const u8 REQUIRED_ARGC[] = { 10, 12, 9, 7, 6 };
 /* pool name no longer used. only here for verification of older versions */
 static const u8 POOL_NAME_ARG_INDEX[] = { 8, 10, 8 };
 
+#define VDO_COMP_LZ4 "lz4"
```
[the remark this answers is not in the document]
lorelei-sakai: This string is used for the table line and for status; nothing outside of the file should care.
later on, the super block changes might wantto refer to this in sone way too, in which case we can move it when we do that work.

src/c++/vdo/base/dm-vdo-target.c
lorelei-sakai: On reflection, I think you're right that we should provide more information here. Respelled.

src/c++/vdo/base/types.h
lorelei-sakai: I figured it would be an enum. We're only going to have a limited number of algorithms that we'll support. But also it feels super weird to me to add a field that only ever has one value. I think we can add this later when adding zstd. (And it'll probably be clearer how to spell it when that work is done.)

lorelei-sakai: Reconsidered. If we define the valid strings as constants, the device_config field can be const char* and just point to the relevant string constant. That's probably more convenient for most uses. We'll still want to have feature flags for each supported algorithm, but there's no reason to tie those directly to this value.

src/c++/vdo/tests/vdoTestBase.c
lorelei-sakai: [long thinking out loud cut] I put this in constants.h.

src/c++/vdo/base/dm-vdo-target.c
lorelei-sakai: I've refactored this. I'm not sure it's simpler, but it will now catch and reject malformed strings such as "lz",  "lz4+", and "lz4:" I'm not sure strchrnul is really better, mostly because I'm ambivalent about how to spell ( *delimiter | delimiter[0] | *delimiter != '\0' ), but I've used it anyway. The option_string branching looks messier but the key_length doesn't need a branch now. (And I can't figure out how to not need the key_length.)

doc/vdo.rst
sweettea: I am deeply baffled why github keeps eating my comments.

Please don't put distant formatting fixups in the same commit, and if you do, please note them in the change description.

src/c++/vdo/base/dm-vdo-target.c
sweettea: @lorelei-sakai have you pushed the new version? I can't see it.

sweettea (changes requested):

src/c++/vdo/base/dm-vdo-target.c
lorelei-sakai: Just pushed sorry. i was going through comments making usre I addressed everything.

src/c++/vdo/base/types.h
sweettea: I'm not a fan of the pointers to string constants... we'll need the enum to store in the zstd compressed blocks anyway, and comparing string pointer equality feels wrong even if it'd be acceptable in this case.

src/c++/vdo/base/dm-vdo-target.c
sweettea: I do like the new version more, but I think https://github.com/sweettea/vdo-devel/blob/956f163a62bc5c4bab3caded3cf39a836749e92c/src/c%2B%2B/vdo/base/dm-vdo-target.c#L702 is still a little more elegant by putting parsing of $delimiter-onward into its own function.

doc/vdo.rst
lorelei-sakai: I'm not sure what aspect of the doc you want me to incorporate. I can't just take the doc from #265 because it describes a feature that doesn't work like what I wrote.

doc/vdo.rst
lorelei-sakai: doc split into its own commit, and format changes noted.

src/c++/vdo/base/dm-vdo-target.c
lorelei-sakai: I did consider using a separate function. It just didn't seem like there was quite enough there to justify it. But I'll admit it's a close call.

Partially I split it the way I did in order to look forward to to adding zstd, so the common parsing bits (alg:options split) happen before checking for a specific algorithm.

src/c++/vdo/base/types.h
lorelei-sakai: Ah, you're right, I forgot we'll need an enum to put in the compressed block format. I mean we could use the string constants for that but it's that's silly. I'll think about that when doing the feature flags bit.

src/c++/vdo/fake/linux/kstrtox.h
lorelei-sakai: addressed.

doc/vdo.rst
lorelei-sakai: re github eating comments: I have no idea. I do get mail notifications though, so I do still see them. I don't think there's anything I missed at the moment.

src/c++/vdo/base/dm-vdo-target.c
```diff
@@ -1144,10 +1185,11 @@ static void vdo_status(struct dm_target *ti, status_type_t status_type,
 		vdo_fetch_statistics(vdo, &vdo->stats_buffer);
 		stats = &vdo->stats_buffer;
 
-		DMEMIT("/dev/%pg %s %s %s %s %llu %llu",
+		DMEMIT("/dev/%pg %s %s %s %s:%d(%s) %llu %llu",
```
raeburn, line 1188: As long as we're changing the format of the status output here, how about changing online/offline to enabled/disabled? "Offline compression" is sometimes used to refer to _doing_ compression but as post-processing rather than as part of the main processing of the request. (Same issue for "offline deduplication".)

src/c++/vdo/tests/vdoTestBase.c
bjohnsto: Sounds good

doc/vdo.rst
sweettea: compressType:
                The compression type, if any, to use. Acceptable values are
                'lz4', which uses the same behavior
                as the 'compression=on' parameter; and 'lz4:<level>' for
                integer levels. If a level is specified, this is used as the
                acceleration factor for lz4. See lz4 docs for the specific
                meanings of various acceleration factors.

src/c++/vdo/base/dm-vdo-target.c
sweettea: Fine reasoning. Thanks.

src/c++/vdo/base/dm-vdo-target.c
lorelei-sakai: This make sense. I've changed this to simply be "on" or "off". This matches what you can pass in for the compression parameter.

raeburn: Is the deduplication status still online/offline? It's got the same issue and it might make sense to change it now as long as parsers are likely to need updating.

lorelei-sakai: Okay, first off, this is completely unrelated to this series, which is about compression. I don't want feature creep here, if you think it's worth investigating open another issue to do that.

The index state has six possible (non-error) states, one of which is "offline". It's not at all obvious to me what a better label would be to communicate that state. (It means the index is all set up to process dedupe requests, but we aren't sending any to it.)

raeburn: Okay, makes sense. I was just thinking of parsers needing changing fewer times, but I doubt there are many tools out there examining those fields anyway.

raeburn (approved):

bjohnsto (approved):
All my comments were taken care of

PR #270, opened 2025-03-04
Replace EXTRA_CFLAGS with ccflags-y

C2Redhat:
Based on:
https://www.kernel.org/doc/Documentation/kbuild/makefiles.txt

EXTRA_FLAGS has been deprecated. We should be using ccflags-y. Replaced EXTRA_CFLAGS with ccflags-y in all our files.

### DIFF
--- a/src/c++/defines
+++ b/src/c++/defines
@@ -32,7 +32,7 @@ ifdef IWYU
   else
     export CC := include-what-you-use
   endif
-  EXTRA_CFLAGS = -Xiwyu --no_fwd_decls
+  ccflags-y = -Xiwyu --no_fwd_decls
   MAKEFLAGS := $(MAKEFLAGS) -k
 endif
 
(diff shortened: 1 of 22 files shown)


src/c++/defines
sweettea, line 1: I'm pretty sure this flie is only used in usrespace so I don't think it needs to change.

src/packaging/src-dist/user/utils/vdo/Makefile
sweettea, line 1: This file is only used in userspace -- so I don't think it needs the ccflags-y change.

sweettea (commented):
Mostly lgtm, just two files seem unneeded.

src/packaging/src-dist/user/utils/vdo/Makefile
lorelei-sakai: I think this is probably correct, but being consistent might help keep us from getting confused in the future.

src/c++/defines
lorelei-sakai: This file does get included in a lot of our other Makefiles.

lorelei-sakai (commented):
Some of these changes may not be strictly necessary, but it's probably better to change everything rather than try to sort out exactly which ones care about this change and which don't.

Also, what happened with the CI tests? They show a failure but I can't figure out what that failure is.

One question is when did the deprecation/changeover happen? Would it make sense for us to backport this to older versions? (I think probably not, there's no reason for the kernel to backport this kind of build change.)

src/c++/defines
sweettea: A reasonable argument.

src/packaging/src-dist/user/utils/vdo/Makefile
sweettea: A good argument, thanks.

sweettea (approved):
Okay, I can buy the argument that it's consistent everywhere, so it's fine by me.

sweettea (changes requested):
One tiny thing: the actual title of the change refers to EXTRA_FLAGS, not EXTRA_CFLAGS.

PR #271, opened 2025-03-10
[TECH-189] Change RHEL_RELEASE_VERSION comparison to exclude RHEL10

C2Redhat:
RHEL10 does not have a us_to_ktime definition, we need to exclude RHEL10 in VDO_USE_NEXT definition.


lorelei-sakai (approved):
Looks good to me. The PR title has weird spacing in "[ TECH-189]" but the change itself looks good.

PR #272, opened 2025-03-12
[TECH-189] Update run OS class from RHEL9 to RHEL10

C2Redhat:


lorelei-sakai (approved):
The PR or commit message could use a bit more explanation about why we're doing this, but it's not a big deal. (and the ticket reference does give us some information.)

PR #275, opened 2025-03-18
[VDO-5864] Remove unused sysfs fields from tests

bjohnsto:
This is the first PR in a series that will remove sysfs from our internal code base. This initial PR is a series of commits; each of which will handle one type of sysfs field or test. Additional PRs will complete the remaining removal from our tests and then the actual removal of the sysfs field creation.


src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -779,23 +779,31 @@ sub waitForIndex {
...
-# @return the contents of the sysfs compression entry for the device
+# @return the value of the compression state from dmsetup status call
 ##
 sub getVDOCompressionStatus {
   my ($self) = assertNumArgs(1, @_);
-  my $machine = $self->getMachine();
-  return $machine->catAndChomp($self->getSysModuleDevicePath("compressing"));
+  my $status = $self->getStatus();
+  $status =~ m/(\S+\s){7}(\S+)/;
+
```
lorelei-sakai, line 782: I started thinking about this regex and then I realized that using split() is a more perl-ish way to do this.
my @fields = split(' ', $status); return $fields[7];

Both here and in getVDODedupeStatus().
(Also making a note for me, I'll need to update this in my PR since I'm changing the status output.)

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -403,12 +403,6 @@ sub postActivate {
     $self->setAffinityList($self->{vdoAffinityList});
   }
 
-  # Set parameters that are device specific
-  my $machine = $self->getMachine();
-  if (defined($self->{vdoMaxDiscardsActive})) {
-    $machine->setProcFile($self->{vdoMaxDiscardsActive},
-                          $self->getSysModuleDevicePath("discards_limit"));
-  }
   # These parameters are only defined on non-release builds
```
lorelei-sakai, line 406: If you're going to remove the use of it, you should remove the field from the VDO device entirely.
Alternatively (and probably what I'd prefer), switch this to setting the maxDiscards value through the --vdosettings options on lvm. (lvchange, maybe?)

I don't like having the parameter defined in the device but not used, though.

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -406,8 +406,8 @@ sub postActivate {
...
   eval {
     foreach my $histogram (keys(%LATENCY_CHECKS)) {
-      $machine->setProcFile(secondsToMS($self->{latencyLimit}),
-                            $self->getSysModuleDevicePath("$histogram/limit"));
+      $self->getMachine()->setProcFile(secondsToMS($self->{latencyLimit}),
+                                       $self->getSysModuleDevicePath("$histogram/limit"));
```
lorelei-sakai, line 410: This change should really be part of commit 3; otherwise applying just that patch leaves an unloadable file. (I couldn't understand the logic of this until i looked back at that other commit.)

lorelei-sakai (changes requested):
This mostly looks pretty good. The main issue is in commit 3, we should either do more or less.

One minor nit, "remove unused sysfs fields" doesn't relaly feel like an accurate description to me, maybe something more like "remove reliance on sysfs fields in tests".

And one more even smaller nit, commit 1,2,3, and 7 have periods at the end of their titles, which they shouldn't.

src/perl/Permabit/BlockDevice/VDO.pm
bjohnsto: My idea was to clean up all uses one at a time then delete all the sysfs fields all at once at the end. I'm not sure yet if I have redone all uses. I could try to do it one sysfs field per commit if you prefer.

lorelei-sakai: I'm not sure I follow. I was saying that the VDO.pm object has a vdoMaxDiscardsActive parameters, and with this change you are making it so setting that parameter does nothing at all. If the parameter is doing nothing, we should remove it.
(I would prefer to have the parameter still do something, but it looks like none of our tests actually set it right now so removing it is probably fine.)

lorelei-sakai (changes requested):
Forgot to actually finish this yesterday somehow.
First two commit look good. A couple of nits in the commit messages (first line, the title, should have no capitalize words, rest of the commit message should be sentences ending with periods, try to be consistent.)

The histogram change seems like it would be a lot more readable if the new histogram implementation was separate from removing the sysfs version. (i.e., in separate commits). Also there appears to be some code motion that may be making the diffs more complicated than they need to be. Some choices in the new histogram implementation look suspect to me but it's possible that's just due to diffing difficulties (specifically in the histogram.c file).

src/c++/vdo/base/histogram.c
```diff
@@ -216,10 +214,19 @@ static const u64 bottom_value[1 + 10 * MAX_LOG_SIZE] = {
 	1000000000000L,
 };
 
-static unsigned int divide_rounding_to_nearest(u64 number, u64 divisor)
+ssize_t set_histogram_limit(struct histogram *h, const char *buf, ssize_t length)
```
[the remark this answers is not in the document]
lorelei-sakai: I've taken a closer look at the new histogram implementation. A couple of points:
a) The confusing part is pretty much limited to histogram.c; the rest of the change looks fine.
b) It looks like some code has been re-ordered and I don't understand the logic in the ordering, if there is one. For example, divide_rounding_to_nearest() has moved from before max_bucket to after it, creating diff churn because now it looks like a delete and an add even though the function itself is unchanged.
c) Several of the histogram writing functions have the same names as before even though the signatures have changed. Is there a reason not to leave them in the same order they used to be in?
d) There is a lot of boilerplate in write functions dealing with prefix and suffix that's largely duplicated. Is there a simpler way to structure this, for instance by using write_sstring() as a helper for other writing functions? It seem plausible (though possibly more annoying) to have a function like write_sstring that can take a format string & arguments so it can be used to implement all the more specific writing functions.

src/c++/vdo/base/histogram.c
```diff
@@ -793,12 +647,36 @@ void enter_histogram_sample(struct histogram *h, u64 sample)
 	}
 }
 
+/**
+ * write_histogram() - Writes histogram info into a bufer.
+ * @histogram: The histogram to write.
+ * @buf: The buffer to write into
+ * @maxlen: The max size of the buffer
+ */
+void write_histogram(struct histogram *histogram, char **buf, unsigned int *maxlen)
```
lorelei-sakai, line 656: Having to deal with maxlen is annoying but I suppose necessary. But did you measure the size of the histogram output? Will we have to worry about cases where we cut off some of the output due to hitting the maxlen?
For a message, the output buffer is usually 4K, right? Is that enough to print most histogram cases?

lorelei-sakai (changes requested):
Most things look fine. There are some relatively trivial formatting nits.

I think I would still like to see the new histogram-writing code as a separate commit. It's a little hard to correlate the new code with what it's replacing now, and if you implement some of the suggestions I had about not duplicating code, it will probably be even more different from the original. But the mess seems confined to histogram.c and it could be reviewed as-is if splitting the commit is too hard.

src/c++/vdo/base/vdo-histograms.c
```diff
@@ -29,50 +26,122 @@ void vdo_initialize_histograms(struct kobject *parent, struct vdo_histograms *hi
...
+		return;
+	}
+}
+
+void vdo_write_histograms(struct vdo_histograms *histograms, char **buf,
+			  unsigned int *maxlen)
+{
+	write_sstring(NULL, "{ ", NULL, buf, maxlen);
+	write_histogram(histograms->discard_ack_histogram, buf, maxlen);
+	write_histogram(histograms->flush_histogram, buf, maxlen);
```
lorelei-sakai: Given the size I recall for histograms, I wonder if it actually makes sense to have one message per histogram. So put a parameter on the message "dmsetup message histogram read_bios" or something with the name. I'm a little skeptical that we'll get all 12 histograms into a single message buffer. (How big are those anyway? I was assuming 4k.)

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -1219,59 +1218,51 @@ sub logHumanVDOStats {
...
-    } @histograms;
-  }
+  # Histograms are subject to the VDO_INTERNAL build macro.  Release builds
+  # will not have histograms.
+  my $output = "no histograms";
+  eval {
+    $self->sendMessage("histograms");
+    $output = $machine->getStdout();
+  };
   $log->info("Logging histograms for $self->{vdoSymbolicPath}");
```
lorelei-sakai, line 1223: It probably makes more sense here to check for EVAL_ERROR and just return if you can't get histograms than to put a pointless line about "no histograms" into the log.

src/c++/vdo/base/vdo-histograms.c
```diff
@@ -18,127 +48,36 @@
...
-						   "writes", "I/O time", 5);
-	histograms->write_queue_histogram =
-		make_logarithmic_jiffies_histogram("write_queue", "Write Queue",
-						   "writes", "queue time", 5);
-	histograms->start_request_histogram =
-		make_logarithmic_jiffies_histogram("bio_start", "Start Request",
-						   "requests", "delay time", 5);
+	for (int i = 0; i < HISTOGRAM_LAST; i++) {
+		struct histogram_info info = histogram_list[i];
+		histograms->histogram[i] = make_logarithmic_jiffies_histogram(info.name, info.label, info.counted_items, info.metric, info.log_size);
```
lorelei-sakai, line 53: This line looks relaly long. Does it need to be wrapped?

src/c++/vdo/base/vdo-histograms.c
```diff
@@ -18,127 +48,36 @@
...
+			return;
+		}
 	}
 }
 
+void vdo_enter_histogram_sample(struct vdo_histograms *histograms,
+				enum histogram_types type, u64 sample)
+{
+	enter_histogram_sample(histograms->histogram[type], sample);
+}
```
lorelei-sakai, line 72: This is an unnecessary function. You whould just merge this into the top of enter_histogram_sample() and call that function directly.

src/c++/vdo/base/vdo-histograms.h
```diff
@@ -8,29 +8,33 @@
...
+	HISTOGRAM_FLUSH,
+	HISTOGRAM_ACKNOWLEDGE_READ,
+	HISTOGRAM_ACKNOWLEDGE_WRITE,
+	HISTOGRAM_ACKNOWLEDGE_DISCARD,
+	HISTOGRAM_BIO_READ,
+	HISTOGRAM_READ_QUEUE,
+	HISTOGRAM_BIO_WRITE,
+	HISTOGRAM_WRITE_QUEUE,
+	HISTOGRAM_BIO_START,
+	HISTOGRAM_LAST,
```
lorelei-sakai, line 24: I'd prefer to call this HISTOGRAM_COUNT instead of HISTOGRAM_LAST, which better reflects its usage most of the time. (You could also define both values if you prefer, just make sure they have the same value.)

src/c++/vdo/base/vdo-histograms.c
lorelei-sakai: In particular, the beginning of enter_histogram_sample() does a NULL check, which now cannot be true, I think. The equivalent problem is if you called this function with a type that is out-of-bounds, which you probably should check for. The cleanup will probably be made clearer by just combining the functions.

src/c++/vdo/base/io-submitter.c
```diff
@@ -123,10 +123,10 @@ static void send_bio_to_device(struct vio *vio, struct bio *bio)
...
 	count_all_bios(vio, bio);
 #ifdef VDO_INTERNAL
-	enter_histogram_sample(((bio_data_dir(bio) == WRITE) ?
-				histograms->write_queue_histogram :
-				histograms->read_queue_histogram),
-			       jiffies - vio->bio_submission_jiffies);
+	vdo_enter_histogram_sample(histograms, ((bio_data_dir(bio) == WRITE) ?
+						HISTOGRAM_WRITE_QUEUE :
+						HISTOGRAM_READ_QUEUE),
+				   jiffies - vio->bio_submission_jiffies);
```
lorelei-sakai, line 129: This line wrapping is super awkward. Whether the trinary needs to be three lines or whether it can be a single line, I'm put a line break after "histogram," so you don't need multiple indent levels here.
(With the new type constants, the trinary might fit on a single line. Or you could also call in out into a separate local if it's still too long.)

lorelei-sakai (changes requested):
I'm not quite sure understand the logic behind the way you split the histogram patches. The diffs are a little more readable, but there are still a lot of collisions between new code and the removal of old code.

I think what would make this clearer is if you split commit 3 into two separate commits:
Commit 3.1 adds all the rewrites functions and moves and deletes nothing.
Commit 3.2 deletes all the sysfs code that is unused after the changes in commit 1.
It would also help a lot if you kept the order of functions when you change the implementation of the functions, so the old and new code are adjacent in the diffs.
As an example of what I do not want to see, histogram_show_unit() moved from line 436 to line 277 for no apparent reason.
As an example of what I do want to see, histogram_show_unacceptable() is lined up so the old and new code are easy to compare. 
(I do like having the histogram list rearrangement in a separate commit, so commit 4 is nice.)

The first patch is fine, and the second, fourth, and fifth patches just have some minor issues. Be sure to check all the unresolved conversations to make sure you are addressing all my review points. There are some from older review cycles which did not seem to get addressed.

PR #277, opened 2025-03-21
[VDO-5896] Rename misleading index states and update tests

lorelei-sakai:
This idea came up while reviewing PR #269. It renames the 'online' and 'offline' index states, and updates the docs to include the 'suspended' state.

While updating tests, it also fixes a problem with test expectations for compression status output that doesn't seem to have a ticket yet.

### DIFF
--- a/doc/vdo.rst
+++ b/doc/vdo.rst
@@ -311,8 +311,8 @@ Status
 
 	index state:
 		The current state of the deduplication index in the vdo
-		volume; values may be 'closed', 'closing', 'error',
-		'offline', 'online', 'opening', and 'unknown'.
+		volume; values may be 'active', 'closed', 'closing',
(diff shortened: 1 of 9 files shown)


sweettea (approved):
Mostly looks good. I think the commit would be better as four -- base/dedupe.c, doc/vdo.rst, perl things, c test things; but either way whatever. 
Please feel free to add:

Reviewed-by: Sweet Tea Dorminy <sweettea-kernel@dorminy.me>

lorelei-sakai:
Splitting the doc from the C changes feels weird to me, especially since they are so small. Splitting out the user space stuff is more plausible. There did end up being more test changes than I had expect when I started this.

Anyway, thanks for the review.

raeburn (approved):

PR #287, opened 2025-04-10
[VDO-5892][VDO-5896][VDO-5897] Restore old status line output when compression type is not set

lorelei-sakai:
Comments from Zdenek make it clear that VDO can only change the status line output when we are passing new parameters. The status line output for device where whe only use existing parameters must be the same as it always was. So in the case of the compressionType parameter, we can provide more informative status output when the --compressionType parameter is set, but not when it is not set.

Commits 1-3 revert status line changes that break backwards compatibility. (and related documentation).
Commit 4-6 implement the new behavior and documentation.

Commit 3 fixes VDO-5897 by allowing dmeventd to parse VDO's status line again.
Commit 1 reverts the fix for VDO-5896, but commit 3 reverts the change that caused VDO-5896, so that issue should remain fixed.
Commit 3 and 4 could be squashed together, but I thought separating the revert from the new implementation would be clearer to read.

When going upstream:
Commit 1-2 are complete patch reversions, so for upstream purposes neither they nor the patches they revert will be included.
Commits 3-4 will be merged with commit 702e4fa22e82 ("dm vdo: add support for compression type and options").
Commit 5 will be merged with commit cb8051445b3f dm vdo ("doc: document new compression type parameter"").

### DIFF
--- a/doc/vdo.rst
+++ b/doc/vdo.rst
@@ -315,9 +315,8 @@ Status
 		'offline', 'online', 'opening', and 'unknown'.
 
 	compression state:
-		The current state of compression in the vdo volume; value
-		names the selected compression algorithm and options,
-		followed by 'on' or 'off' in parentheses.
+		The current state of compression in the vdo volume; values
+		may be 'offline' and 'online'.
 
 	used physical blocks:
 		The number of physical blocks in use by the vdo volume.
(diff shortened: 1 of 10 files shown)


raeburn (changes requested):
Code changes look good but in the commit message "--compressionType" shouldn't have the dashes, should it?

lorelei-sakai:
> Code changes look good but in the commit message "--compressionType" shouldn't have the dashes, should it?

Ah, I vascillated about that. Anyway, I removed the dashes and re-pushed.

raeburn (approved):

PR #289, opened 2025-04-25
[VDO-5892] Add feature flags to super block format

lorelei-sakai:
Update the super block format to allow for feature flags to indicate which future VDO features are required by any given volume. The super block format had an unused 4 bytes (previously the release version number) which have been repurposed to use as feature flags,so the encoded size of the super block does not change.

At the moment, there is only one flag, VDO_FEATURE_FLAG_LZ4 indicating the possible existence of lz4 compressed blocks in the VDO volume.

When going upstream, this will be presented as part of series, with the previous compressionType work coming before it, and the zstd support coming after it.


raeburn (dismissed):

src/c++/vdo/base/encodings.c
```diff
@@ -1422,16 +1433,35 @@ int vdo_decode_component_states(u8 *buffer, struct volume_geometry *geometry,
...
+	} else {
+		result = validate_version(VDO_VOLUME_VERSION_67_0,
+					  states->volume_version, "volume");
+		if (result != VDO_SUCCESS)
+			return result;
+
+		/*
+		 * Version 67 may have other data in the flags field. Preserve
+		 * it for backwards compatibility but don't validate it.
+		 */
```
sweettea, line 1461: I'm not seeing what other data it could have... can you elaborate this?

sweettea:
Mostly looks good to me. I think it would benefit from a pickle test also, a la the old Upgrade_t1.

src/c++/vdo/base/encodings.c
lorelei-sakai: The field that was called "unused" used to be "release_version" back when that was a thing. So a really old superblock might still have a release version here and we should be save it to write out again in case they also still have a really old driver somewhere.

lorelei-sakai:
> Mostly looks good to me. I think it would benefit from a pickle test also, a la the old Upgrade_t1.

Can you elaborate what this test would do? I'm not sure I understand. I did update the VDOVersion_t1 pickle.

sweettea:
> > Mostly looks good to me. I think it would benefit from a pickle test also, a la the old Upgrade_t1.
> 
> Can you elaborate what this test would do? I'm not sure I understand. I did update the VDOVersion_t1 pickle.

I envision something that verifies that we can read a v68 superblock, and the data within it, successfully, and save it out as a (v68?) superblock still.

lorelei-sakai:
> > > Mostly looks good to me. I think it would benefit from a pickle test also, a la the old Upgrade_t1.
> > 
> > 
> > Can you elaborate what this test would do? I'm not sure I understand. I did update the VDOVersion_t1 pickle.
> 
> I envision something that verifies that we can read a v68 superblock, and the data within it, successfully, and save it out as a (v68?) superblock still.

This exists. In SuperBlock_t1, the existing test (testCurrentSuperBlock) tests to 68-> 68 pathway, while the new test (testSuperBlock) test the 67-> 67 pathway. I expect there will later be a third test for the 67 -> 68 upgrade pathway, but I can't add that until there is actually anew flag to prompt that upgrade.
(The new VDOVersion_t1 pickle also has a v68 super block, commit 5)

sweettea:
I think a pickled 67 is also valuable insofar as it makes sure we don't have unexpectedly missed details. For instance, if the feature flags field for a loaded 67 wasn't properly initialized, and that feature flag had an impact on the readability of the data, a pickle could catch this when this manual initialization might not.

lorelei-sakai:
Pushed a new version renaming feature_flags to required_flags.

lorelei-sakai:
> I think a pickled 67 is also valuable insofar as it makes sure we don't have unexpectedly missed details.

I don't think a pickle can prove what you want it to prove. When we start a vdo device and write out the super block, some fields are not guaranteed to be the same so simply comparing to a pickled block isn't very helpful. The new test does compares the first ~100 bytes of the written out super block to check that those have not changed, at least. (That range covers the volume version and the flags field, among other things.) I think that should be sufficient.

src/c++/vdo/base/encodings.c
```diff
@@ -121,6 +121,15 @@ const struct version_number VDO_VOLUME_VERSION_67_0 = {
 	.minor_version = 0,
```
raeburn, line 121: If you're removing (and later re-adding but under INTERNAL) the header declaration for VDO_VOLUME_VERSION_67_0, shouldn't it be changed to static or STATIC at the definition?

lorelei-sakai:
Pushed a new version.
Changes:
- left old volume version visible.
- switched to volume version 67.1 instead of 68.0.
- removed unused version function.
- reordered commits so cleanups are first.

raeburn (approved):

PR #292, opened 2025-05-05
[VDO-5878] Remove VDOTest::SimpleMigration::testMigrateAndUpgrade

C2Redhat:
This test does not work.  Remove it.


sweettea (changes requested):
It would be helpful to add more context, I think -- can you put into the commit message some information about why we once thought this scenario was useful (and presumably the test passed back then?) and why we now do not / why the test fails now?

PR #296, opened 2025-05-09
[VDO-5708] Fix race in BlockMapTreeWrites_t1

lorelei-sakai:
Loosen the assertion in assertGeneration because it is not true as written. It is possible (though rare) for a flush for generation N to be enqueued while there are still pages from generation N-1 that have not been enqueued yet.

In particular, the problem in test set up can only happen because  the test is writing previously uninitialized block map pages, which are written twice, and the test only cares about the second write. In cases where the first write on the pages from generation 2 are slow, and the first write for the flusher from generation 3 is fast, the second flush write can be renqueued before the second writes from the previous generation have been enqueued. This race cannot happen with previously initialized block map pages

I believe the test set up could be rearranged to close this gap, but also the assertion as originally written is not a condition required (or guaranteed) by the system, so we shouldn't be trying to assert on it anyway.

### DIFF
--- a/src/c++/vdo/tests/BlockMapTreeWrites_t1.c
+++ b/src/c++/vdo/tests/BlockMapTreeWrites_t1.c
@@ -26,10 +26,17 @@
 #include "vdoTestBase.h"
 
 enum {
+  // Limit the entries per journal block to make it faster to fill them.
   ENTRIES_PER_BLOCK         = 16,
+  // How much data to write normally to fill a journal block.
   WRITES_PER_BLOCK          = ENTRIES_PER_BLOCK,
+  // The number of internal tree pages that get allocated for a new LBN.
+  // We exclude the root page because it never gets allocated or journalled.
   INTERIOR_HEIGHT           = VDO_BLOCK_MAP_TREE_HEIGHT - 1,
-  NEW_TREE_WRITES_PER_BLOCK = (ENTRIES_PER_BLOCK - INTERIOR_HEIGHT),
+  // How much data to write to fill a recovery block when the block map tree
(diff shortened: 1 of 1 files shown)


src/c++/vdo/tests/BlockMapTreeWrites_t1.c
```diff
@@ -231,7 +231,8 @@ static void testBlockMapTreeWrites(void)
   root_count_t trees = 4;
   setCompletionEnqueueHook(countWriteHook);
   for (root_count_t i = 0; i < trees; i++) {
-    writeData(i * VDO_BLOCK_MAP_ENTRIES_PER_PAGE, 0, (ENTRIES_PER_BLOCK - 4), VDO_SUCCESS);
+    writeData(i * VDO_BLOCK_MAP_ENTRIES_PER_PAGE, 0,
+              NEW_TREE_WRITES_PER_BLOCK, VDO_SUCCESS);
```
bjohnsto, line 235: I kinda wish NEW_TREE_WRITES_PER_BLOCK was identified more clearly in a comment or something in BlockMapTreeWrites_t1.c since it is an aggregate, but that's not your fault.

bjohnsto (dismissed):
I kin

src/c++/vdo/tests/BlockMapTreeWrites_t1.c
lorelei-sakai: Comments added. I had to spend a bit of time puzzling out what those numbers mean, so I might as well write it down to save the next person some time.

bjohnsto (approved):

bjohnsto (approved):

PR #297, opened 2025-05-12
dm vdo: Disable FEDORA40 testing and replace head with FEDORA41

C2Redhat:
Disable FEDORA40 from testing and replace any build reference from FEDORA39 to FEDORA41.


src/perl/Permabit/SupportedScenarios.yaml
```diff
@@ -9,11 +9,6 @@ X86_RHEL9_head:
   arch: X86_64
   moduleVersion: head
 
-X86_FEDORA40_head:
-  rsvpOSClass: FEDORA40
```
lorelei-sakai, line 13: We don't want to remove this one, we want to replace it with a FEDORA41 scenario.

lorelei-sakai (changes requested):
Generally looks good, except for  one thing in the supported scenarios.

src/perl/Permabit/SupportedScenarios.yaml
C2Redhat: I have updated it.  Thanks!

lorelei-sakai (changes requested):
Sorry I didn't catch this the first time. This PR is missing files.
1) Some migration tests actually use the X86_FEDORA40_head scenario, so they should be updated to use the new one (I see SimpleMigration.pm at least, there ight be more.)
2) NightlyBuildType/VDO.pm uses FEDORA40DEBUG for some suites. I think this should get replaced with FEDORA41DEBUG. (But it is not clear that we have FEDORA41DEBUG machines? So this issue might warrant a separate PR.)

Also the PR description mentions FEDORA39 instead of Fedora 40, that might be worth fixing.

C2Redhat:
> Quote reply

I have updated the PR with the changes that you requested.

lorelei-sakai (approved):
Looks good to me now.

PR #298, opened 2025-05-20
[Tech-203] Disable 'used only once' warning

C2Redhat:
ScriptFramework.pm causes an error during "make jenkins" in F41 with the 
following warning:
Name "main::log" used only once: possible typo at ./ScriptFramework.pm line 169.

Add ScriptFramework.pm to the NO_CHECK list.

### DIFF
--- a/src/perl/Permabit/ScriptFramework.pm
+++ b/src/perl/Permabit/ScriptFramework.pm
@@ -42,6 +42,9 @@ use Permabit::Assertions qw(assertMinArgs
 use Permabit::SystemUtils qw(assertCommand
                              assertSystem);
 
+# Turn off warning about 'Name "main::log" used only once:'
+no warnings qw(once);
+
 our %PROPERTIES
   = (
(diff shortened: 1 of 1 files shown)


rhawalsh:
> We can't do this, it will break scriptFramework. The use of that main::og value is not obvious and I think it's not even in this file, but it's necessary.
> 
> The right fix is to not do the checks that produce his warning. There is no NAK option so I'll marked this "Changes requested" but really we can't do this.

The way we have addressed it elsewhere is to disable the 'used once' warning.  Found in https://github.com/dm-vdo/common/blob/b02de5932995a21102a6953f9b36a32f57010c6a/perl/Permabit/Log4perlUtils.pm#L80

```
  # Turn off warning about APPENDER_BY_NAME only being used once
  no warnings qw(once);
```

lorelei-sakai (dismissed):
Looks good.

lorelei-sakai (approved):
Looks good.

PR #300, opened 2025-05-27
Fix missing ' " ' in $DEFAULT_OS_CLASSES at FEDORA42

C2Redhat:


lorelei-sakai (approved):
The title is a bit confusing to parse with the three quote characters in a row, but otherwise everything looks good.

PR #301, opened 2025-06-02
[VDO-5903] dm vdo test: Fix KernelUtils LowMemGenData01 test

C2Redhat:
Starting with RHEL9, we need to use command line grubby to adjust kernel memory limit. Since we have no RHEL8 in our lab, replace the RHEL9 test, isPlow(), with isRedHat() to adjust kernel memory limit.

### DIFF
--- a/src/perl/Permabit/Grub.pm
+++ b/src/perl/Permabit/Grub.pm
@@ -13,7 +13,7 @@ use List::Util qw(max);
 use Log::Log4perl;
 use Permabit::Assertions qw(assertDefined assertNumArgs);
 use Permabit::Constants;
-use Permabit::PlatformUtils qw(isPlow);
+use Permabit::PlatformUtils qw(isRedHat);
 use Permabit::SystemUtils qw(
   assertCommand
   runCommand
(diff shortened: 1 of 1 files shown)


src/perl/Permabit/Grub.pm
```diff
@@ -79,7 +79,8 @@ sub _installGrubConfig {
 ##
 sub modifyOption {
   my ($self, $kernelOption, $optionValue) = assertNumArgs(3, @_);
-  if (isPlow($self->{host})) {
+  if (isRedHat($self->{host})) {
+    $log->info("AAA Setting memory to ${optionValue}");
     assertCommand($self->{host},
```
lorelei-sakai, line 84: Did you mean to leave this log line in the commit?
If so,  what does the AAA mean?

lorelei-sakai (changes requested):
Generalizing this treatment to all RHEL platforms seems reasonable. I'm a little sad to lose the RHEL 8 support, but I agree that we wouldn't actually use it much, if ever.

src/perl/Permabit/Grub.pm
C2Redhat: I missed the log line, I will remove it.  I can change it to isCoughlan() for rhel10.  
But I think isRedHat is a better change.

lorelei-sakai (approved):
Looks good.

PR #302, opened 2025-06-03
dm vdo test: fix createThinPool in VolumeGroup.pm

bjohnsto:
For some reason, we forgot to add $dataType to the lvcreate string in createThinPool, which is what is needed to create proper vdo thin pools.


lorelei-sakai (commented):
The change itself looks perfectly plausible, and I would approve it. But I am concerned about the CI failures, so I'm deferring approval until we can get that sorted out.

Might be an issue with the vdo-vpt machines?

PR #303, opened 2025-06-05
[VDO-5908] Omit compression type testing when using the upstream module

lorelei-sakai:
Only run tests of the compressionType option on versions of vdo that implement that feature. In particular, don't run these tests again the code in linux-next upstream.


src/perl/vdotest/VDOTest/Dmsetup.pm
```diff
@@ -21,6 +21,7 @@ use Permabit::Assertions qw(
   assertRegexpMatches
   assertTrue);
 use Permabit::Constants;
+use Permabit::PlatformUtils qw(getDistroInfo);
```
raeburn, line 24: This function doesn't seem to be called, can this go away?

raeburn (changes requested):

raeburn (approved):

PR #305, opened 2025-06-11
dm vdo test: fix problem running Dory test in RAWHIDE

C2Redhat:
Fix #ifdef that VDO_USE_NEXT will be set in the latest RAWHIDE distro, 6.16.0.

### DIFF
--- a/src/c++/devices/common.c
+++ b/src/c++/devices/common.c
@@ -33,11 +33,11 @@ char *bufferToString(const char *buf, size_t length)
 }
 
 #undef VDO_USE_NEXT
-#if defined(RHEL_RELEASE_CODE)
-#if (RHEL_RELEASE_CODE > RHEL_RELEASE_VERSION(10, 1)) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
+#if defined(RHEL_RELEASE_CODE) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
+#if (RHEL_RELEASE_CODE > RHEL_RELEASE_VERSION(10, 1))
 #define VDO_USE_NEXT
 #endif
-#else /* RHEL_RELEASE_CODE */
+#else
 #if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0))
 #define VDO_USE_NEXT
 #endif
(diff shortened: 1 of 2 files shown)


src/c++/devices/common.c
```diff
@@ -36,12 +36,11 @@ char *bufferToString(const char *buf, size_t length)
 #if defined(RHEL_RELEASE_CODE)
 #if (RHEL_RELEASE_CODE > RHEL_RELEASE_VERSION(10, 1)) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
 #define VDO_USE_NEXT
-#endif
-#else /* RHEL_RELEASE_CODE */
-#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0))
+#elif (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0))
```
lorelei-sakai, line 39: This does not seem right. If RHEL_RELEASE_CODE is not defined at all, this code won't work.

It should probably look more like the version in dm-vdo-target.c, where we test for the RHEL_MINOR version on the same line where we check for RHEL_RELEASE_CODE.

lorelei-sakai, line 39: Please don't use #elif, I'd rather keep the format the same for all of them.

src/c++/devices/common.h
```diff
@@ -166,12 +166,11 @@ char *bufferToString(const char *buf, size_t length);
 #if defined(RHEL_RELEASE_CODE)
 #if (RHEL_RELEASE_CODE > RHEL_RELEASE_VERSION(10, 1)) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
 #define VDO_USE_NEXT
-#endif
-#else /* RHEL_RELEASE_CODE */
-#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0))
+#elif (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 15, 0))
 #define VDO_USE_NEXT
-#endif
+#endif
```
lorelei-sakai, line 171: Also pleae be careful not to add extra new lines (both file) or extra spaces at the end of lines (like here).

lorelei-sakai (changes requested):
I'm really confused why this change appears to fix things. Is the difference simply that RAWHIDE defines  RHEL_RELEASE_CODE?

src/c++/devices/common.h
C2Redhat: I have made the changes that you requested

PR #306, opened 2025-06-16
[VDO-5870] dm vdo test: fix dnf hang problem during tests.

C2Redhat:
In MAINLINENEXT test scenario, multiple dnf can run at the same time on different systems as user nightly. Both tests will try to lock the   ~nightly/.local/state/dnf5.log file and cause deadlock. Use XDG_STATE_HOME environment variable to change the log file location to avoid deadlock.


raeburn (changes requested):
"MAINTLINENEXT" in the description is a typo I think ... otherwise looks okay

raeburn (approved):

PR #308, opened 2025-06-23
[VDO-5919] dm vdo test: Fix build problem with RAWHIDE and FEDORANEXT distro

C2Redhat:
Fix the macro to set VDO_USE_NEXT in RAWHIDE and FEDROANEXT distro
for from_timer change in pull request 304:
https://github.com/dm-vdo/vdo-devel/pull/304

The change in pull request should also apply to src/c++/vdo/fake/linux/timer.h.

### DIFF
--- a/src/c++/vdo/fake/linux/timer.h
+++ b/src/c++/vdo/fake/linux/timer.h
@@ -40,7 +40,25 @@ int mod_timer(struct timer_list *timer, unsigned long expires);
 int del_timer_sync(struct timer_list *timer);
 int timer_delete_sync(struct timer_list *timer);
 
+#ifndef VDO_UPSTREAM
+#undef VDO_USE_NEXT
+#include <linux/version.h>
+#if defined(RHEL_RELEASE_CODE) && defined(RHEL_MINOR) && (RHEL_MINOR < 50)
+#if (RHEL_RELEASE_CODE > RHEL_RELEASE_VERSION(10, 1))
+#define VDO_USE_NEXT
+#endif
+#else /* RHEL_RELEASE_CODE */
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 16, 0))
+#define VDO_USE_NEXT
+#endif
+#endif /* RHEL_RELEASE_CODE */
+#endif /* VDO_UPSTREAM */
+#ifndef VDO_USE_NEXT
 #define from_timer(var, callback_timer, timer_fieldname) \
 	container_of(callback_timer, typeof(*var), timer_fieldname)
+#else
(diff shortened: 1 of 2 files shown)


lorelei-sakai (changes requested):
This looks wrong. You're literally just removing the code I added in PR#304. (If you mean to be reverting that change, you should probably say that you're reverting it.)

Did you test this change against MAINLINENEXT? We might need to change the conditions so that it works with FEDORANEXT also.

(Also, in general, VDO_USE_NEXT should get updated to use VDO_USE_ALTERNATE, not removed directly. There is a real change this is supporting, it's just not released everywhere yet. This is not surprising because the original change went it two days after the change landed in linux-next.)

C2Redhat:
> Member

I have fix the blank space and updated the title

src/c++/vdo/fake/linux/timer.h
```diff
@@ -40,7 +40,25 @@ int mod_timer(struct timer_list *timer, unsigned long expires);
...
+#else /* RHEL_RELEASE_CODE */
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(6, 16, 0))
+#define VDO_USE_NEXT
+#endif
+#endif /* RHEL_RELEASE_CODE */
+#endif /* VDO_UPSTREAM */
+#ifndef VDO_USE_NEXT
 #define from_timer(var, callback_timer, timer_fieldname) \
-	container_of(callback_timer, typeof(*var), timer_fieldname)
+        container_of(callback_timer, typeof(*var), timer_fieldname)
```
lorelei-sakai, line 58: Sorry I missed this yesterday: The indent here is eight spaces but it's supposed to be a tab.

src/c++/vdo/fake/linux/timer.h
```diff
@@ -40,7 +40,25 @@ int mod_timer(struct timer_list *timer, unsigned long expires);
...
+#else
+#define timer_container_of(var, callback_timer, timer_fieldname) \
+        container_of(callback_timer, typeof(*var), timer_fieldname)
```
lorelei-sakai, line 61: Also in this line: Initial eight spaces should be a tab.

lorelei-sakai (changes requested):
The blank lines are fixed, but the indents in the fake header are spaces instead of tabs.
Sorry I didn't catch that earlier.

(In normal code the check-style  target checks that, but got fake files we don't actually check the style. Nevertheless, it's inconsistent.)

lorelei-sakai (approved):
Looks great. Thanks for your patience.

C2Redhat:
> Looks great. Thanks for your patience.

Thanks for all the checking.  I am going to merge my change now.

PR #310, opened 2025-06-25
[VDO-5840] Removes vdo/statistics sysfs fields

bjohnsto:
Removes the sysfs usage of vdo statistics. This is an internal only change as the sysfs fields have already been removed by vdo's upstream code.


lorelei-sakai (approved):
Looks great! Yay.

PR #312, opened 2025-06-30
dm vdo test: update KERNEL_VERSION comparison from 15 to 16

C2Redhat:
Latest Fedora release version has been updated from 14 to 15, but the commonPrepareIoctl definition is not in the latest release yet. We need to update the comparison from 15 to 16 when defining VDO_USE_NEXT.

PR #313, opened 2025-06-30
[VDO-5870] dm vdo test: fix yum hang problem during tests

C2Redhat:
This is the same issue as
https://github.com/dm-vdo/vdo-devel/pull/306

instead of dnf, it is yum.  We use the same fix as 306 and use the XDG_STATE_HOME variable to avoid locking problems.


lorelei-sakai (approved):
This makes sense to me. Be sure to re-run the unit tets once PR #312 is submitted, I assume that's why they broke.

PR #322, opened 2025-07-29
Delete .github/workflows/notify.yml

bjohnsto:
This file is no longer needed. We have a global notifier now.


lorelei-sakai (approved):
Yep, looks great.

PR #323, opened 2025-07-31
[VDO-5752][VDOSTORY-292] Add super block intiialization into kernel code

bjohnsto:
This commit:
    - Adds vdo_initialize_component_states to encodings.c
    - Adds compute_forest_size to encodings.c
    - Adds RECOVERY_JOURNAL_STARTING_SEQUENCE_NUMBER const used in intiialization to constants.h
    - Removes local version of RECOVERY_JOURNAL_STARTING_SEQUENCE_NUMBER
    - Removes configreVDO from vdoConfig.c
    - Updates user code to call vdo_initialize_component_states
    - Removes computeForestSize from vdoConfig.c
    - Updates user code to call compute_forest_size

### DIFF
--- a/src/c++/vdo/user/vdoConfig.c
+++ b/src/c++/vdo/user/vdoConfig.c
@@ -21,10 +21,6 @@
 #include "userVDO.h"
 #include "vdoVolumeUtils.h"
 
-enum {
-  RECOVERY_JOURNAL_STARTING_SEQUENCE_NUMBER = 1,
-};
-
 /**********************************************************************/
 int initializeLayoutFromConfig(const struct vdo_config  *config,
                                physical_block_number_t   startingOffset,
@@ -78,63 +74,6 @@ computeForestSize(block_count_t logicalBlocks,
   // This can be a slight over-estimate since the tree will never have to
   // address these blocks, so it might be a tiny bit smaller.
   return (approximateNonLeaves + approximateLeaves);
-}
-
-/**
- * Configure a new VDO.
- *
- * @param vdo  The VDO to configure
- *
- * @return VDO_SUCCESS or an error
- **/
-static int __must_check configureVDO(UserVDO *vdo)
-{
-  struct vdo_config *config = &vdo->states.vdo.config;
-
-  // The layout starts 1 block past the beginning of the data region, as the
-  // data region contains the super block but the layout does not.
-  physical_block_number_t startingOffset = vdo_get_data_region_start(vdo->geometry) + 1;
-  int result = initializeLayoutFromConfig(config, startingOffset, &vdo->states.layout);
-  if (result != VDO_SUCCESS) {
-    return result;
-  }
-
-  vdo->states.recovery_journal = configureRecoveryJournal();
-
-  struct slab_config slabConfig;
-  result = vdo_configure_slab(config->slab_size,
-                              config->slab_journal_blocks,
-                              &slabConfig);
-  if (result != VDO_SUCCESS) {
-    return result;
-  }
-
-  const struct partition *partition =
-    getPartition(vdo, VDO_SLAB_DEPOT_PARTITION, "no allocator partition");
-  result = vdo_configure_slab_depot(partition, slabConfig, 0, &vdo->states.slab_depot);
-  if (result != VDO_SUCCESS) {
-    return result;
-  }
-
-  setDerivedSlabParameters(vdo);
-
-  if (config->logical_blocks == 0) {
-    block_count_t dataBlocks = slabConfig.data_blocks * vdo->slabCount;
-    config->logical_blocks
-      = dataBlocks - computeForestSize(dataBlocks,
-                                       DEFAULT_VDO_BLOCK_MAP_TREE_ROOT_COUNT);
-  }
-
-  partition = getPartition(vdo, VDO_BLOCK_MAP_PARTITION, "no block map partition");
-  vdo->states.block_map = (struct block_map_state_2_0) {
-    .flat_page_origin = VDO_BLOCK_MAP_FLAT_PAGE_ORIGIN,
-    .flat_page_count = 0,
-    .root_origin = partition->offset,
-    .root_count = DEFAULT_VDO_BLOCK_MAP_TREE_ROOT_COUNT,
-  };
-
-  vdo->states.vdo.state = VDO_NEW;
-  return VDO_SUCCESS;
 }
 
 /**********************************************************************/
@@ -252,13 +191,18 @@ static int configureAndWriteVDO(UserVDO                   *vdo,
     return result;
   }
 
-  vdo->states.vdo.config     = *config;
-  vdo->states.vdo.nonce      = nonce;
-  vdo->states.volume_version = VDO_VOLUME_VERSION_67_0;
   vdo->states.required_flags = 0;
-  result = configureVDO(vdo);
+  result = vdo_initialize_component_states(config, &vdo->geometry, nonce, &vdo->states);
   if (result != VDO_SUCCESS) {
     return result;
+  }
+
+  setDerivedSlabParameters(vdo);
+
+  if (vdo->states.vdo.config.logical_blocks == 0) {
+    block_count_t dataBlocks = vdo->states.slab_depot.slab_config.data_blocks * vdo->slabCount;
+    vdo->states.vdo.config.logical_blocks =
+	    dataBlocks - computeForestSize(dataBlocks, DEFAULT_VDO_BLOCK_MAP_TREE_ROOT_COUNT);
   }
 
   result = clearPartition(vdo, VDO_BLOCK_MAP_PARTITION);


src/c++/vdo/base/encodings.c
```diff
@@ -1571,6 +1571,135 @@ int vdo_decode_super_block(u8 *buffer)
...
+				       offset + 1,
+				       DEFAULT_VDO_BLOCK_MAP_TREE_ROOT_COUNT,
+				       states->vdo.config.recovery_journal_size,
+				       VDO_SLAB_SUMMARY_BLOCKS,
+				       &states->layout);
+	if (result != VDO_SUCCESS)
+		return result;
+
+	result = vdo_configure_slab(states->vdo.config.slab_size,
+				    states->vdo.config.slab_journal_blocks,
```
lorelei-sakai, line 1649: Again, you have a reference to the vdo_config, better to use that than dig in out of "states".

src/c++/vdo/base/encodings.c
```diff
@@ -1571,6 +1571,135 @@ int vdo_decode_super_block(u8 *buffer)
...
+
+	result = vdo_configure_slab_depot(partition, slab_config, 0,
+					  &states->slab_depot);
+	if (result != VDO_SUCCESS) {
+		vdo_uninitialize_layout(&states->layout);
+		return result;
+	}
+
+	/* Set derived parameters */
+	slab_size_shift = ilog2(states->vdo.config.slab_size);
```
lorelei-sakai, line 1671: slab_size_shift is only use once, in the next statement, maybe inline it?
Also third time's the charm maybe, you have a reference to vdo_config. Be sure to clean up all the place where you dig in out of the states structure.

src/c++/vdo/base/encodings.c
```diff
@@ -1571,6 +1571,135 @@ int vdo_decode_super_block(u8 *buffer)
...
+	slab_count = vdo_compute_slab_count(states->slab_depot.first_block,
+					    states->slab_depot.last_block,
+					    slab_size_shift);
+
+	if (states->vdo.config.logical_blocks == 0) {
```
lorelei-sakai, line 1676: logical_blocks can't actually be 0 in the kernel can it? We should try to restrict this this computation to the user side (incluing compute_forest_size(); that function doesn't seem to be used elsewhere). It doesn't look like this value is used further down this function, so the user side could just do this fixup after calling this function.

If I'm right that the kernel side won't use this though, it really shouldn't be here.

src/c++/vdo/base/encodings.c
```diff
@@ -1571,6 +1571,135 @@ int vdo_decode_super_block(u8 *buffer)
...
+{
+	unsigned int result;
+	unsigned int slab_size_shift;
+	slab_count_t slab_count;
+	struct slab_config slab_config;
+	struct partition *partition;
+
+	states->vdo.config = *vdo_config;
+	states->vdo.nonce = nonce;
+	states->volume_version = VDO_VOLUME_VERSION_67_0;
```
lorelei-sakai, line 1627: You dropped setting the required_flags fields. Since it's not something you really want in your commit, I'm not how we deal with it, though. Maybe put it in a separate patch that we can drop from the final product?
We could also elie it if we know that the "states" structure has already been initialized to all zeroes, which might be the case but I can't tell here.

src/c++/vdo/base/encodings.c
```diff
@@ -1571,6 +1571,135 @@ int vdo_decode_super_block(u8 *buffer)
...
+ * vdo_initialize_component_states() - Initialize the components so they can be written out.
+ * @vdo_config: The config used for component state initialization.
+ * @offset: The offset on disk where the data region begins.
+ * @nonce: The nonce to use to identify the vdo.
+ * @states: The component states to initialize
+ *
+ *
+ * Return: VDO_SUCCESS or an error code.
+ */
+int vdo_initialize_component_states(const struct vdo_config *vdo_config,
```
lorelei-sakai, line 1614: It's a bit weird here to add this function but not to add its caller. In this case I'm concerned because the way his is written right now, the function assumes the caller will do certain things correctly and I  can't verify that here.
(I think it's just 1) setting the required_flags field and 2) passing in the correct offset. For the latter, it might make more sense for the passed-in offset to do the one-block offset instead of doing it in this function, but it's hard to say without knowing what the caller looks like.)

src/c++/vdo/base/encodings.c
```diff
@@ -1571,6 +1571,135 @@ int vdo_decode_super_block(u8 *buffer)
...
+				    physical_block_number_t offset,
+				    nonce_t nonce,
+				    struct vdo_component_states *states)
+{
+	unsigned int result;
```
lorelei-sakai, line 1619: "result" is usually a regular int, not unsigned. (It is the return value of the function.)

src/c++/vdo/base/encodings.c
lorelei-sakai: Argh, it's even worse, you don't even need the slab_count except for this sizing calculation.
I was going to say that it's awkward to compute the slab_count separately when vdo_configure_slab_depot() has already calculated it. (It just doesn't expose the result.) But you don't need to do it at all.

src/c++/vdo/user/vdoConfig.c
```diff
@@ -21,10 +21,6 @@
 #include "userVDO.h"
 #include "vdoVolumeUtils.h"
 
-enum {
```
lorelei-sakai, line 24: It's a touch weird to put this in a separate commit instead of including with either the one before or the one after. But it's fine.

In the description, though, I don't think you need to say 'we have to merge this one if we merge the previous one'. I think we can take it as a given that a PR series needs to be merged as a whole and not picked apart.

src/c++/vdo/user/vdoConfig.c
```diff
@@ -248,11 +158,9 @@ static int configureAndWriteVDO(UserVDO                   *vdo,
     return result;
   }
 
-  vdo->states.vdo.config     = *config;
-  vdo->states.vdo.nonce      = nonce;
-  vdo->states.volume_version = VDO_VOLUME_VERSION_67_0;
-  vdo->states.required_flags = 0;
```
lorelei-sakai, line 258: The "required_flags" setting should remain since it din' move into the kernel. Or we should explicitly ensure the states structure is zeroed, perhaps.

src/c++/vdo/user/vdoConfig.c
```diff
@@ -248,11 +158,9 @@ static int configureAndWriteVDO(UserVDO                   *vdo,
...
-  result = configureVDO(vdo);
+  physical_block_number_t offset = vdo_get_data_region_start(vdo->geometry);
```
lorelei-sakai, line 161: Should we move the one-block increment to skip the super block here (along with its comment)? For now I think that's cleaner, but depending on what the kernel-side caller will look like, it might not be cleaner in the long run.

lorelei-sakai (changes requested):
Okay, generally looks good. I think the biggest thing is that the (logical_blocks == 0) case isn't meaningful in the kernel, so the handling for that should stay in userspace.
There's also some uncertainty about what the kernel-side calling code will look like, but I think it's probably fine if you an explain to me how that will look. Or we could patch it up later when you get to that part.

src/c++/vdo/base/encodings.c
```diff
@@ -1571,6 +1571,135 @@ int vdo_decode_super_block(u8 *buffer)
...
+	block_count_t approximate_non_leaves =
+		vdo_compute_new_forest_pages(root_count, NULL, logical_blocks, &new_sizes);
+
+	/*
+	 * Exclude the tree roots since those aren't allocated from slabs,
+	 * and also exclude the super-roots, which only exist in memory.
+	 */
+	approximate_non_leaves -=
+		root_count * (new_sizes.levels[VDO_BLOCK_MAP_TREE_HEIGHT - 2] +
+			     new_sizes.levels[VDO_BLOCK_MAP_TREE_HEIGHT - 1]);
```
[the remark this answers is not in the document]
bjohnsto: fixed

src/c++/vdo/base/encodings.c
```diff
@@ -1571,6 +1571,135 @@ int vdo_decode_super_block(u8 *buffer)
...
+
+	return (approximate_non_leaves + approximate_leaves);
+}
+
+/**
+ * vdo_initialize_component_states() - Initialize the components so they can be written out.
+ * @vdo_config: The config used for component state initialization.
+ * @offset: The offset on disk where the data region begins.
+ * @nonce: The nonce to use to identify the vdo.
+ * @states: The component states to initialize
```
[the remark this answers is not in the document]
bjohnsto: fixed

src/c++/vdo/base/encodings.c
```diff
@@ -1571,6 +1571,135 @@ int vdo_decode_super_block(u8 *buffer)
...
+	};
+
+	/*
+	 * The layout starts 1 block past the beginning of the data region, as the
+	 * data region contains the super block but the layout does not.
+	 */
+	result = vdo_initialize_layout(vdo_config->physical_blocks,
+				       offset + 1,
+				       DEFAULT_VDO_BLOCK_MAP_TREE_ROOT_COUNT,
+				       states->vdo.config.recovery_journal_size,
```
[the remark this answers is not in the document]
bjohnsto: done

src/c++/vdo/base/encodings.c
bjohnsto: The old code had it this way, but you're right. Should be cleaned up

src/c++/vdo/base/encodings.c
bjohnsto: I've put this entire block around an ifdef. I was going to move the whole thing back into the user mode code but it needs structures which are set up here.

src/c++/vdo/user/vdoConfig.c
bjohnsto: I can squash it if you like. I thought it would be useful to pull out since its a compile issue. So while you don't necessarily need to take the main user mode commit (if you decided to just keep vdoformat.c calling the same code and never touch the kernel code) this one you would actually need. At least that was my thought process

src/c++/vdo/base/encodings.c
bjohnsto: Actually I think I'm going to try to move it, and compute_forest_size back to user space only code

src/c++/vdo/user/vdoConfig.c
bjohnsto: squashed with other user mode commit

src/c++/vdo/user/vdoConfig.c
bjohnsto: I wonder if I can just move the offset inside the function.

src/c++/vdo/user/vdoConfig.c
```diff
@@ -220,6 +126,39 @@ static int __must_check clearPartition(UserVDO *vdo, enum partition_id id)
...
+                 newSizes.levels[VDO_BLOCK_MAP_TREE_HEIGHT - 1]);
+
+  block_count_t approximateLeaves =
+    vdo_compute_block_map_page_count(logicalBlocks - approximateNonLeaves);
+
+  // This can be a slight over-estimate since the tree will never have to
+  // address these blocks, so it might be a tiny bit smaller.
+  return (approximateNonLeaves + approximateLeaves);
+}
+
```
lorelei-sakai, line 161: Thanks for moving this back to userspace. Can you put it back where it used to be (above formatVDO) though, so it doesn't look like you're moving this function for no reason?

src/c++/vdo/base/encodings.c
```diff
@@ -1608,6 +1608,97 @@ static int vdo_compute_index_blocks(const struct index_config *config,
...
+	struct partition *partition;
+	block_count_t index_blocks = 0;
+	physical_block_number_t offset;
+
+	/* Calculate the data region offset from the index configuration */
+	result = vdo_compute_index_blocks(index_config, &index_blocks);
+	if (result != VDO_SUCCESS)
+		return result;
+
+	offset = 2 + index_blocks; /* 1 for super block, 1 for geometry */
```
lorelei-sakai, line 1636: Okay, this does feel like an abstraction violation to me. You're essentially depending on the index size to take up the entire region allocated to the index (which is currently always true, but...). I think the vdo formatting really should be anchored to the start of the data_region, like before, rather than to the beginning of the device. Which unfortunately means passing in either the geometry structure or the offset.

src/c++/vdo/user/vdoConfig.c
```diff
@@ -252,15 +191,20 @@ static int configureAndWriteVDO(UserVDO                   *vdo,
...
-  vdo->states.vdo.nonce      = nonce;
-  vdo->states.volume_version = VDO_VOLUME_VERSION_67_0;
   vdo->states.required_flags = 0;
-  result = configureVDO(vdo);
+  result = vdo_initialize_component_states(config, indexConfig, nonce, &vdo->states);
   if (result != VDO_SUCCESS) {
     return result;
   }
 
+  setDerivedSlabParameters(vdo);
```
lorelei-sakai, line 200: Hm, you're essentially inlining configureVDO() here. But if we don't have anything left which wants to call that as a separate function that probably fine. This function isn't getting too much bigger.

src/c++/vdo/user/vdoConfig.c
```diff
@@ -252,15 +191,20 @@ static int configureAndWriteVDO(UserVDO                   *vdo,
...
+
+  if (vdo->states.vdo.config.logical_blocks == 0) {
+    block_count_t dataBlocks = vdo->states.slab_depot.slab_config.data_blocks * vdo->slabCount;
+    vdo->states.vdo.config.logical_blocks =
+	    dataBlocks - computeForestSize(dataBlocks, DEFAULT_VDO_BLOCK_MAP_TREE_ROOT_COUNT);
```
lorelei-sakai, line 205: These lines are super long (90+ characters) but also they don't wrap well so it's fine. I just felt compelled to mention it.

lorelei-sakai (changes requested):
This is much cleaner, nice. The only remaining issue is the offset calculation. I think I did prefer the older version where we passed in the offset.

src/c++/vdo/base/encodings.c
bjohnsto: really weird. I know I made the changes. Maybe I didn't properly save. I'm still getting used to cursor.

src/c++/vdo/user/vdoConfig.c
bjohnsto: See my comments below. passing in volume_geometry into function makes the most sense to me

src/c++/vdo/base/encodings.c
bjohnsto: fixed. hopefully it works this time

lorelei-sakai (approved):
Yeah, this all looks good to me now.

One question is whether the kernel-side caller will have the geometry easily available to pass into this function, but it seems likely. If not we can fix that up when you add it.

PR #324, opened 2025-08-01
dm vdo test: Use dmidecode memory map when limiting host memory. [VDO-5920]

raeburn:
The existing code assumes the mem= value given to the kernel is the size of memory we're requesting. In fact, on x86, it's the highest physical address the kernel is allowed to use for general RAM. When there are multiple memory address regions with a gap between them, as seems more common now in our lab systems, we get the wrong result.

* Set memory limits on one host at a time. This makes the logic simpler, and doesn't make the assumption that all the machines are (roughly) the same, but will be slower if multiple machines are specified.

  I don't think we have any low-mem tests that use multiple machines. If this turns out to be incorrect we can go back to concurrent processing (and, more importantly, rebooting) but should still process each machine's info independently.

* Use a physical memory map retrieved with dmidecode. With a sorted list of address ranges, we can easily work out the address limit below which the desired amount of memory will be found.

  Most of this code was initially LLM-generated.

* Revert Jan 29 change allowing more iterations. That change simply papered over our failure to determine the mem= value correctly. We still need a loop, to deal with the varying difference between the requested value and the resulting MemTotal.


src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +357,176 @@ sub _isNear {
...
+# @param host           The host whose memory should be limited
+# @param minimumMemory  The minimum amount of memory to limit host to
 # @param extraMemory    The amount of memory above the minimum to allow,
 #                       if it is available
 #
 # @croaks on various error conditions
 ##
-sub setupKernelMemoryLimiting {
-  my ($hosts, $minimumMemory, $extraMemory) = assertNumArgs(3, @_);
+sub _setupKernelMemoryLimiting1Host {
```
lorelei-sakai, line 440: Gah, don't just mix digits into functions names, expecially when that digit could be confused for a letter.
How about  "OneHost" or "SingleHost" instead? (This name is awkwardly long in general, too. Is there a clearer, shorter name like "limitHostKernelMemory"?)

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +357,176 @@ sub _isNear {
...
+  my ($host, $minimumMemory, $extraMemory) = assertNumArgs(3, @_);
+  my $hosts = [ $host ];
```
lorelei-sakai, line 442: This $hosts list isn't actually used is it? Seems weird to define a one--element list in the function.

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +357,176 @@ sub _isNear {
...
-    confess("unable to adjust system memory parameter: @hostsInProgress");
+    confess("unable to adjust system memory parameter: $host");
   }
 
   # Set the appropriate hung task timeout now that we've rebooted.
-  map { setHungTaskTimeout($_) } @$hosts;
+  setHungTaskTimeout($host);
+}
+
+
```
lorelei-sakai, line 501: Extra blank line; pay attention to formatting.

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +357,176 @@ sub _isNear {
...
+sub setupKernelMemoryLimiting {
+  my ($hosts, $minimumMemory, $extraMemory) = assertNumArgs(3, @_);
+
+  foreach my $host (@$hosts) {
+    eval {
+      _setupKernelMemoryLimiting1Host($host, $minimumMemory, $extraMemory);
+    };
+    if (my $error = $EVAL_ERROR) {
+      # Log something here because the remove... cleanup may be verbose.
+      $log->error("_setupKernelMemoryLimiting1Host($host,$minimumMemory,$extraMemory) failed: $error");
```
lorelei-sakai, line 522: Could this error message maybe be English? The parameter values aren't useful unless you're actually looking at the code. If the message just needs to get you to this function, it could be simpler. Alternately, the message could actually describe what the function was trying to do.

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +357,176 @@ sub _isNear {
...
+# @param ranges     A memory map returned by _parseDmidecode
+# @param target_kb  The desired memory size
+#
+# @return a maximum address to pass to the kernel via mem=<N>k
+##
+sub _findAddressForMemoryKB {
+  my ($ranges, $target_kb) = assertNumArgs(2, @_);
+  my $accum_kb = 0;
+
+  for my $seg (@$ranges) {
```
lorelei-sakai, line 416: Please don't abbreviate variable names for no reason. use e.g. $range or $segment.

Further down, $accum_kb isn't even shorter than $totalKB.
Also please don't mix camelcase with snakecase. Reviewing LLM output includes making sure it aligns with our style.

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +357,176 @@ sub _isNear {
...
+  }
+
+  return \@merged;
+}
+
+#############################################################################
+# Given a memory map and desired memory size in kB, return maximum
+# address in kB. If too large a size is requested, an error is raised.
+#
+# Written initially by ChatGPT, then modified.
```
lorelei-sakai, line 405: What does this disclaimer do for us? What is it communicating, and to whom?

Either you have not reviewed the LLM output, in which case we shouldn't accept this, or you have reviewed it for correctness and to align it with our style, in which case, why is important for the reader to know what tools you used?

We probably need to think about this some more. I kind of agree that we should say _something_ but I'm really not sure what we can said that is meaningful or useful to read later.

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +357,176 @@ sub _isNear {
...
+  # Sort and merge ranges
+  @raw_ranges = sort { $a->[0] <=> $b->[0] } @raw_ranges;
+  my @merged;
+
+  for my $range (@raw_ranges) {
+    my ($start, $end) = @$range;
+    if (!@merged || $start > $merged[-1][1] + 1) {
+      push @merged, [$start, $end];
+    } else {
+      $merged[-1][1] = $end if $end > $merged[-1][1];
```
lorelei-sakai, line 394: I had to comment on this line specifically. $merged[-1][1] is a confusing thing to read, and the 'if'  in the middle of our statement is not our style. Specifically because our style is to write perl that's legible to someone used to reading C or Java and not the sort of garbage perl that you can find in random scripts of the internet. Just because perl lets you do it doesn't make it a good idea.

lorelei-sakai (commented):
Leaving aside the question of the right way to gather this information, I made a number of comments of the style. (I don't have enough expertise to weigh in on the best way to get the information we need.)

The style comments seem to boil down to, be more careful about adapting LLM output to match our existing coding style. Something to keep in mind whether you change the functionality or not.

src/perl/Permabit/KernelUtils.pm
raeburn: There are a couple of calls (_rebootWithKernelOption, removeKernelMemoryLimiting) that do want lists of names passed. I suppose I could construct the lists at the call sites if you'd prefer.

raeburn: removeKernelMemoryLimiting is called if we use the maximum number of iterations without setting $done to be true. (The calling function also calls it on all the hosts in case of a failure, so maybe this is redundant.)

src/perl/Permabit/KernelUtils.pm
raeburn: should be fixed now

src/perl/Permabit/KernelUtils.pm
raeburn: renamed the new variables; existing ones like currMem I haven't touched despite changing types

src/perl/Permabit/KernelUtils.pm
raeburn: reworded

src/perl/Permabit/KernelUtils.pm
raeburn: this code is gone now; the iomem entries appear to be ordered and I haven't noticed any where a merging pass would do anything, not that the merging was important anyway

src/perl/Permabit/KernelUtils.pm
raeburn: used OneHost

raeburn: just pushed an update to use _limitHostKernelMemory

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +358,173 @@ sub _isNear {
...
+      }
+
+      # Look for lines that end with ": System RAM"
+      if ($line =~ /^([0-9a-fA-F]+)-([0-9a-fA-F]+)\s*:\s*System RAM\s*$/) {
+        my $startHex = $1;
+        my $endHex = $2;
+
+        # Convert hex strings to Math::BigInt objects
+        my $start = Math::BigInt->new("0x$startHex");
+        my $end = Math::BigInt->new("0x$endHex");
```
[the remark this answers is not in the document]
raeburn: We do have 64-bit hosts only in our lab, so hex() would do the job, but it generates a warning for larger numbers. Rather than have to suppress the warning, it seems cleaner to just use BigInt and not have the 64-bit requirement. Updating the comment...

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +358,173 @@ sub _isNear {
...
+    my @lines = split /\n/, $iomemContent;
+
+    foreach my $line (@lines) {
+      # Skip indented lines (they start with whitespace)
+      if ($line =~ /^\s/) {
+        next;
+      }
+
+      # Look for lines that end with ": System RAM"
+      if ($line =~ /^([0-9a-fA-F]+)-([0-9a-fA-F]+)\s*:\s*System RAM\s*$/) {
```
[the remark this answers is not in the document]
raeburn: Tweaking the whitespace pattern to just match the actual content.

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +358,173 @@ sub _isNear {
...
+        # Add the range as an array ref to our results
+        push @ramRanges, [$start, $end];
+      }
+    }
+
+    return \@ramRanges;
+}
+
+#############################################################################
+# Given a memory map and desired memory size in kB, return maximum
```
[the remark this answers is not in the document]
raeburn: Doing all the math in kB is awkward to think about and describe (especially for addresses), so I'm doing it all in bytes until we calculate the mem= value. Variable names changed accordingly.

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +358,173 @@ sub _isNear {
...
+# address in kB. If too large a size is requested, an error is raised.
```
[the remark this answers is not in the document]
raeburn: fixed

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +358,173 @@ sub _isNear {
...
+sub _findAddressForMemoryKB {
+  my ($ranges, $targetKB) = assertNumArgs(2, @_);
+  my $accumulatedKB = 0;
+
+  for my $segment (@$ranges) {
+    my ($start, $end) = @$segment;
+    my $segmentKB = int(($end - $start + 1) / 1024);
+    if ($accumulatedKB + $segmentKB >= $targetKB) {
+      my $needed_kb = $targetKB - $accumulatedKB;
+      return int(($start + $needed_kb * 1024) / 1024);
```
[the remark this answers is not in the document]
raeburn: As noted above, now doing everything in bytes and only scaling once, instead of (1) scaling everywhere, or (2) managing everything including addresses scaled by 1024. Just seems easier to describe and reason about.

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +358,173 @@ sub _isNear {
...
+#
+# @return an array ref of memory address ranges
+##
+sub _parseSystemRAM {
+    my ($iomemContent) = assertNumArgs(1, @_);
+
+    my @ramRanges = ();
+
+    # Split content into lines
+    my @lines = split /\n/, $iomemContent;
```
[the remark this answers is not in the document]
raeburn: fixed, also a 'push' without parens

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +358,161 @@ sub _isNear {
...
+# @return a maximum address to pass to the kernel
+#
+# @croaks if the requested size is too large
+##
+sub _findAddressForMemory {
+  my ($ranges, $target) = assertNumArgs(2, @_);
+  my $accumulated = 0;
+
+  for my $segment (@$ranges) {
+    my ($start, $size) = @$segment;
```
lorelei-sakai, line 413: Usually we write @{$segment}  (use the braces) instead of using @$.

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +358,173 @@ sub _isNear {
...
+#
+# @return a maximum address to pass to the kernel via mem=<N>k
+##
+sub _findAddressForMemoryKB {
+  my ($ranges, $targetKB) = assertNumArgs(2, @_);
+  my $accumulatedKB = 0;
+
+  for my $segment (@$ranges) {
+    my ($start, $end) = @$segment;
+    my $segmentKB = int(($end - $start + 1) / 1024);
```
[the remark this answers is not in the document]
lorelei-sakai: Doing it all in bytes is much cleaner, yes.

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +358,161 @@ sub _isNear {
...
+      eval {
+        # Remove mem= options even if they weren't working.
+        removeKernelMemoryLimiting($hosts);
+        foreach my $cleanupHost (@$hosts) {
+          setHungTaskTimeout($cleanupHost);
+        }
+      };
+      confess("unable to adjust $host system memory parameter: $error");
+    }
+  };
```
lorelei-sakai, line 515: Silly nitpick, but I think this } closes the foreach and doesn't need the semicolon. (It's not a statement.)

lorelei-sakai (changes requested):
Sorry for the delay.

This PR is basically fine now , I just called out a few tiny nitpicks.

src/perl/Permabit/KernelUtils.pm
```diff
@@ -357,77 +358,161 @@ sub _isNear {
...
+##
+sub _findMemoryRanges {
+  my ($host) = assertNumArgs(1, @_);
+  my $iomemContent = assertCommand($host, "sudo cat /proc/iomem")->{stdout};
+  my @ramRanges = ();
+  my @lines = split(/\n/, $iomemContent);
+
+  foreach my $line (@lines) {
+    # Look for top-level range lines (not indented) labeled "System
+    # RAM"
```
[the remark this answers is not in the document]
raeburn: done

src/perl/Permabit/KernelUtils.pm
raeburn: fixed

lorelei-sakai (approved):
Look good to me.

Sorry about the @$ confusion, I somehow didn't parse that this file was already using the brace-free style.

PR #326, opened 2025-08-06
[VDO-5930] dm vdo: Fix npe in vdo-histograms.c

bjohnsto:
Two new functions in vdo-histograms looped through possible null histograms. Add checks for null in both cases and just skip those histograms.


lorelei-sakai (approved):
Sorry reflex, meant to approve.

PR #328, opened 2025-08-13
[VDO-5752] [VDOSTORY-292] Move is_zero_block to vio.c

bjohnsto:
Move the function that checks for zero block to a more generic location so it can be accessed by the kernel formatting code.

This commit:
- Moves is_zero_block from data-vio.c to vio.c
- Renames function to vdo_is_vio_data_zero
- Changes the parameter to a vio structure
- Renames call from is_zero_block() to vdo_is_vio_data_zero()


lorelei-sakai (changes requested):
I don't really remember how the formatting code wants to use this, which is really the whole justification for this move. If it's vio-specific, the function really should take a vio as its argument. (If it's not vio-specific, maybe there's another place it would fit better? But the only thing hat seems even vaguel sensible in encoding.[ch].)

Also, the commit descriptions sem sort of narrow to me. Are you wrapping at 50 characters? Other upstream commits seem to wrap no less than 70, so using 70-72 is probably better.

src/c++/vdo/base/vio.c
```diff
@@ -530,3 +530,22 @@ void vdo_count_completed_bios(struct bio *bio)
 	atomic64_inc(&vio->completion.vdo->stats.bios_completed);
 	count_all_bios_completed(vio, bio);
 }
+
+
+bool vdo_is_vio_data_zero(char *data)
```
raeburn, line 535: Upstream kernels have mem_is_zero in the kernel string support. It's available in F41 and RHEL9 kernels now. We should probably switch. I don't know if you want to do that as part of your change, but if not, moving the function may call upstream attention to the fact that we're not using the available kernel facility.

IMO it's probably good to just call mem_is_zero(,VDO_BLOCK_SIZE) and drop our custom function (and associated test case) rather than move it.

lorelei-sakai: If there's a kernel built-in, I agree that seems better. We'd need to add a userspace fake implementation, but that's probably not that hard.

bjohnsto:
> I don't really remember how the formatting code wants to use this, which is really the whole justification for this move. If it's vio-specific, the function really should take a vio as its argument. (If it's not vio-specific, maybe there's another place it would fit better? But the only thing hat seems even vaguel sensible in encoding.[ch].)
> 
> Also, the commit descriptions sem sort of narrow to me. Are you wrapping at 50 characters? Other upstream commits seem to wrap no less than 70, so using 70-72 is probably better.


i know its been a while but the way we decide to format is by checking to see if the geometry block is all zeroes. This was the function that would be called to check that

src/c++/vdo/base/vio.h
```diff
@@ -200,5 +200,6 @@ void free_vio_pool(struct vio_pool *pool);
 bool __must_check is_vio_pool_busy(struct vio_pool *pool);
 void acquire_vio_from_pool(struct vio_pool *pool, struct vdo_waiter *waiter);
 void return_vio_to_pool(struct pooled_vio *vio);
+bool vdo_is_vio_data_zero(char *data);
```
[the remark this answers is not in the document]
bjohnsto: I'm going to follow Ken's advice and try to use mem_is_zero. It will require making a string.h in userlinux/uds/linux for the user mode side of things, but thats fine. Stayed tuned for a new push

src/c++/vdo/base/data-vio.c
```diff
@@ -16,6 +16,7 @@
 #include <linux/lz4.h>
 #include <linux/minmax.h>
 #include <linux/sched.h>
+#include <linux/string.h>
```
lorelei-sakai, line 19: Alphabetically, "string" comes after "spinlock".

src/c++/uds/userLinux/uds/linux/string.h
```diff
@@ -0,0 +1,25 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+/*
+ * Copyright 2023 Red Hat
```
lorelei-sakai, line 3: Copyright notice feels a bit weird here? But if you include it, at least use the current year.

src/c++/uds/userLinux/uds/linux/string.h
```diff
@@ -0,0 +1,25 @@
...
+
+#ifndef STRING_H
+#define STRING_H
+
+#include <stdbool.h>
+#include <stddef.h>
+
+static inline bool mem_is_zero(const void *ptr, size_t size)
+{
+	if (!ptr)
```
lorelei-sakai, line 14: Where does this implementation come from? It doesn't seem to match the code I could find in the kernel.

I'm a little concerned that this implementation may be significantly slower than our existing implementation, which checks 8 bytes at a time, instead of checking bytes one at a a time like here. Can we just use (something like) the implementation we already had?

src/c++/vdo/tests/IsZeroBlock_t1.c
```diff
@@ -7,11 +7,11 @@
  */
 
 #include <linux/random.h>
-#include <string.h>
+#include <linux/string.h>
```
lorelei-sakai, line 10: We might not even need this test case at all; we don't generally test things that from the general kernel utilities. If we keep it, though, the updates look fine.

lorelei-sakai (dismissed):
This PR would actually be a good candidate to commit dirently on main rather than making it part of the vdoformat branch. It's not directly related to the formatting feature and its justifiable on its own merits.

raeburn (approved):

PR #333, opened 2025-09-15
dm vdo: Use kernel facilities for structs with flexible arrays

raeburn:
(Kept as several commits now to keep it clearer what the various updates are; the parts we might choose to send upstream should be squashed to a smaller set)

Use the kernel's size_mul, struct_size facilities for allocating structs with flexible arrays. Annotate the structs with counted_by attribute when appropriate. A few additional cases could be annotated with code changes at the allocation/initialization sites.

Remove type argument to VDO's allocation macros since it's redundant with the pointed-to type. (Requires fixing a few callers to pass the correct pointer types.)

### DIFF
--- a/src/c++/uds/kernelLinux/tests/RequestQueue_p1.c
+++ b/src/c++/uds/kernelLinux/tests/RequestQueue_p1.c
@@ -49,7 +49,7 @@ typedef struct queueableBatch {
   struct mutex            mutex;
   spinlock_t              spin;
   struct semaphore        semaphore;
-  Queueable               queueables[];
+  Queueable               queueables[] __counted_by(count);
 } QueueableBatch;
 
 /**********************************************************************/
--- a/src/c++/uds/userLinux/uds/linux/compiler_attributes.h
+++ b/src/c++/uds/userLinux/uds/linux/compiler_attributes.h
@@ -24,4 +24,10 @@
 #define fallthrough	do {} while (0)  /* fallthrough */
 #endif
 
+#if __has_attribute(__counted_by__)
+#define __counted_by(X) __attribute__((__counted_by__(X)))
+#else
+#define __counted_by(X)
+#endif
+
 #endif /* LINUX_COMPILER_ATTRIBUTES_H */
--- a/src/c++/uds/src/uds/index.h
+++ b/src/c++/uds/src/uds/index.h
@@ -58,7 +58,7 @@ struct uds_index {
 
 	index_callback_fn callback;
 	struct uds_request_queue *triage_queue;
-	struct uds_request_queue *zone_queues[];
+	struct uds_request_queue *zone_queues[] __counted_by(zone_count);
 };
 
 enum request_stage {
--- a/src/c++/uds/src/uds/open-chapter.h
+++ b/src/c++/uds/src/uds/open-chapter.h
@@ -40,7 +40,7 @@ struct open_chapter_zone {
 	/* The number of slots in the hash table */
 	unsigned int slot_count;
 	/* The hash table slots, referencing virtual record numbers */
-	struct open_chapter_zone_slot slots[];
+	struct open_chapter_zone_slot slots[] __counted_by(slot_count);
 };
 
 int __must_check uds_make_open_chapter(const struct index_geometry *geometry,
--- a/src/c++/vdo/base/block-map.h
+++ b/src/c++/vdo/base/block-map.h
@@ -279,7 +279,7 @@ struct block_map {
 	block_count_t next_entry_count;
 
 	zone_count_t zone_count;
-	struct block_map_zone zones[];
+	struct block_map_zone zones[] __counted_by(zone_count);
 };
 
 /**
--- a/src/c++/vdo/base/dedupe.c
+++ b/src/c++/vdo/base/dedupe.c
@@ -309,7 +309,7 @@ struct hash_zones {
 	/* The number of zones */
 	zone_count_t zone_count;
 	/* The hash zones themselves */
-	struct hash_zone zones[];
+	struct hash_zone zones[] __counted_by(zone_count);
 };
 
 /* These are in milliseconds. */
--- a/src/c++/vdo/base/logical-zone.h
+++ b/src/c++/vdo/base/logical-zone.h
@@ -60,7 +60,7 @@ struct logical_zones {
 	/* The number of zones */
 	zone_count_t zone_count;
 	/* The logical zones themselves */
-	struct logical_zone zones[];
+	struct logical_zone zones[] __counted_by(zone_count);
 };
 
 int __must_check vdo_make_logical_zones(struct vdo *vdo,
--- a/src/c++/vdo/base/physical-zone.c
+++ b/src/c++/vdo/base/physical-zone.c
@@ -202,7 +202,7 @@ struct pbn_lock_pool {
 	/** @idle_list: A list containing all idle PBN lock instances. */
 	struct list_head idle_list;
 	/** @locks: The memory for all the locks allocated by this pool. */
-	idle_pbn_lock locks[];
+	idle_pbn_lock locks[] __counted_by(capacity);
 };
 
 /**
--- a/src/c++/vdo/base/repair.c
+++ b/src/c++/vdo/base/repair.c
@@ -147,7 +147,7 @@ struct repair_completion {
 	 * The page completions used for playing the journal into the block map, and, during
 	 * read-only rebuild, for rebuilding the reference counts from the block map.
 	 */
-	struct vdo_page_completion page_completions[];
+	struct vdo_page_completion page_completions[] __counted_by(page_count);
 };
 
 /*
--- a/src/c++/vdo/base/slab-depot.h
+++ b/src/c++/vdo/base/slab-depot.h
@@ -509,7 +509,7 @@ struct slab_depot {
 	struct slab_summary_entry *summary_entries;
 
 	/* The block allocators for this depot */
-	struct block_allocator allocators[];
+	struct block_allocator allocators[] __counted_by(zone_count);
 };
 
 struct reference_updater;


src/c++/vdo/tests/mempool.c
```diff
@@ -31,11 +31,11 @@ mempool_t *mempool_create(int min_nr,
 			  void *pool_data)
 {
   mempool_t *pool;
-  VDO_ASSERT_SUCCESS(vdo_allocate_extended(mempool_t,
-                                           min_nr,
-                                           void *,
-                                           __func__,
-                                           &pool));
+  VDO_ASSERT_SUCCESS(vdo_allocate_flex(mempool_t,
+				       min_nr,
```
lorelei-sakai: There's no reason to replace spaces with tabs here, or in any of our test files. For me, it shows up as messed-up indentation.

Side note a) Is this an LLM suggestion thing? Chung's been doing this a lot lately, too.
Side note b) I have no idea why github displays a tab as eight spaces in vdo/base files, but as four spaces in vdo/test files. It's driving me nuts.

src/c++/vdo/base/recovery-journal.c
```diff
@@ -709,10 +709,9 @@ int vdo_decode_recovery_journal(struct recovery_journal_state_7_0 state, nonce_t
 	struct recovery_journal *journal;
 	int result;
 
-	result = vdo_allocate_extended(struct recovery_journal,
-				       RECOVERY_JOURNAL_RESERVED_BLOCKS,
-				       struct recovery_journal_block, __func__,
-				       &journal);
+	result = vdo_allocate_flex(struct recovery_journal,
```
lorelei-sakai: Minor nit: could we break this line 2/3, or is that too long? If not, 1/1/3 probably looks more even than this 1/3/1 distribution.

src/c++/uds/kernelLinux/tests/RequestQueue_p1.c
```diff
@@ -411,8 +411,8 @@ static void funnelEventConsume(QueueableBatch *qb)
 static QueueableBatch *allocateBatch(long stream, long count)
 {
   QueueableBatch *qb;
-  UDS_ASSERT_SUCCESS(vdo_allocate_extended(QueueableBatch, count, Queueable,
-                                           __func__, &qb));
+  UDS_ASSERT_SUCCESS(vdo_allocate_flex(QueueableBatch, count, q,
```
lorelei-sakai: Could we rename the flexible array to something longer that "q"? Maybe "queuables"? I know it's a test, but in the macro I'd really like a name that somewhat descriptive.

(Somewhat amazingly to me, this is the only one that obviously needs to changes; most structures already have reasonable names.)

src/c++/uds/userLinux/uds/memoryAlloc.c
```diff
@@ -80,8 +80,9 @@ int vdo_allocate_memory(size_t size, size_t align, const char *what, void *ptr)
 void *vdo_allocate_memory_nowait(size_t size, const char *what)
 {
 	void *p = NULL;
+	int _unused __attribute__((__unused__));
```
lorelei-sakai, line 83: I feel like "int result" is a more standard name here. It's only two lines apart so a reader would still see that it's unused.

src/c++/uds/src/uds/memory-alloc.h
```diff
@@ -81,15 +81,11 @@ static inline int __must_check __vdo_do_allocation(size_t count, size_t size,
...
-		int _result;						\
 		TYPE **_ptr = (PTR);					\
-		_result = __vdo_do_allocation(COUNT,			\
-					      sizeof(((TYPE*)0)->FIELD[0]),	\
-					      sizeof(TYPE),		\
-					      __alignof__(TYPE),	\
-					      WHAT,			\
-					      _ptr);			\
-		_result;						\
+		vdo_allocate_memory(struct_size(*_ptr, FIELD, (COUNT)),	\
```
lorelei-sakai: ... Do we want to bypass __vdo_allocate_memory? You call that out in the commit message, but I don't know why we should consider that good a priori. I think what I mean is the commit message should have a clearer justification for why you want to do this.

(The change itself is a lot cleaner, and obviates some of my earlier comments about the macro formatting.)

src/c++/uds/src/uds/memory-alloc.h
```diff
@@ -63,8 +34,9 @@ static inline int __must_check __vdo_do_allocation(size_t count, size_t size,
  *
  * Return: VDO_SUCCESS or an error code
  */
-#define vdo_allocate(COUNT, TYPE, WHAT, PTR) \
-	__vdo_do_allocation(COUNT, sizeof(TYPE), 0, __alignof__(TYPE), WHAT, PTR)
+#define vdo_allocate(COUNT, TYPE, WHAT, PTR)			\
```
lorelei-sakai: I suppose this is the justification for the previous commit. If  they're in a series, maybe it doesn't need to be justified all that explicitly, then.

src/c++/uds/userLinux/uds/linux/compiler_attributes.h
```diff
@@ -24,4 +24,10 @@
 #define fallthrough	do {} while (0)  /* fallthrough */
 #endif
 
+#if __has_attribute(__counted_by__)
```
lorelei-sakai, line 27: I'm not at all clear that this __counted_by() annotation is good for anything. Why do this? What's the benefit?

src/c++/uds/src/uds/memory-alloc.h
```diff
@@ -42,23 +42,18 @@ int __must_check vdo_allocate_memory(size_t size, size_t align, const char *what
...
  */
-#define vdo_allocate_flex(TYPE, COUNT, FIELD, WHAT, PTR)		\
-	__extension__({							\
-		TYPE **_ptr = (PTR);					\
-		vdo_allocate_memory(struct_size(*_ptr, FIELD, (COUNT)),	\
-				    __alignof__(TYPE),			\
-				    WHAT,				\
-				    _ptr);				\
-	})
+#define vdo_allocate_flex(COUNT, FIELD, WHAT, PTR)			\
```
lorelei-sakai, line 51: I don't really like the loss of specificity from not explicitly naming the type, but I think that's just reflex. It doesn't look like there's any real benefit to keeping it. So this is fine.

src/c++/uds/src/uds/memory-alloc.h
```diff
@@ -28,15 +28,14 @@ int __must_check vdo_allocate_memory(size_t size, size_t align, const char *what
...
- * @TYPE: The type of objects to allocate. This type determines the alignment of the allocation.
  * @WHAT: What is being allocated (for error logging)
  * @PTR: A pointer to hold the allocated memory
  *
  * Return: VDO_SUCCESS or an error code
  */
-#define vdo_allocate(COUNT, TYPE, WHAT, PTR)			\
-	vdo_allocate_memory(size_mul((COUNT), sizeof(TYPE)),	\
-			    __alignof__(TYPE), WHAT, PTR)
+#define vdo_allocate(COUNT, WHAT, PTR)					\
```
lorelei-sakai, line 36: Same comment as the previous change, this feels like giving up a measure of control, but it's probably not anything we really need.

This one does touch a really large number of files, it might be worth getting an opinion for Mikulas it this is fine or if we should break it up. Breaking it into stages would be a lot more annoying, but better than getting pushback upstream. (Realistically, we're not going to get pushback since this is all within dm-vdo. But I'd like to get some confirmation anyway.)

Also do remember to do the reformatting mentioned in the current commit message.

src/packaging/src-dist/MANIFEST.yaml
```diff
@@ -197,6 +197,7 @@ tarballs:
             - limits.h
             - log2.h
             - mutex.h
+            - overflow.h
```
lorelei-sakai, line 200: Good catch. But can you squash this into the change where you added overflow.h? Or at least to be stacked directly after it?

lorelei-sakai (changes requested):
Now that I've had a few days to get over my visceral reaction, I've got to admit this looks like a good bit of cleanup. I'm not sold on the __counted_by commit, but all the rest of it seems like a good idea.

src/c++/uds/src/uds/memory-alloc.h
```diff
@@ -13,6 +13,7 @@
 #include <stdlib.h>
 #include <string.h>
 #endif
+#include <linux/overflow.h>
```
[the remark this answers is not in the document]
raeburn: I'm assuming that, after we decide how much of the change functionality we want, I'll be squashing them down to a smaller set of changes, and cleaning up the messages at that point.

src/c++/vdo/tests/mempool.c
raeburn: That's my emacs session defaulting to using tabs for indentation, when I went to fix the indentation after changing the length of the macro name. I really wish we could just switch to kernel style for everything instead of this mix of styles we have here, but that's not for this PR to tackle.

src/c++/uds/src/uds/memory-alloc.h
raeburn: I could merge the changes together if that's clearer or less likely to need justification.
Maybe combine it with the __must_check change?

src/c++/uds/userLinux/uds/linux/compiler_attributes.h
raeburn: "GCC may use this information to improve detection of object size information for such structures and provide better results in compile-time diagnostics and runtime features like the array bound sanitizer and the __builtin_dynamic_object_size. "  https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html

(The "sanitizer" extensions are used in KASAN and UBSAN and such.)

The clang compiler supports this also.

The kernel maintainers seem to think it useful enough to use in a lot of variable-size kernel data structures.

Functionally, it shouldn't change anything for normal operation; just refine the ability to check our code at compile time or run time.

src/c++/vdo/tests/mempool.c
lorelei-sakai: Doing it all in one go seems like it might be noisy (but I haven't looked at it) but certainly I think it's reasonable to do that conversion gradually, a file or a directory at a time. (But in a separate commit, I think.) Especially for C code that mimics the kernel, it makes sense.

I believe the original argument was we weren't going to do it because we didn't need to do it for upstream, but hat was an argument for deferring until later (like now?) rather than never doing it.

src/c++/uds/kernelLinux/uds/memory-alloc.c
```diff
@@ -532,16 +532,18 @@ int vdo_reallocate_memory(void *ptr, size_t old_size, size_t size, const char *w
...
 		vdo_free(ptr);
 		*(void **) new_ptr = NULL;
 		return VDO_SUCCESS;
 	}
 
-	result = vdo_allocate(size, char, what, new_ptr);
+	result = vdo_allocate(size, what, &temp_ptr);
 	if (result != VDO_SUCCESS)
 		return result;
+	*(void **)new_ptr = temp_ptr;
```
lorelei-sakai: I kind of want a space here between the cast and "new_ptr". Specifically, I think the formatting should match that on line 539, just above this, and line 552 below.

src/c++/uds/src/tests/RadixSort_t1.c
```diff
@@ -45,8 +45,8 @@ static const u8 **sortAndVerify(const u8 *keys[], unsigned int count,
 {
   // Make a copy of the keys we're going to sort.
   u8 *bytes;
-  UDS_ASSERT_SUCCESS(vdo_allocate(count * sizeof(keys[0]), u8, "keys",
-                                  &bytes));
+  UDS_ASSERT_SUCCESS(vdo_allocate(count * sizeof(keys[0]), "keys",
+				  &bytes));
```
lorelei-sakai: Another spaces-to-tabs line (and honestly this probably doesn't need to line wrap, anyway).

src/c++/uds/src/uds/index.c
```diff
@@ -1196,7 +1196,7 @@ int uds_make_index(struct uds_configuration *config, enum uds_open_index_type op
 		return result;
 	}
 
-	result = vdo_allocate(index->zone_count, struct index_zone *, "zones",
+	result = vdo_allocate(index->zone_count, "zones",
```
lorelei-sakai: You could unwrap this line, it would still be shorter that the original first line.

src/c++/uds/userLinux/tests/ConvertToLVM_x1.c
```diff
@@ -200,12 +200,12 @@ static void slide_file(off_t bytes)
   enum {
     BUFFER_SIZE  = 4096,
   };
-  void *buf;
+  char *buf;
```
lorelei-sakai: Same here, changed old type u8 to char.

src/c++/uds/userLinux/uds/memoryAlloc.c
```diff
@@ -79,9 +79,10 @@ int vdo_allocate_memory(size_t size, size_t align, const char *what, void *ptr)
  */
 void *vdo_allocate_memory_nowait(size_t size, const char *what)
 {
-	void *p = NULL;
+	char *p = NULL;
+	int result __attribute__((__unused__));
 
-	vdo_allocate(size, char *, what, &p);
+	result = vdo_allocate(size, what, &p);
```
lorelei-sakai: Is the new spelling enforcing checking the return code of the allocation?

src/c++/vdo/base/slab-depot.c
```diff
@@ -4232,9 +4225,7 @@ static int allocate_components(struct slab_depot *depot,
 
 	depot->summary_origin = summary_partition->offset;
 	depot->hint_shift = vdo_get_slab_summary_hint_shift(depot->slab_size_shift);
-	result = vdo_allocate(MAXIMUM_VDO_SLAB_SUMMARY_ENTRIES,
-			      struct slab_summary_entry, __func__,
-			      &depot->summary_entries);
+	result = vdo_allocate(MAXIMUM_VDO_SLAB_SUMMARY_ENTRIES, __func__, &depot->summary_entries);
```
lorelei-sakai: This looks like exactly 99 characters to me... maybe wrapping it is actually better?

src/c++/vdo/tests/JournalThresholds_t1.c
```diff
@@ -113,10 +113,9 @@ static void initialize(void)
   };
   initializeRecoveryModeTest(&parameters);
 
-  VDO_ASSERT_SUCCESS(vdo_allocate(vdo->depot->slab_count, logical_block_number_t,
-                                  __func__, &slabLBNs));
-  VDO_ASSERT_SUCCESS(vdo_allocate(vdo->depot->slab_count, logical_block_number_t,
-                                  __func__, &slabLBNs2));
+  VDO_ASSERT_SUCCESS(vdo_allocate(vdo->depot->slab_count, __func__, &slabLBNs));
+  VDO_ASSERT_SUCCESS(vdo_allocate(vdo->depot->slab_count, __func__,
+                                  &slabLBNs2));
```
lorelei-sakai: One character pushes it over the wrapping limit? I mean, it's fine, it just seemed odd.

src/c++/uds/src/uds/memory-alloc.h
```diff
@@ -81,28 +82,18 @@ static inline int __vdo_do_allocation(size_t count, size_t size, size_t extra,
...
- *         allocated memory.
  * @COUNT: The number of objects to allocate
- * @TYPE2: The type of array objects to allocate
+ * @FIELD: The flexible array field at the end of the TYPE structure
  * @WHAT: What is being allocated (for error logging)
  * @PTR: A pointer to hold the allocated memory
  *
  * Return: VDO_SUCCESS or an error code
  */
-#define vdo_allocate_extended(TYPE1, COUNT, TYPE2, WHAT, PTR)		\
```
lorelei-sakai: Putting this is a separate converstaion so you don't miss it:
Sign your commits! Makes upstreaming so much easier if you do it at this stage.
Commit messages otherwise look good.

src/c++/uds/userLinux/uds/memoryAlloc.c
raeburn: Yes, with these changes, vdo_allocate is essentially __must_check now.

src/c++/uds/src/tests/RadixSort_t1.c
raeburn: True, unwrapped it

src/c++/vdo/tests/workQueue.c
```diff
@@ -161,9 +161,8 @@ int vdo_make_work_queue(const char *thread_name_prefix,
 
   STATIC_ASSERT((int) NO_HOOK_FLAG > (int) VDO_WORK_Q_DEFAULT_PRIORITY);
   STATIC_ASSERT((int) WORK_FLAG    > (int) VDO_WORK_Q_DEFAULT_PRIORITY);
-  VDO_ASSERT_SUCCESS(vdo_allocate_extended(struct vdo_work_queue,
-                                           priorityLevels,
-                                           struct funnel_queue *,
+  VDO_ASSERT_SUCCESS(vdo_allocate_extended(priorityLevels,
+                                           queues,
```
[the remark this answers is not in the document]
raeburn: Changed these. Though, the style is used in other, unrelated calls in the area as well, even when multiple arguments would fit on a line.

src/c++/uds/userLinux/tests/ConvertToLVM_n1.c
```diff
@@ -90,12 +90,12 @@ static void slide_file(off_t bytes)
   enum {
     BUFFER_SIZE  = 4096,
   };
-  void *buf;
+  char *buf;
```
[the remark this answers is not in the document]
raeburn: Yeah, I didn't mean to change it. Putting it back...

src/c++/uds/src/uds/index.c
raeburn: fixed

src/c++/vdo/base/slab-depot.c
raeburn: put the output arg back on its own line

PR #336, opened 2025-09-19
[VDO-5787] Finalize lvmdevice usage for our stack

bjohnsto:
Updates the code that adds to and removes from the lvm devices file.

In order to properly handle the relationship between our devices and their entries in the lvm devices file, we need to do the following:

1. Add the device ONLY AFTER the device is successfully added to the system.
2. Remove the device ONLY AFTER the device is successfully removed from the system.
3. Find a common location for the removal, given that devices can be removed by a call to dmsetup remove, lvchange an, lvremove, mdadm, etc. Also, these calls are not all done in the same place in the lifecycle of a VDO BlockDevice.
4. Handle situation where some calls to dmsetup remove are actually in code outside of the vdo-devel project.
5. Make sure the device name that is removed is the same one that is used when added to the LVM devices file, in case of reboots or things that might rename devices.


src/perl/Permabit/BlockDevice.pm
```diff
@@ -325,25 +325,117 @@ sub start {
   }
 }
 
+########################################################################
+# Get the device path to use for LVM devices file operations.
+# This can be overridden by subclasses that need special device path logic.
+#
+# @return the device path to use for LVM devices file operations
+##
+sub getDevicesFilePath {
```
lorelei-sakai, line 334: The name of this function is confusing to me. It sounds the path the lvmdevices file.
But it's actually the path to the device that will be used in the lvmdevices file... yes?

Anyway, I the function logic (including the override in ISCSI.pm) looks sensible, I think this just needs a better name.

src/perl/Permabit/BlockDevice.pm
```diff
@@ -325,25 +325,117 @@ sub start {
...
+  my ($self) = assertNumArgs(1, @_);
+  return $self->getDevicePath();
+}
+
+########################################################################
+# Check if lvmdevices is unavailable or the devices file is disabled.
+#
+# @return true if lvmdevices cannot be used, false if it can be used
+##
+sub _isLvmDevicesUnavailable {
```
lorelei-sakai, line 344: This name is also confusing.
Point 1, after reading the code I realized this should read "Is lvmdevices unavailable?" But in that case it should be capitalized "_isLvmdevicesUnavailable".
Point 2, it's usually clearer not to put a negative in the function name, so this should be inverted to isLvmdevicesAvailable().
Point 3, there might be a clearer way to describe what you're actually checking, for instance "isDevicesFileAvailable" or "canUseDevicesFile".

There a certain confusion between the device itself, the Perl class describing the device, and the devices file, so I think it's worth being extra careful to make sure you're describing things precisely.

src/perl/Permabit/BlockDevice.pm
```diff
@@ -325,25 +325,117 @@ sub start {
...
+  }
+
+  return 0;
+}
+
 ########################################################################
 # Perform action to add device to lvm devices file.
 ##
-sub addToDevicesFile {
+sub addToDevicesFileIfExists {
```
lorelei-sakai, line 360: "IfExists" feels redundant on this function name (and also the remove function). I'd remove it, but I'm fine if you think it's helpful, too.

src/perl/Permabit/BlockDevice.pm
```diff
@@ -325,25 +325,117 @@ sub start {
...
   my ($self) = assertNumArgs(1, @_);
-  my $device = $self->getDevicePath();
-  my $addDevCmd = "sudo lvmdevices -y --adddev $device";
-  $self->runOnHost($addDevCmd);
-  $self->addDeactivationStep(sub { $self->removeFromDevicesFile($device); })
+  my $device = $self->getDevicesFilePath();
+
+  # If no device path is available, nothing to do
```
lorelei-sakai, line 364: This comment, and the one on 369, feel trivial and thus fairly useless (were they LLM suggestions?) I'd prefer you cut them.

src/perl/Permabit/BlockDevice.pm
```diff
@@ -325,25 +325,117 @@ sub start {
...
+      if ($resolvedDevName eq $device) {
+        $storedDeviceName = $devName;
+        $log->info("LVM stored device as: $storedDeviceName (resolved: $resolvedDevName, original: $device)");
+        last;
+      }
+    }
+  }
+
+  # Store the path that LVM actually used, or fall back to our original path
+  $self->{_devicesFilePath} = $storedDeviceName || $device;
```
lorelei-sakai, line 394: If you're storing a value in a local field, it's worth adding it to the %BLOCKDEVICE_PROPERTIES hash just so it doesn't catch a later reader by surprise. (It looks like you can initialize it to undef and the code will do the same thing as now.)

src/perl/Permabit/BlockDevice.pm
```diff
@@ -325,25 +325,117 @@ sub start {
...
 # Perform action to add device to lvm devices file.
 ##
-sub addToDevicesFile {
+sub addToDevicesFileIfExists {
   my ($self) = assertNumArgs(1, @_);
-  my $device = $self->getDevicePath();
-  my $addDevCmd = "sudo lvmdevices -y --adddev $device";
-  $self->runOnHost($addDevCmd);
-  $self->addDeactivationStep(sub { $self->removeFromDevicesFile($device); })
+  my $device = $self->getDevicesFilePath();
```
lorelei-sakai, line 362: Similar to the comment on the function name, is there a clearer name for this variable? The meaning of "$device" is kind of overloaded in this file.

src/perl/Permabit/BlockDevice.pm
```diff
@@ -325,25 +325,117 @@ sub start {
...
+  # Check if lvmdevices is available and devices file is enabled
+  if ($self->_isLvmDevicesUnavailable()) {
+    return;
+  }
+
+  # Check if the block device still exists on the system using current device path
+  if ($machine->sendCommand("test -e " . $self->getDevicePath()) == 0) {
+    # Non-Raw devices created themselves, so if they still exist, don't remove from
+    # the devices file
+    if (!$self->isa('Permabit::BlockDevice::Raw')) {
```
lorelei-sakai, line 418: I think you probably want to use Permabit::BlockDevice::Bottom instead of ...Raw here. There are a few Bottom devices besides Raw, though we don't use them much. But I think the logic should be the same for the others.

Comments would need to be updated accordingly.

src/perl/Permabit/BlockDevice.pm
```diff
@@ -325,25 +325,117 @@ sub start {
...
+    # remove from the devices file
+    if (scalar($self->getChildren()) > 0) {
+      return;
+    }
+  }
+
+  # Normally the device would get removed from the devices file when the device is
+  # destroyed, but if it's not, use sed to directly remove the entry from the devices
+  # file as a safety fallback
+  my $escaped_device = $self->{_devicesFilePath};
```
lorelei-sakai, line 433: Maybe $escapedDeviceName or $escapedDevicePath for clarity? At any rate, could you use camel-case here like elsewhere?

src/perl/Permabit/BlockDevice.pm
```diff
@@ -550,6 +642,10 @@ sub destroy {
 sub teardown {
   my ($self) = assertNumArgs(1, @_);
   assertTrue($self->{_destroying});
+
+  # Clean up devices file entry if one was created
+  $self->removeFromDevicesFileIfExists();
```
lorelei-sakai, line 647: I'm not sure this is wrong, but I don't understand how this matches the description in the commit message. The commit message says we have to remove the device from the devices file after the device is removed.

... Is the logic here that this gets called after destroy() so the device is already gone by the time we call teardown()? I could use some explanation about how this does what you say it does. (To be clear, I mean let's have a conversation in Slack. If this code is right, there's no need to belabor it in code comments.)

lorelei-sakai (changes requested):
The code logic looks reasonable, aside from small note about non-Raw Bottom devices. most of the comments are about improving the readability, mostly through naming.

src/perl/Permabit/BlockDevice.pm
bjohnsto: Raw is the one device that doesn't create its own associated system device. It uses whats already on the machine. RAID and Loop and other Bottom devices create their own device using mdadm or losetup. Therefore we expect to remove those devices at this point.

src/perl/vdotest/VDOTest/LVMDevicesTest.pm
```diff
@@ -0,0 +1,130 @@
...
+#############################################################################
+# Test that creating and destroying a raw device properly maintains the
+# LVM devices file state and that the system device still exists after destruction
+##
+sub testRawDevice {
+  my ($self) = assertNumArgs(1, @_);
+  my $machine = $self->getUserMachine();
+
+  if (!isLvmdevicesAvailable($machine)) {
+    return;
```
lorelei-sakai, line 44: Looking at this now, I think there should be a log (info) message in this branch, something to the effect of "skipping lvmdevices test becuase lvmdevices file is not available". Otherwise we might easily end up in the situation where we think this test is working, but it's not actually testing anything.

This comment applies to both test cases.

src/perl/Permabit/LVMUtils.pm
```diff
@@ -0,0 +1,141 @@
...
+      return 1;
+    }
+  }
+  return 0;
+}
+
+########################################################################
+# Parse LVM devices output and return an array of invalid devices.
+#
+# This function runs 'sudo lvmdevices', parses the output for DEVNAME entries,
```
lorelei-sakai, line 76: This function does not seem to do the correct thing, at all, but it also doesn't seem to get use. You could just remove it.

If you want to keep it, it should do the right thing, though. Currently it looks like it just reads the devices file and declares every name it finds to be invalid? Also the doc claims that it filters names that can't be resolved, but it doesn't do any filtering.

src/perl/Permabit/LVMUtils.pm
```diff
@@ -0,0 +1,141 @@
...
+#
+# @param machine  The machine object to run commands on
+# @param devicePath  The path of the device to check
+#
+# @return true if the device is found in the devices file, false otherwise
+##
+sub isInLVMDevicesFile {
+  my ($machine, $devicePath) = assertNumArgs(2, @_);
+
+  my @validDevices = getValidDevicesInLVMDevicesFile($machine);
```
lorelei-sakai, line 61: This works, but it feels like it's doing more work than necessary. Instead of parsing the whole devices file to check for a single name, you could check for that single name as you read the file. In our cases, though, the file probably get large enough for this to be a big deal, so it probably doesn't matter that much.

src/perl/Permabit/LVMUtils.pm
```diff
@@ -0,0 +1,141 @@
...
+  assertNumArgs
+);
+use Permabit::RemoteMachine;
+
+use base qw(Exporter);
+
+our @EXPORT_OK = qw(
+  getInvalidDevicesInLVMDevicesFile
+  getValidDevicesInLVMDevicesFile
+  isLvmdevicesAvailable
```
lorelei-sakai, line 24: Alphabetically, these last two should be swapped; "isIn..." comes before"isLvm.."

src/perl/Permabit/LVMUtils.pm
```diff
@@ -0,0 +1,141 @@
...
+##
+package Permabit::LVMUtils;
+
+use strict;
+use warnings FATAL => qw(all);
+use English qw(-no_match_vars);
+use Log::Log4perl;
+
+use Permabit::Assertions qw(
+  assertMinMaxArgs
```
lorelei-sakai, line 14: I don't see assertMinMaxArgs being used, so you don't need it.
Also, I _think_ you don't need to import RemoteMachine either, is there something I'm missing there?

src/perl/Permabit/LVMUtils.pm
```diff
@@ -0,0 +1,141 @@
+##
+# Utility functions for working with LVM devices
+#
+# $Id$
+##
+package Permabit::LVMUtils;
```
lorelei-sakai, line 6: You know, all four of the function in this file take the machine as their primary argument. Woul it be better to just put these in UserMachine.pm instead of in a separate utility file?

UserMachine currently only has one lvm reference (and that it a string regex), but it does have other functions for using dmsetup and vgs. I also think that would mean you wouldn't need the exporter mechanism, they'd just be functions directly on the machine object.

lorelei-sakai (changes requested):
The biggest tihng is that getInvalidDevicesInLVMDevicesFiles() doesn't apear to do what it claims, but it also doesn't get used, that should get fixed one way or the other.

Otherwise mostly nits and refactoring questions.

src/perl/Permabit/LVMUtils.pm
bjohnsto: Moving functions to UserMachine.pm

src/perl/Permabit/LVMUtils.pm
lorelei-sakai: Okay. I had overlooked the --check argument, that makes sense. But I'm also fine with removing it.

src/perl/Permabit/LVMUtils.pm
lorelei-sakai: I'm a bit confused here because the mail message doesn't match the current comment, so I presume you edited it after the initial comment.

So top-line the  current spelling is fine. The old spelling is probably also fine. Performance may be important, but in our tests the file not be big enough that it matters that much.

One possible improvement in the current spelling would be to factor out a function to try to resolve the device name so that machinery can be written only once, and in a separate location. But also I agree that it's not worth bikeshedding about these details, we can look at it later maybe.

src/perl/vdotest/VDOTest/LVMDevicesTest.pm
```diff
@@ -0,0 +1,128 @@
...
+  if (!$machine->isLvmdevicesAvailable()) {
+    $log->info("LVM devices file is not available ... skipping test");
+    return;
+  }
+
+  my $rawDevice = $self->createTestDevice("raw");
+  my $devicePath = $rawDevice->getDevicePath();
+  $log->info("Created raw device: " . $devicePath);
+
+  my $deviceInFile = $machine->isInLVMDevicesFile( $devicePath);
```
lorelei-sakai, line 48: There's an extra space in the parameter list, before $devicePath. I think it's a cut and paste issue, because it seems to be in every isInLVMDevicesFile() call in these tests.

src/perl/Permabit/UserMachine.pm
```diff
@@ -684,6 +684,95 @@ sub resolveSymlink {
   return $resolvedPath;
 }
 
+
```
lorelei-sakai, line 687: You've added an extra blank line here, could we remove it?

src/perl/Permabit/BlockDevice/ISCSI.pm
```diff
@@ -228,29 +228,19 @@ sub getDevicePath {
...
-sub addToDevicesFile {
+sub getDevicePathForDevicesFile {
   my ($self) = assertNumArgs(1, @_);
   if ($self->isPassThrough()) {
-    # Pass through doesn't actually exist
-    return;
+    # Pass through doesn't have a separate device path
+    return undef;
   }
+
```
lorelei-sakai, line 237: I'd prefer to leave the blank line here, after the passthrough check. But it could be okay without.

lorelei-sakai (changes requested):
Literally just whitespace issues left, the rest looks good.

src/perl/vdotest/VDOTest/LVMDevicesTest.pm
lorelei-sakai: Okay, weird. You fixed the ones on lines 47 and 78.
Now you just need to fix the ones on lines 52, 84, 90, 96, 105, and 120. I did mean literally every call.

src/perl/Permabit/BlockDevice/ISCSI.pm
lorelei-sakai: Sorry about this, you did have the blank line there.

lorelei-sakai (changes requested):
Missed some of the weird formatting in the test. Otherwise, all good.

src/perl/vdotest/VDOTest/LVMDevicesTest.pm
bjohnsto: fixed

src/perl/Permabit/UserMachine.pm
bjohnsto: fixed

lorelei-sakai (approved):
Ugh, tests failed. And test artifact upload also failed (missing filename esacaping?)
But the commits themselves are fine.

PR #339, opened 2025-09-24
[OPS-21689] dm vdo: add VDO_USE_NEXT macro for upstream change with struct bio.

C2Redhat:
Apply the VDO_USE_NEXT macro to handle recent upstream changes to struct bio.
This update removes the bio->bi_io_vec member and replaces its references
with bio_inline_vecs(), reflecting the modifications from upstream
commit d86eaa0f3c56 ("block: remove the bi_inline_vecs variable sized array from struct bio")

### DIFF
--- a/src/c++/vdo/base/vio.c
+++ b/src/c++/vdo/base/vio.c
@@ -11,6 +11,9 @@
 #ifdef __KERNEL__
 #include <linux/ratelimit.h>
 #endif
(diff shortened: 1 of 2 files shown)


lorelei-sakai (approved):
The code content looks correct now. As previously mentioned, I'd prefer you adjust the commit message to include the title of the upstream patch, but it's not a big deal.

PR #342, opened 2025-09-25
dm vdo user: Update READ/WRITE_ONCE definitions [VDO-5943]

raeburn:
The old TEST_INTERNAL versions of the definitions, dating back about 7 years, seemed to trigger some test failures when enabled for some directories. These updated versions, based loosely on current kernel code (minus per-architecture customizations, plus suppression for GCC diagnostics only used in the user-mode build), seem to work.

### DIFF
--- a/src/c++/uds/userLinux/uds/linux/atomic.h
+++ b/src/c++/uds/userLinux/uds/linux/atomic.h
@@ -208,58 +208,59 @@ static inline void smp_mb__after_atomic(void)
  * the required ordering.
  */
 
-#define READ_ONCE(x)                                 \
-  __extension__ ({                                   \
-    union { __typeof__(x) __val; char __c[1]; } __u; \
-    read_once_size(&(x), __u.__c, sizeof(x));        \
-    __u.__val;                                       \
-  })
+#define __scalar_type_to_expr_cases(type)				\
+		unsigned type:	(unsigned type)0,			\
+		signed type:	(signed type)0
 
-#define WRITE_ONCE(x, val)                          \
-  __extension__ ({                                  \
-    union { __typeof__(x) __val; char __c[1]; } __u \
-      = { .__val = (__typeof__(x)) (val) };         \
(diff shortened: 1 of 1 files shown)


lorelei-sakai:
TEST_INTERNAL means exactly that: It is only for tests and should not be published anywhere. So yes, we need to strip test-internal code from anything we publish. If there is code that we should be including in our userpace package (for whatever reason) it should not use TEST_INTERNAL, it should be using a different tag. (Something like VDO_USER, for example.)

TEST_INTERNAL is currently only define for UDS code, not for VDO code. Is there actually any reason to change this?

raeburn:
If we keep the conditional stuff as is, then the reason for update is that apparently the current TEST_INTERNAL implementation doesn't work with the current compilers, at least for some types or some uses. (My guess is it's a pointer type issue.) That may not break any of our code right now, if we don't use it on pointers in code that defines TEST_INTERNAL, which I think is the case. And if it's not just pointers, well, most of our uses are for statistics where we wouldn't notice anyway.

In short, I think the old version is broken, but we've gotten lucky in how we use it, where most cases won't matter, and the race conditions are few.

When TEST_INTERNAL isn't defined, our user-mode programs use dumb versions (like "(dest) = (src)") that give no guarantee that the value will be read or written only once as the caller intends, nor guard against reordering. It's simply reading a value, or assigning to a value, and if the compiler finds reason during optimization to change how it's done, it will. Again, most of the uses are for stats where we won't care...but not all of them are. Luckily, it looks like all the user-mode tools I just tried out run single-threaded, so nothing going to customers would care. If we ever make one of our tools multithreaded, that could change.

So, if we don't want the big, ugly version in the shipping user-mode tools, we'll probably be okay. We should still fix the TEST_INTERNAL version to be pointer-safe, to deal with possible lurking issues in our tests.

raeburn:
Closing in favor of approach in PR #385

PR #344, opened 2025-09-30
[VDO-5787] Add synchronous r/w function

bjohnsto:
Add code to read from and write to the disk synchronously. Use that new function in read_geometry_block().

### DIFF
--- a/src/c++/vdo/base/io-submitter.c
+++ b/src/c++/vdo/base/io-submitter.c
@@ -394,6 +394,37 @@ void __submit_metadata_vio(struct vio *vio, physical_block_number_t physical,
 }
 
 /**
+ * vdo_submit_metadata_vio_wait() - Submit I/O for a metadata vio and wait for completion.
+ * @vio: the vio for which to issue I/O
+ * @physical: the physical block number to read or write
+ * @operation: the type of I/O to perform
+ *
+ * The function operates similarly to __submit_metadata_vio except that it will
+ * block until the work is done. It can be used to do i/o before work queues
+ * and thread completions are set up.
+ *
+ * Return: VDO_SUCCESS or an error.
+ */
+int vdo_submit_metadata_vio_wait(struct vio *vio,
+				 physical_block_number_t physical,
+				 blk_opf_t operation)
+{
+	int result;
+
+#ifdef VDO_INTERNAL
+	vio->bio_submission_jiffies = jiffies;
+
+#endif /* not VDO_INTERNAL */
+	result = vio_reset_bio(vio, vio->data, NULL, operation | REQ_META, physical);
+	if (result != VDO_SUCCESS)
+		return result;
+
+	bio_set_dev(vio->bio, vdo_get_backing_device(vio->completion.vdo));
+	submit_bio_wait(vio->bio);
+	return blk_status_to_errno(vio->bio->bi_status);
+}
+
+/**
  * vdo_make_io_submitter() - Create an io_submitter structure.
  * @thread_count: Number of bio-submission threads to set up.
  * @rotation_interval: Interval to use when rotating between bio-submission threads when enqueuing
--- a/src/c++/vdo/base/vdo.c
+++ b/src/c++/vdo/base/vdo.c
@@ -317,8 +317,6 @@ static int initialize_super_block(struct vdo *vdo, struct vdo_super_block *super
  */
 static int __must_check read_geometry_block(struct vdo *vdo)
 {
-	struct vio *vio = &vdo->geometry_block.vio;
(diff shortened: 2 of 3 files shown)


src/c++/vdo/base/io-submitter.c
```diff
@@ -393,6 +393,38 @@ void __submit_metadata_vio(struct vio *vio, physical_block_number_t physical,
 	vdo_launch_completion_with_priority(completion, get_metadata_priority(vio));
 }
 
+/**
+ * vdo_submit_metadata_vio_wait() - Submit I/O for a metadata vio and wait for completion.
+ * @vdo: the vdo to use
```
lorelei-sakai, line 398: This function doesn't take a @vdo argument, you should drop it from the doc.

src/c++/vdo/base/io-submitter.c
```diff
@@ -393,6 +393,38 @@ void __submit_metadata_vio(struct vio *vio, physical_block_number_t physical,
...
+	vio->bio_submission_jiffies = jiffies;
+
+#endif /* not VDO_INTERNAL */
+	result = vio_reset_bio(vio, vio->data, NULL, operation | REQ_META, physical);
+	if (result != VDO_SUCCESS)
+		return result;
+
+	bio_set_dev(vio->bio, vdo_get_backing_device(vio->completion.vdo));
+	submit_bio_wait(vio->bio);
+	return blk_status_to_errno(vio->bio->bi_status);
```
lorelei-sakai, line 425: This shouldn't be returning an errno, given the earlier path (return from vio_reset_bio) returns a VDO error code. I'm not sure what errors can result here; is -EIO sufficient or is there a more appropriate error here? What do other metadata write paths do?

src/c++/vdo/base/vdo.c
```diff
@@ -317,47 +317,17 @@ static int initialize_super_block(struct vdo *vdo, struct vdo_super_block *super
  */
 static int __must_check read_geometry_block(struct vdo *vdo)
 {
-	struct vio *vio;
-	char *block;
 	int result;
-
```
lorelei-sakai, line 285: Even though the next line is a comment, I think we still need a blank line here after local variable declarations.

src/c++/vdo/base/vdo.c
```diff
@@ -317,47 +317,17 @@ static int initialize_super_block(struct vdo *vdo, struct vdo_super_block *super
...
 	 * This is only safe because, having not already loaded the geometry, the vdo's geometry's
 	 * bio_offset field is 0, so the fact that vio_reset_bio() will subtract that offset from
 	 * the supplied pbn is not a problem.
 	 */
-	result = vio_reset_bio(vio, block, NULL, REQ_OP_READ,
-			       VDO_GEOMETRY_BLOCK_LOCATION);
-	if (result != VDO_SUCCESS) {
-		free_vio(vdo_forget(vio));
-		vdo_free(block);
+	result = vdo_submit_metadata_vio_wait(&vdo->geometry_block.vio,
```
lorelei-sakai, line 326: A minor quibble, but it might scan better if the switch to using the new geometry block structure was in the previous (first) commit instead of in the second one.

The second commit talks about adding vdo_submit_metadata_vio_wait(), but the switchover to the geometry_block structure isn't really related, and I think the switchover gets a little lost here.

lorelei-sakai (changes requested):
I hadn't realized (though I should have) these are substantially the same commit as in the larger series. ome comments copied from there, for things which don't appear to have been addressed, and a couple minor new things.

src/c++/vdo/base/io-submitter.c
bjohnsto: Tricky to know what to do here. I'm not sure -EIO is sufficient. We might have ENOMEM as well I think. The metadata paths seem to merely look to see if its VDO_SUCCESS and if not set the completion result to the blk_status_to_errno return value.

src/c++/vdo/base/vdo.c
bjohnsto: ok i'll swap them around

src/c++/vdo/base/vdo.c
bjohnsto: ok i'll add a blank line

src/c++/vdo/base/io-submitter.c
lorelei-sakai: Honestly, if the existing metadata pathways return the bi_status errno directly, it seems reasonable that we do that here, too. It just looked weird. But if it's consistent with other metadata writes, it's fine.

PR #346, opened 2025-10-03
Reduce Makefile dependecies and other overlay script improvements

lorelei-sakai:
First in a series of cleaning up and committing improvements to the overlay script that I've been using in my workspace for months, but which I never actually committed.

This one reduces the overlay script's reliance on Makefile targets, and also makes several minor improvements to the format of the generated upstream patches.

ETA: This is a newer version of work that was originally PR #37, so check there for some discussion and comments. But I managed to screw that PR up somehow, so I've made this new one.

### DIFF
--- a/src/packaging/kpatch/commit_to_overlay.sh
+++ b/src/packaging/kpatch/commit_to_overlay.sh
@@ -364,7 +364,7 @@ _cleanup() {
     rm -fv ${commit_file}
   fi
 
-  unset -f COMMIT_FILE DEBUG GITDATE GITAUTHOR LINUX_SRC
+  unset -f COMMIT_FILE DEBUG LINUX_SRC
 }
 
 ###############################################################################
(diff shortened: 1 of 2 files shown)


src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -445,19 +450,30 @@ for change in ${COMMIT_SHAS[@]}; do
...
+  cp -r ${WORK_DIR}/dm-vdo/* ${LINUX_VDO_SRC} && \
+  cp -f ${WORK_DIR}/vdo.rst ${WORK_DIR}/vdo-design.rst ${LINUX_DOC_SRC}/
+  if [[ $? != 0 ]]; then
+    echo -e "${COLOR_RED}ERROR: Generating the kernel overlay failed.${NO_COLOR}"
+    exit
+  fi
 
+  echo "Applying the kernel overlay to branch '${OVERLAY_BRANCH}'"
+  pushd ${LINUX_SRC} &> /dev/null
+  git add . && \
```
raeburn, line 467: Would "git -C ${LINUX_SRC} ..." be more straightforward than the pushd/popd sequence?

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -445,19 +450,30 @@ for change in ${COMMIT_SHAS[@]}; do
...
+  git commit "${gitauthor}" "${gitdate}" --file "${commit_file}"
+  popd &> /dev/null
+  echo -e "Committed"
   if [[ $? != 0 ]]; then
     return=$(tail -5 ${build_log_file} | grep "^[nN]othing to commit" | wc -l)
```
raeburn, line 472: 1) If the "git commit" output isn't written to build_log_file any more, this grep check won't work, and pointing the user to build_log_file isn't suitable.
2) something like "git diff-index --quiet --cached HEAD" might tell in advance if there's anything to commit

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -445,19 +450,30 @@ for change in ${COMMIT_SHAS[@]}; do
...
+    exit
+  fi
 
+  echo "Applying the kernel overlay to branch '${OVERLAY_BRANCH}'"
+  pushd ${LINUX_SRC} &> /dev/null
+  git add . && \
+  git commit "${gitauthor}" "${gitdate}" --file "${commit_file}"
+  popd &> /dev/null
+  echo -e "Committed"
   if [[ $? != 0 ]]; then
```
raeburn, line 471: Both popd and echo commands will reset the value of $? set by git so this check won't work now

raeburn (changes requested):

PR #348, opened 2025-10-07
[TECH-219][TECH-151] Allow use of an existing kernel repo for upstream overlays

lorelei-sakai:
Add some parsing to allow the kernel repo to be spcified by URL or by path. If it is specified by path, use that repo instead of cloning a new kernel repo, and verify that the required branch exists in the kernel repo. This avoids the kernel repository clone, which can take a significant amount of time.

In practice, I use this option to create an overlay branch in my local fork of dm-linux (which mirrors the kernel), using the manual push option. This allows me to check and tweak the overlay branch details and then manually push it up to create a PR for dm-linux.

### DIFF
--- a/src/packaging/kpatch/commit_to_overlay.sh
+++ b/src/packaging/kpatch/commit_to_overlay.sh
@@ -25,13 +25,19 @@
 #     This local clone will be removed upon exiting the script, unless the DEBUG flag is set
 #     or the manual push option is selected. It is recommended that the SSH URL is used,
 #     as GitHub prompts for a username and password when an HTTPS URL is used.
+#     The <kernel_tree_branch_name> argument must be a branch in the given repository, which
+#     will be the starting point of the new overlay branch.
 #
 #   - If the specifier is a path, it must point to the top level of a kernel clone, and must not
 #     be a bare repository. This clone will create the new overlay branch but it will not otherwise
 #     be altered. A relative path will be interpreted relative to the current working directory.
+#     The <kernel_tree_branch_name> argument must resolve to a commit in the given repository,
+#     which will be the starting point of the new overlay branch. Additionally, if the
+#     <kernel_tree_branch_name> contains a slash, the substring before the first slash is
+#     used as the remote to push the new branch to. The default remote is 'origin'.
 #
 # - The local branch option (-b) relies on detecting differences between the input branch and
-#   origin/HEAD. It is only effective before a rebase of the fork is performed.
+#   the remote HEAD. It is only effective before a rebase of the fork is performed.
 # - The local vdo source tree will be cloned into a local absolute path. By default, it will
 #   be removed upon exiting the script, unless the DEBUG flag is set.
 # - Output from building the VDO tree while processing each change can be found in the
@@ -40,7 +46,8 @@
 #   flag is set.
 #
 # Potential Future Modifications:
-# - Remove the kernel_tree_branch_name input argument if it is always main.
+# - Add a proper optional argument for specifying the remote instead of using part of
+#   <kernel_tree_branch_name>.
 ##
 
 if [[ -z ${DEBUG} ]]; then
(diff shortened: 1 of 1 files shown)


raeburn:
typos "exiting" -> "existing", "spcified" -> "specified"

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -108,6 +110,21 @@ process_args() {
       ;;
   esac
 
+  if [[ ${KERNEL_REPO_URL} =~ @.*?: ]]; then
+    echo -e "Using kernel repo URL (${KERNEL_REPO_URL})"
+  else
+    echo -e "Using kernel repo path (${KERNEL_REPO_URL})"
+    LINUX_SRC=${RUN_DIR}/${KERNEL_REPO_URL}
```
raeburn, line 117: So the local-repo argument must also be a relative path from the working directory, not an absolute path?
And it must not be a bare repo?

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -108,6 +110,21 @@ process_args() {
...
+    KERNEL_REPO_URL=$(git -C ${LINUX_SRC} remote get-url origin)
```
raeburn, line 118: Assuming "origin" is the name of the remote you want seems iffy. I frequently use repos with multiple remotes. One might have a kernel repo with the torvalds, linux-next, linux-dm, and ark remotes.

Related to the usage change, though, if we're pointing at a local repo, can the branch name be something like "origin/main" or "linux-dm/foobar" that identifies the remote? Or does that not make sense?

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -108,6 +110,21 @@ process_args() {
       ;;
   esac
 
+  if [[ ${KERNEL_REPO_URL} =~ @.*?: ]]; then
+    echo -e "Using kernel repo URL (${KERNEL_REPO_URL})"
```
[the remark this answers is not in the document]
raeburn: Ah. Yeah, that may be too strict. Looking through a few of my local repos, I've got remotes that look like "https://host/path", "host:path", "git@github.com:raeburn/path", "git://host/path", or just "/path", though remotes on local disk are a case we probably don't need to handle.

Checking for the colon is probably enough.

src/packaging/kpatch/commit_to_overlay.sh
raeburn: A bare repo can handle branch changes and commits but doesn't automatically give you a working tree to drop files into. It's possible to muck around with defining your own index file and update it manually with files set up in a temporary working directory, but requiring that the repo not be bare is simpler. Running "git rev-parse --is-bare-repository" would let us confirm that before we go making changes. (rev-parse has a lot of random seeming stuff in it.)

This does mean it's got to be a clean tree available for scribbling into. With the git worktree support we could create a scratch working tree for the script's purposes and delete it when we finish. Though of course the user can just do that before running the script, too. In any case that's also for later consideration.

src/packaging/kpatch/commit_to_overlay.sh
raeburn: It's assuming "origin" for the current repo (source of the changes), and it's correct about the newly checked out kernel repo because it's creating the clone. Pointing to an existing local repo that might not have used "origin" for the desired remote repo is where the assumption becomes dicey.

src/packaging/kpatch/commit_to_overlay.sh
lorelei-sakai: Re the clean tree, that doesn't seem like a huge assumption to me. I haven't tried it yet, but presumably you could have a dedicated worktree for doing overlays that won't ever have other changes in it. The script does check that when you do a branch, the branch succeeds (which won't happen if the index is too different from the branch you're moving to). And the overlay process completely replaces the tree so you shouldn't pick up any straggling changes either (though I suppose it would effectively delete them).

I take your point that just as we make a new branch, we could create an entire new worktree, and that would guarantee it was clean relative to the branch we want. Definitely worth future consideration.

src/packaging/kpatch/commit_to_overlay.sh
lorelei-sakai: Not a new parameter, but I've allowed adding the remote name to the base branch. But now it occurs to me that I've done it in a weird way because I couldn't actually resolve the remote branch name locally. Do we need to deal with that now?

src/packaging/kpatch/commit_to_overlay.sh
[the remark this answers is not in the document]
raeburn: I was thinking at least a usage/comment update to demonstrate exactly what you intend for the interface changes to be. Instead, I'm guessing what you intend based on the code changes.

src/packaging/kpatch/commit_to_overlay.sh
raeburn: No, probably not

raeburn (changes requested):
Mostly looks good, just a question on diff vs diff-index, and the updated usage info.

src/packaging/kpatch/commit_to_overlay.sh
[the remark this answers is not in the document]
lorelei-sakai: Usage notes added, at least for the new features. General usage cleanup will follow later.

src/packaging/kpatch/commit_to_overlay.sh
```diff
@@ -463,7 +463,7 @@ for change in ${COMMIT_SHAS[@]}; do
   fi
 
   echo "Applying the kernel overlay to branch '${OVERLAY_BRANCH}'"
-  git -C ${LINUX_SRC} diff-index --cached --quiet HEAD
+  git -C ${LINUX_SRC} diff --quiet HEAD
```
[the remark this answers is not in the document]
lorelei-sakai: diff-index didn't seem to always detect the changes I wanted, especially when handling a series of commits (e.g. it worked for the first commit, but for the second commit it appeared to detect all cumulative changes). 'git diff' seems to reliably detectwhat I want. We can refactor later if there's a more efficient way to get this information, though.

raeburn (approved):

PR #353, opened 2025-10-21
[VDO-5966] Update nightly on RHEL 10 branch

lorelei-sakai:
This merges two commits from the main branch to 8.3. The first changes the RHEL target class to RHEL10, and the second splits out the upstream tests (e.g. linux-next testing) so that the RHEL nightlies don't run them.

Both of these commits need to be merged because we're finally trying to run RHEL10 nightly from the actual release branch now.


raeburn:
The changes look good to me but where did the build problem come from?
EDIT: Never mind, just caught up in slack...

PR #365, opened 2025-10-29
Merge vdo.spec and src-dist RPM

C2Redhat:
This merges commit d160f576 ("Update vdo.spec and src-dist RPM to include to install vdocalculatesize") 
to 8.3 from PR 364


lorelei-sakai (approved):
Look good.

PR #373, opened 2025-11-06
[VDO-5928][VDO-5878][VDO-5778] Fix migration tests with RHEL 9

lorelei-sakai:
The first commit updates the Chlorine version used for upgrade tests; the old version 8.2.4.9 can no longer be built by dkms on our lab machines for various reasons. This fixes VDO-5878, and all its children (VDO-5928, VDO-5949, and VDO-5964).
Once the build problem is fixed, the tests start failing with a GLIBC version difference: utilities built on Fedora can't run on RHEL 9, as documented in VDO-5778. The second commit fixes this by using the utilities (genDataBlocks and genDiscard) installed from our test utility package that was created for this purpose some months ago. This fixes VDO-5778, and also changes how all tests call these utilities.

### DIFF
--- a/src/perl/Permabit/GenDataFiles.pm
+++ b/src/perl/Permabit/GenDataFiles.pm
@@ -156,7 +156,6 @@ my $log = Log::Log4perl->get_logger(__PACKAGE__);
 
 my $setNumber = 0;
 my $BLOCK_SIZE = 4 * $KB;
-my $GEN_DATA_BLOCKS = "src/c++/vdo/bin/genDataBlocks";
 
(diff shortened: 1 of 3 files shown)


raeburn:
With these changes, do we still share genDiscard etc executables that are now never used? If so, can we stop?

lorelei-sakai:
> With these changes, do we still share genDiscard etc executables that are now never used? If so, can we stop?

Good point, I'll check. The public testing isn't working in a way that's probably some work to fix, so I'll take a look at this next week.

PR #383, opened 2025-11-21
dm vdo user: add vdoCalculateSize to ignore list

lorelei-sakai:
When Chung added the vdoCalculateSize, we forgot to add it to the ./gitignore file so it keeps showing popping up in my draft commits.


C2Redhat (approved):
yay

PR #391, opened 2025-12-11
Fix some conflicts when running dnf/yum as a user [VDO-5867,VDO-5960,VDO-6013]

raeburn:
Multiple tests running concurrently on different machines but sharing an NFS home directory can collide on naming the cache files used by yum/dnf. Work around this by using local directories for the cache.

Previous fixes had to make XDG_STATE_HOME a local directory as well; this applies in exactly the same cases.

However, the previous fix created all the files or directories directly in /tmp, and didn't clean them up, potentially leading to permission problems if the machine is used by different users for these tests. So we move them to the test's scratch directory.

We also need to catch errors when running "yum list".

### DIFF
--- a/src/perl/Permabit/BlockDevice/VDO.pm
+++ b/src/perl/Permabit/BlockDevice/VDO.pm
@@ -558,11 +558,16 @@ sub suspend {
 sub getModuleVersion {
   my ($self) = assertNumArgs(1, @_);
   if ($self->{useUpstreamModule}) {
-    my $getVerCmd = "XDG_STATE_HOME=/tmp " .
+    # Concurrent tests on multiple machines can collide accessing NFS
+    # homedirs.
+    my $scratchDir = $self->getMachine()->getScratchDir();
+    my $xdgStateDir = "$scratchDir/xdgState";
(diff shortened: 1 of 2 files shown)


src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -35,6 +35,7 @@ use Permabit::PlatformUtils qw(getDistroInfo);
 use Permabit::ProcessUtils qw(delayFailures);
 use Permabit::Statistics::VDO;
 use Permabit::SystemUtils qw(
+  assertCommand
```
bjohnsto: Why are we using assertCommand directly here? Couldn't we use something like runOnHost. Won't that croak is the errno is not 0?

raeburn: Because our various run-command-and-maybe-croak-on-error routines and methods are such a mess I don't remember them all? And runCommand -> assertCommand seemed like a minimal change to do the job...

Using runOnHost would make sense here, I'll change it. But the other runCommand uses that don't need changing I'm going to leave alone...

PR #392, opened 2025-12-11
Sync storage before migrating iSCSI volume [VDO-5758]

raeburn:
This is an attempt to address an intermittent failure with data writes disappearing on migration.


bjohnsto (approved):
Lets hope for the best

PR #396, opened 2026-01-08
Always use grubby for editing boot options. [VDO-5988]

raeburn:


lorelei-sakai (changes requested):
The code change looks reasonable.

I'd like a more detailed commit message, though, saying a bit about what the problem is and why this is a reasonable fix, just a sentence or two. (It would probably be fine to put that in the ticket, but I don't see an explanation there, either.)

I'd also prefer the title use the kernel format, but for an internal change like this, it's not a big deal.

PR #399, opened 2026-01-12
Remove VDO_USE_NEXT conditional for bi_io_vec

C2Redhat:
Refactor VDO_USE_NEXT to VDO_USE_ALTERNATE. Updated the #ifdef logic
to prioritize the bio_inline_vecs interface introduced in Fedora 6.18
while preserving a legacy path for older environments.


C2Redhat:
> This change looks basically correct to me, but the private test packaging build failed.
> 
> Do we need to adjust for that, or does that simply mean we need to update the lab machines as well? (Build failed on 6.17.11-200.fc42.x86_64.)

Yes,  I am updating all the system now so we won't have a version confliict.

PR #400, opened 2026-01-14
[VDO-5752] [VDOSTORY-292] Add a new suite of tests for in kernel formatting.

bjohnsto:
This PR consists of the following changes:
vdotests.suites: create an actual suite that consists of a broad enough range of tests to sufficiently test the format in kernel code.
VDO.pm: create a new new build type to run the new suite that sets the appropriate kernel formatting flag.

### DIFF
--- a/src/perl/vdotest/vdotests.suites
+++ b/src/perl/vdotest/vdotests.suites
@@ -107,6 +107,19 @@ $suiteNames{filesystem}
      "Zero01",
     ];
 
+# VDO tests that exercise the vdoformat in kernel code.
+$suiteNames{kernelformat}
+  = [
+     (map { "Bare$_" } (
+       @{$suiteNames{basic}},
+       "Direct06",
+       "AuditTool",
+       "Compress01",
+       "Direct04",
+     )),
+     "DoryRebuildFast",
+    ];
+
 ###########################################################################
 # Rebuild stress tests
 ##
@@ -915,6 +928,7 @@ $aliasNames{RawDirect06}
 
 $aliasPrefixes{"512B"}   = "--emulate512Enabled=1 --compressibleChunkSize=512";
 $aliasPrefixes{Dist}     = "--useDistribution=1";
+$aliasPrefixes{Bare}     = "--deviceType=vdo";
 $aliasPrefixes{Ext3}     = "--fsType=ext3";
 $aliasPrefixes{Ext4}     = "--fsType=ext4";
 $aliasPrefixes{Fedora42} = "--clientClass=FEDORA42";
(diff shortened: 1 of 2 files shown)


raeburn (commented):
With all these suites expanded, how many tests is this? What about the new code are we trying to test with all these new tests? Do we really need more than the formatInKernel version of Basic01, and a few tests of passing format parameters through?

src/perl/vdotest/vdotests.suites
```diff
@@ -915,6 +931,7 @@ $aliasNames{RawDirect06}
 
 $aliasPrefixes{"512B"}   = "--emulate512Enabled=1 --compressibleChunkSize=512";
 $aliasPrefixes{Dist}     = "--useDistribution=1";
+$aliasPrefixes{Dmvdo}    = "--deviceType=vdo";
```
lorelei-sakai: Dmvdo seems like a vague prefix since all vdo devices are dm-vdo...
This is an Unmanaged device, right? That is, one using dmsetup directly and not using lvm tooling?
This could use a better name, but I'm not thinking of one immediately.

src/perl/vdotest/vdotests.suites
```diff
@@ -107,6 +107,22 @@ $suiteNames{filesystem}
...
     ];
 
+# VDO tests that exercise the vdoformat in kernel code.
+$suiteNames{kernelformat}
+  = [
+     map { "Dmvdo$_" } (
+       @{$suiteNames{basic}},
+       @{$suiteNames{direct}},
+       @{$suiteNames{genData}},
+       @{$suiteNames{collide}},
```
lorelei-sakai: What value are the collide tests adding here? It's not clear to me they test anything about the metadata layout.

src/perl/vdotest/vdotests.suites
```diff
@@ -107,6 +107,22 @@ $suiteNames{filesystem}
...
+       "AuditTool",
+       "Compress01",
+       "QuickDedupe",
+       "Zero01",
+     ),
+     "DoryRebuildFast",
```
lorelei-sakai: Why not add the prefix on this test?
Or looked at a different way, why include this test (also in checkin) if we're not changing the way we run it?

src/perl/vdotest/vdotests.suites
lorelei-sakai: Were you going to change the prefix name? Maybe something like "Bare" or "NoLVM", that communicates how this is different from the default device setup (with lvm).

bjohnsto: I went with Bare.

PR #410, opened 2026-01-22
dm vdo test: Enable verbose XFS logging for tests mounting file systems [VDO-5647]

raeburn:
Bump the fs.xfs.error_level sysctl setting to 11 before mounting the first XFS file system of the test, and restore the old value at the end of the test.

## Summary by CodeRabbit

* **Chores**
  * Enhanced XFS filesystem configuration management by implementing automatic per-machine setup for error level settings during mount operations, with improved logging for tracking configuration changes and restoration actions.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/perl/Permabit/FileSystem/Xfs.pm
+++ b/src/perl/Permabit/FileSystem/Xfs.pm
@@ -9,12 +9,15 @@ use strict;
 use warnings FATAL => qw(all);
 use English qw(-no_match_vars);
 use File::Basename qw(basename);
+use Log::Log4perl;
 use Permabit::Assertions qw(assertFalse assertMinMaxArgs assertNumArgs);
 use Permabit::PlatformUtils qw(isMaipo);
 use Permabit::Utils qw(makeFullPath);
 use Storable qw(dclone);
 
 use base qw(Permabit::FileSystem);
+
+my $log = Log::Log4perl->get_logger(__PACKAGE__);
 
 #############################################################################
 # @paramList{new}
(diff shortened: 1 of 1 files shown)


raeburn:
@coderabbitai review

src/perl/Permabit/FileSystem/Xfs.pm
```diff
@@ -62,11 +65,55 @@ sub getBlockSizeMkfsOption {
...
+#############################################################################
+# Per-machine setup (once per test) of XFS error_level
+##
+sub _ensureXfsErrorLevelConfigured {
+  my ($self) = assertNumArgs(1, @_);
+  my $machine = $self->getMachine();
+
+  if (!$machine->{_xfs_error_level_configured}) {
+    # Ensure xfs sysctl is present (ignore failures)
+    $machine->sendCommand("sudo modprobe xfs");
```
lorelei-sakai: Won't we need to make sure the xfs module is loaded to do anything with xfs? If we really need this step at all (which seems questionable), it seems like it should be earlier in the initialization of the  perl Xfs object.

src/perl/Permabit/FileSystem/Xfs.pm
```diff
@@ -62,11 +65,55 @@ sub getBlockSizeMkfsOption {
...
+
+    # Read and remember current value
+    $machine->runSystemCmd("sysctl -n fs.xfs.error_level");
+    my $originalErrorLevel = $machine->getStdout();
+    chomp($originalErrorLevel);
+
+    $machine->{_xfs_error_level_configured} = 1;
+
+    # Only register cleanup if we captured a value; restore on cleanup; log failures, don’t croak
+    if ($originalErrorLevel ne '') {
```
lorelei-sakai: Can this happen? Seems like this should be gated on the flag for "have we set a cleanup step yet". Which I guess we already are.

lorelei-sakai (changes requested):
This would be fine as written , I suppose.
It seems way too complicated, though. Why bother adding a cleanup step at all? If we want to that this whenever we use xfs in a test, just leave it set all the time. Andy said that Eric Sandeen said that the xfs logging isn't even that expensive. We'd have extra logging for xfs filesystems that aren't being run as part of our tests, but they're test machines, how frequent is that?

src/perl/Permabit/FileSystem/Xfs.pm
raeburn: I guess only if sysctl doesn't output anything, and yet still gives a successful exit status, which is a combination that shouldn't happen. I'll drop the check.

lorelei-sakai (approved):
Okay, this seems reasonable to me.

I'm still not sure the cleanup effort is really that helpful, but we  can always adjust it later if it ends up doing something odd.

PR #411, opened 2026-01-22
dm vdo test: Coordinate vdorecover checking loop with test [VDO-5647]

raeburn:
Instead of sending "y" to the vdorecover script every 10 seconds and hoping that that and the test's freeing of space are reasonably in sync, use a sentinel file to indicate when the test is ready for vdorecover to continue doing its work.

Also, drop support for varying the TMPFILESZ parameter to vdorecover that was never changed in the test.

## Summary by CodeRabbit

* **Tests**
  * Improved VDO recovery testing coordination and synchronization mechanisms for more reliable test execution.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/perl/vdotest/VDOTest/Vdorecover.pm
+++ b/src/perl/vdotest/VDOTest/Vdorecover.pm
@@ -9,6 +9,7 @@ use strict;
 use warnings FATAL => qw(all);
 use English qw(-no_match_vars);
 use File::Basename qw(dirname);
+use Carp qw(confess);
 use Log::Log4perl;
 use Permabit::Assertions qw(
   assertEqualNumeric
@@ -17,7 +18,7 @@ use Permabit::Assertions qw(
 );
 use Permabit::Constants;
 use Permabit::GenDataFiles qw(genDataFiles);
-use Permabit::Utils qw(retryUntilTimeout);
+use Permabit::Utils qw(makeFullPath retryUntilTimeout);
 use Permabit::VDOTask::SliceOperation;
 
 use base qw(VDOTest);
@@ -25,7 +26,7 @@ use base qw(VDOTest);
 my $log = Log::Log4perl->get_logger(__PACKAGE__);
 
 my $DEFAULT_TMP_STORAGE_SIZE = 5 * $GB / $KB;
-my $INTERVAL = 10;
+my $RECOVER_WAIT_TIME = 200;
 
 #############################################################################
 # @paramList{getProperties}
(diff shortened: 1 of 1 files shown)


raeburn:
@coderabbitai review

src/perl/vdotest/VDOTest/Vdorecover.pm
```diff
@@ -26,6 +26,9 @@ my $log = Log::Log4perl->get_logger(__PACKAGE__);
 
 my $DEFAULT_TMP_STORAGE_SIZE = 5 * $GB / $KB;
 my $INTERVAL = 10;
+my $SENTINEL_BASENAME = "vdorecover-continue";
+
+
```
lorelei-sakai: Lots of extra blank space here.

src/perl/vdotest/VDOTest/Vdorecover.pm
```diff
@@ -111,7 +125,11 @@ sub testBasic {
   $slice->write(tag => "recov", dedupe => 0.9, direct => 1);
 
   my $vdoDeviceName = $self->_getVDODevicePath();
-  $self->runVdorecover($vdoDeviceName);
+  my $machine = $self->getDevice()->getMachine();
+  my $sentinel = $self->_sentinelPath($machine);
+  $machine->runSystemCmd("rm -f '$sentinel'");
+  $machine->runSystemCmd("touch '$sentinel'");
```
lorelei-sakai: A little pedantic perhaps, but in other test cases, you touch the sentinel file after starting vdorecover. Shouldn't this testcase use the same pattern?

src/perl/vdotest/VDOTest/Vdorecover.pm
```diff
@@ -286,4 +317,19 @@ sub testFullFilesystem {
   sleep($INTERVAL);
 }
 
+# Override tear_down to perform test-specific cleanup:
+# - remove the sentinel file if present
+sub tear_down {
+  my ($self) = assertNumArgs(1, @_);
+
+  my $machine;
+  eval { $machine = $self->getDevice()->getMachine(); };
```
lorelei-sakai: Do you need an eval here? Can this fail?
At worst it seems like spelling it as
"if defined($device) { $machine = $device->getMachine();}"
is clearer to read.

But  I really think this can't fail, at worst we end up creating a UserMachine at this point.
I guess what I'm saying is, it looks like you're trying to handle potential problems that we already handle.

src/perl/vdotest/VDOTest/Vdorecover.pm
```diff
@@ -43,6 +46,18 @@ our %PROPERTIES
...
+#############################################################################
+# Return the path to the sentinel file for a specific machine.
+#
+# @param machine  The machine object
+#
+# @return the full path to the sentinel file
+##
+sub _sentinelPath {
+  my ($self, $machine) = assertNumArgs(2, @_);
+  return makeFullPath($machine->getScratchDir(), $SENTINEL_BASENAME);
```
lorelei-sakai: We only use $SENTINEL_BASENAME once, here. should we just hard code it, or define it in this function.
Calling it "basename" which we don't add anytihng to it feels a bit weird, too.
Could be fine, just wanted to mention it.

lorelei-sakai (changes requested):
Looks cleaner this time, Mostly just some formatting and error handling items.

I kind of feel like the LLM is being extra paranoid because it doesn't know our code, but we don't usually need that.

PR #419, opened 2026-02-03
[VDO-6059] Fix CI for 8.3 branch

bjohnsto:
Update the the test Makefiles to run 8.3 branch on RHEL10 machines. 

This PR needs to wait on https://github.com/dm-vdo/vdo-devel/pull/430 to be done first as it requires an 8.3 CI file to exist on the main branch.


bjohnsto:
> The contents seem reasonable.
> 
> But, if this is fixing CI testing on the 8.3 branch... shouldn't the CI testing pass? (The test failures look as though the STATIC label is not being defined in these builds, somehow.)

I've updated the PR.

PR #421, opened 2026-02-12
[VDO-6024] Fix upstream tests that were passing in new parameters to old vdo kernel drivers

bjohnsto:
Add device-mapper target version checking to decide when to add format in kernel parameters to the vdo table line.

### DIFF
--- a/src/perl/Permabit/BlockDevice/VDO.pm
+++ b/src/perl/Permabit/BlockDevice/VDO.pm
@@ -867,6 +867,22 @@ sub getTable {
 }
 
 ########################################################################
+# Get the target version of the VDO device mapper module.
+#
+# @return array of target version numbers, croaks if cannot be determined
+##
+sub getTargetVersion {
+  my ($self) = assertNumArgs(1, @_);
+  my $output = $self->runOnHost("sudo dmsetup target-version vdo");
+  chomp($output);
+  # Output format is typically: "vdo              v9.3.1"
+  if ($output =~ /v([\d.]+)/) {
+    return split(/\./, $1);
+  }
+  croak("Failed to extract target version from output: $output");
+}
+
+########################################################################
 # Disable the compression on a VDO device
 ##
 sub disableCompression {
--- a/src/perl/Permabit/BlockDevice/VDO/Unmanaged.pm
+++ b/src/perl/Permabit/BlockDevice/VDO/Unmanaged.pm
@@ -167,6 +167,14 @@ sub makeConfigString {
 
   my @optional = ();
 
+  # Don't add new parameters if we are using vdo modules that are older than 9.2.0.
+  my @targetVersion = $self->getTargetVersion();
+  if ($targetVersion[0] <= 8 || ($targetVersion[0] == 9 && $targetVersion[1] <= 1)) {
+    $extraArgs->{albireoMem} = -1;
+    $extraArgs->{albireoSparse} = -1;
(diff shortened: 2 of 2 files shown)


src/perl/Permabit/BlockDevice/VDO/Unmanaged.pm
```diff
@@ -183,31 +183,35 @@ sub makeConfigString {
...
 
-  my $memorySize
-    = defined($extraArgs->{albireoMem}) ? $extraArgs->{albireoMem} : $self->{memorySize};
-  if (defined($memorySize)) {
-    # magic value -1 suppresses the option completely
-    if ($memorySize != -1) {
-      push(@optional, ["indexMemory", "$memorySize"]);
+  my @targetVersion = $self->getTargetVersion();
+  # Check if version is >= 9.2.0 by comparing array elements
+  if ($targetVersion[0] > 9 || ($targetVersion[0] == 9 && $targetVersion[1] > 2)) {
```
lorelei-sakai: The comment seems correct, but does not match the code.
I would think that you want to check if it's strictly greater that 9.1, especially if you do change the version number for this branch.

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -866,6 +866,25 @@ sub getTable {
...
+########################################################################
+# Get the target version of the VDO device mapper module.
+#
+# @return array of target version numbers, undef if the version cannot be determined
+##
+sub getTargetVersion {
+  my ($self) = assertNumArgs(1, @_);
+  # Use "vdo" as the target name for dmsetup (not the kernel module name)
+  my $output = $self->runOnHost("sudo dmsetup target-version vdo");
+  chomp($output);
```
lorelei-sakai: Technically you don't need this chomp() here since you never use $output except for the matching expression, and it won't change result of the match.
Hm, well, I guess you use it in the "parse error" case, so maybe.

lorelei-sakai (changes requested):
In addition to the code comments, the commit message title is a little vague. (What is the problem you are fixing? It doesn't say.)
Also, are there ticket references that should be added here? I assume if you're fixing tests, that means we have a test failure somewhere you're trying to correct.

The general shape seems reasonable, though.

src/perl/Permabit/BlockDevice/VDO/Unmanaged.pm
```diff
@@ -167,6 +167,14 @@ sub makeConfigString {
 
   my @optional = ();
 
+  # Don't add new parameters if we are using vdo modules that
```
lorelei-sakai: This comment appears unfinished?

src/perl/Permabit/BlockDevice/VDO.pm
```diff
@@ -866,6 +866,22 @@ sub getTable {
   return $output;
 }
 
+########################################################################
+# Get the target version of the VDO device mapper module.
+#
+# @return array of target version numbers, empty array if cannot be determined
```
lorelei-sakai: You changed this, but it's not right again. It'll error out if it can't parse the string.

bjohnsto: fixed

lorelei-sakai (approved):
Yes, this all looks good to me. I hadn't noticed you'd pushed a new version of this, thanks for the comments.

PR #439, opened 2026-04-26
[VDO-6145] Change the gfp_ flag we use to call blkdev_issue_zeroout

bjohnsto:
There is an issue in the kernel around issuing zero writes. 

blkdev_issue_zeroout() often decides how to zero the range and routes down different code paths. In one of the paths, there is no check for bio_alloc returning null. When VDO passes in GFP_NOWAIT, that gfp_mask goes all the way down to bio_alloc on that path and we end up oops-ing because bio_alloc returns NULL. 

This change set basically switches the flag our call to blkdev_issue_zeroout uses from GFP_NOWAIT to using GFP_NOIO. GFP_NOWAIT is the wrong flag to use here. The usual flag for block level calls where we want to be able to be able to sleep is to use a flag like GFP_NOIO. Its much more likely not to fail just because the system is tight.


raeburn (approved):

lorelei-sakai (approved):
Looks good.

PR #440, opened 2026-04-30
[VDO-6150] Work on getting vdo to run dmtest from device-mapper-utils.

bjohnsto:
Replace the call to src/scripts/make-config with a direct createRemoteFile call in DMTest.pm, eliminating the dependency on the external script.


lorelei-sakai (approved):
Looks like that's doing the same thing as the script, and you've presumably tested that it works.

Looks good.

PR #441, opened 2026-05-06
[VDO-6150] : Update dmtest to run with device-mapper-utils repo

bjohnsto:
Updates the vdo-devel dmtest test framework to work with the device-mapper-utils repo.

Changes the default dmtest repo to point at device-mapper-utils. Also updates Grouped.pm to use the generated db of test results as it is more reliable. We ran into an issue testing where the print statements in dmtest weren't showing up because of flush timing issues.

### DIFF
--- a/src/perl/dmtest/DMTest/Grouped.pm
+++ b/src/perl/dmtest/DMTest/Grouped.pm
@@ -40,7 +40,6 @@ my $log = Log::Log4perl->get_logger(__PACKAGE__);
 #
 # The leaf nodes in the hash will have a value containing the result
 # and time the test took or '-' if it has not been run yet.
-# 
 #
 # @param tree  The tree returned from dmtest list.
 #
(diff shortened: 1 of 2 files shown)


src/perl/dmtest/DMTest/Grouped.pm
```diff
@@ -124,8 +124,12 @@ sub hashesToResult {
 sub testRunner {
   my ($self) = assertNumArgs(1, @_);
 
-  my $dmtestRun = $self->runTests($self->{dmtestName});
```
lorelei-sakai, line 127: You're removing this call the runTests() which look like the only call to runTests() in this tree. It probably makes sense to either update this function, or remove it altogether.

lorelei-sakai (changes requested):
Just one tiny cleanup nit. The general ideas here look great.

lorelei-sakai (approved):
Looks good.

PR #448, opened 2026-06-17
A small simplication in volume-index.c

corwin:
modifies the geometries, so there is no need to duplicate the enclosing configs. This makes the action of the function clearer, at the cost of requiring compute_volume_sub_index_parameters() and compute_volume_sub_index_save_bytes() to take the geometry as a separate argument rather than using the geometry in the supplied config.

### DIFF
--- a/src/c++/uds/src/uds/volume-index.c
+++ b/src/c++/uds/src/uds/volume-index.c
@@ -60,6 +60,12 @@
  * the index.
  */
 
+enum sub_index_parameters_slot {
+	DENSE = 0,
+	HOOK = 0,
+	NON_HOOK = 1,
+};
+
 struct sub_index_parameters {
 	/* The number of bits in address mask */
 	u8 address_bits;
@@ -77,16 +83,10 @@ struct sub_index_parameters {
 	size_t memory_size;
 	/* The number of bytes the index should keep free at all times */
 	size_t target_free_bytes;
-};
-
-struct split_config {
-	/* The hook subindex configuration */
-	struct uds_configuration hook_config;
-	struct index_geometry hook_geometry;
-
-	/* The non-hook subindex configuration */
-	struct uds_configuration non_hook_config;
-	struct index_geometry non_hook_geometry;
+	/* The mean delta for the volume index */
+	u32 volume_index_mean_delta;
+	/* The number of threads used to process index requests */
+	unsigned int zone_count;
 };
 
 struct chapter_range {
@@ -220,8 +220,11 @@ unsigned int uds_get_volume_index_zone(const struct volume_index *volume_index,
 
 #define DELTA_LIST_SIZE 256
 
-static int compute_volume_sub_index_parameters(const struct uds_configuration *config,
-					       struct sub_index_parameters *params)
+static int compute_sub_index_parameters(const struct uds_configuration *config,
+					u64 records_per_chapter,
+					u32 chapters_per_volume,
+					bool reduced,
+					struct sub_index_parameters *params)
 {
 	u64 entries_in_volume_index, address_span;
 	u32 chapters_in_volume_index, invalid_chapters;
@@ -231,21 +234,19 @@ static int compute_volume_sub_index_parameters(const struct uds_configuration *c
 	u64 index_size_in_bits;
 	size_t expected_index_size;
 	u64 min_delta_lists = MAX_ZONES * MAX_ZONES;
-	struct index_geometry *geometry = config->geometry;
-	u64 records_per_chapter = geometry->records_per_chapter;
 
 #ifdef TEST_INTERNAL
 	if (min_volume_index_delta_lists > 0)
 		min_delta_lists = min_volume_index_delta_lists;
 #endif /* TEST_INTERNAL */
-	params->chapter_count = geometry->chapters_per_volume;
+	params->chapter_count = chapters_per_volume;
 	/*
 	 * Make sure that the number of delta list records in the volume index does not change when
 	 * the volume is reduced by one chapter. This preserves the mapping from name to volume
 	 * index delta list.
 	 */
 	rounded_chapters = params->chapter_count;
-	if (uds_is_reduced_index_geometry(geometry))
+	if (reduced)
 		rounded_chapters += 1;
 	delta_list_records = records_per_chapter * rounded_chapters;
 	address_count = config->volume_index_mean_delta * DELTA_LIST_SIZE;
@@ -302,7 +303,44 @@ static int compute_volume_sub_index_parameters(const struct uds_configuration *c
 	params->memory_size = expected_index_size * 106 / 100;
 
 	params->target_free_bytes = expected_index_size / 20;
+	params->volume_index_mean_delta = config->volume_index_mean_delta;
+	params->zone_count = config->zone_count;
 	return UDS_SUCCESS;
+}
+
+static int compute_volume_sub_index_parameters(const struct uds_configuration *config,
+					       struct sub_index_parameters *params)
+{
+	struct index_geometry *geometry = config->geometry;
+	u64 sample_records;
+	u64 dense_chapters;
+	int result;
+	bool reduced = uds_is_reduced_index_geometry(geometry);
+
+	if (!uds_is_sparse_index_geometry(config->geometry)) {
+		return compute_sub_index_parameters(config,
+						    geometry->records_per_chapter,
+						    geometry->chapters_per_volume,
+						    reduced,
+						    &params[DENSE]);
+	}
+
+	dense_chapters = geometry->chapters_per_volume - geometry->sparse_chapters_per_volume;
+	sample_records = geometry->records_per_chapter / config->sparse_sample_rate;
+
+	result = compute_sub_index_parameters(config,
+					      sample_records,
+					      geometry->chapters_per_volume,
+					      reduced,
+					      &params[HOOK]);
+	if (result != UDS_SUCCESS)
+		return result;
+
+	return compute_sub_index_parameters(config,
+					    geometry->records_per_chapter - sample_records,
+					    dense_chapters,
+					    reduced,
+					    &params[NON_HOOK]);
 }
 
 static void uninitialize_volume_sub_index(struct volume_sub_index *sub_index)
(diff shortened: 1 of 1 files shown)


src/c++/uds/src/uds/volume-index.c
```diff
@@ -353,52 +349,49 @@ static int compute_volume_sub_index_save_bytes(const struct uds_configuration *c
 }
 
 /* This function is only useful if the configuration includes sparse chapters. */
-static void split_configuration(const struct uds_configuration *config,
-				struct split_config *split)
+static struct split_geometry split_geometry(const struct uds_configuration *config)
```
lorelei-sakai: I'm not sure why you changed the return value here from void to returning the whole struct. The split_geometry struct is smaller than the split_config struct, but it's still > 300 bytes. I think that's why we previously had the caller allocate one and passed a pointer to it to be initialized. So I'd prefer you change this back to using a struct pointer.

lorelei-sakai (changes requested):
Hi Corwin, great hearing from you.

I couldn't think of a good anchor for my biggest concern, so I'm just going to talk about it here. I'm very sympathetic to the idea here; the split_configuration is quite awkward. But I'm not really fond of passing both a uds_configuration and an index_geometry to functions, mostly because the config has a pointer to a geometry, and so maintainers will have to forever be careful to use the correct geometry, and not the potentially incorrect pointer that's right there. That's especially true in passthrough functions like initialize_volume_sub_index(), where it's not at all obvious that the two geometries might be different.

However, while looking at this, I noticed that most of the things you changed are just funneling the geometry to compute_volume_sub_index_parameters(), and that that function only uses one non-geometry value from the config (volume_index_mean_delta). So I think we could improve this whole thing by adding that one field to the index_geometry (4 bytes), and then most of these functions could omit the config and just pass the geometry, which seems much cleaner to me. (initialize_volume_sub_index() also wants the zone_count, which we could pass separately, or leave in the config.) It would require making sure we set the volume_index_mean_delta when we create the index_geometry, though. (I don't mind having that one field duplicated between the structures.)

-----
In addition to that issue, could you rewrite the commit message to have a clear one-line title/summary? I'll want one for upstream anyway, and current message made the PR description come out weird. Also, for upstream, do you want to be credited as author or otherwise?

src/c++/uds/src/uds/volume-index.c
corwin: Both call sites pass in a split config from a local variable, so it is getting allocated on the stack either way. As such, it seems more straight-forward to me to just return it.

corwin:
I am loath to change the geometry struct since it gets written out, does it not? Rather than adding the mean delta to the geometry, why not just pass that one value along with the geometry rather than passing the config?

corwin:
> I am loath to change the geometry struct since it gets written out, does it not? Rather than adding the mean delta to the geometry, why not just pass that one value along with the geometry rather than passing the config?

Oh, I see, the second call site also needs the zone_count, that's not the worst.

src/c++/uds/src/uds/volume-index.c
```diff
@@ -59,7 +59,12 @@
  * index sizes, this requires about 4 chapters of old entries per 1024 chapters of valid entries in
  * the index.
  */
-
```
lorelei-sakai, line 62: I'd prefer not to remove the blank line between the doc and the first declaration, please.

src/c++/uds/src/uds/volume-index.c
```diff
@@ -59,7 +59,12 @@
...
+enum sub_index_parameters_slot {
+	SUB_INDEX_PARAMETERS_DENSE,
```
lorelei-sakai: Are these constants really better in an enum than defined directly? You are counting on SUB_INDEX_PARAMETERS_DENSE == 0, maybe we should just say that.

Either way, these names make for some very long lines where they're used, and since they're localized to this file, it's probably worth making the names more succinct. DENSE/HOOK/NON_HOOK might be fine.

src/c++/uds/src/uds/volume-index.c
```diff
@@ -59,7 +59,12 @@
...
+	SUB_INDEX_PARAMETERS_HOOK = SUB_INDEX_PARAMETERS_DENSE,
+	SUB_INDEX_PARAMETERS_NON_HOOK = 1,
+};
+
```
lorelei-sakai: There's a tab in this line; please be careful that blank lines really are empty.

src/c++/uds/src/uds/volume-index.c
```diff
@@ -302,9 +302,46 @@ static int compute_volume_sub_index_parameters(const struct uds_configuration *c
...
+	if (result != UDS_SUCCESS)
+		return result;
+
+	return compute_sub_index_parameters(config,
+					    geometry->records_per_chapter - sample_records,
+					    dense_chapters,
+					    reduced,
+					    &params[SUB_INDEX_PARAMETERS_NON_HOOK]);
+}
+
```
lorelei-sakai: Extra tab on this line, too.

src/c++/uds/src/uds/volume-index.c
```diff
@@ -336,73 +373,29 @@ void uds_free_volume_index(struct volume_index *volume_index)
...
-	if (!uds_is_sparse_index_geometry(config->geometry))
-		return compute_volume_sub_index_save_bytes(config, bytes);
-
-	split_configuration(config, &split);
-	result = compute_volume_sub_index_save_bytes(&split.hook_config, &hook_bytes);
+	struct sub_index_parameters parameters[2] = {
+		{ .address_bits = 0 },
+		{ .address_bits = 0 },
+	};
+	int result = compute_volume_sub_index_parameters(config, parameters);
```
lorelei-sakai: I think this violate the "declarations before code" pattern.
Could you just put the "int result;" declaration back on its own line?

src/c++/uds/src/uds/volume-index.c
```diff
@@ -336,73 +373,29 @@ void uds_free_volume_index(struct volume_index *volume_index)
...
 	if (result != UDS_SUCCESS)
 		return result;
 
-	result = compute_volume_sub_index_save_bytes(&split.non_hook_config,
-						     &non_hook_bytes);
-	if (result != UDS_SUCCESS)
-		return result;
+	*bytes = compute_volume_sub_index_save_bytes(&parameters[0]);
```
lorelei-sakai: It is odd (here, and two lines down) to define constants for these slots and then use a base 0 and 1 here.
This function should probably use the labels, too. 

That might mean changing the structure back more like it was, so you can be precise about distinguishing the DENSE case from the HOOK case.

src/c++/uds/src/uds/volume-index.c
```diff
@@ -336,73 +373,29 @@ void uds_free_volume_index(struct volume_index *volume_index)
...
+	if (uds_is_sparse_index_geometry(config->geometry)) {
+		*bytes += compute_volume_sub_index_save_bytes(&parameters[1]) + sizeof(struct volume_index_data);
```
lorelei-sakai: This line is very long; could you wrap it or break it into two please?

src/c++/uds/src/uds/volume-index.c
```diff
@@ -1250,48 +1243,43 @@ void uds_get_volume_index_stats(const struct volume_index *volume_index,
...
-	if (result != UDS_SUCCESS)
-		return result;
-
-	sub_index->address_bits = params.address_bits;
-	sub_index->address_mask = (1u << params.address_bits) - 1;
-	sub_index->chapter_bits = params.chapter_bits;
-	sub_index->chapter_mask = (1u << params.chapter_bits) - 1;
-	sub_index->chapter_count = params.chapter_count;
-	sub_index->list_count = params.list_count;
+
```
lorelei-sakai: Another extra tab here.

src/c++/uds/src/uds/volume-index.c
```diff
@@ -1302,19 +1290,26 @@ static int initialize_volume_sub_index(const struct uds_configuration *config,
...
 
 	volume_index->zone_count = config->zone_count;
 
+	result = compute_volume_sub_index_parameters(config, parameters);
+	if (result != UDS_SUCCESS)
+		return result;
+	
 	if (!uds_is_sparse_index_geometry(config->geometry)) {
-		result = initialize_volume_sub_index(config, volume_nonce, 'm',
+		result = initialize_volume_sub_index(&parameters[SUB_INDEX_PARAMETERS_DENSE], volume_nonce, 'm',
```
lorelei-sakai: This line (and the two later ones that also use the parameter slot constants) is very long, please wrap them, or shorten the constant names. For reference I'd prefer they be no longer than 90-95 characters (counting the tabs as 8).

lorelei-sakai (changes requested):
I like this approach much, much better than the previous ideas in general, that's great. I just have a bunch of fussy comments about formatting.

Also, can you squash these two commits together? The previous one isn't really relevant anymore and it just makes it confusing to read.

src/c++/uds/src/uds/volume-index.c
lorelei-sakai: The new approach of adding the relevant fields to the sub_index_parameters is much better, so I'm going to consider all this older discussion resolved.

src/c++/uds/src/uds/volume-index.c
```diff
@@ -336,73 +374,32 @@ void uds_free_volume_index(struct volume_index *volume_index)
...
-	if (!uds_is_sparse_index_geometry(config->geometry))
-		return compute_volume_sub_index_save_bytes(config, bytes);
-
-	split_configuration(config, &split);
-	result = compute_volume_sub_index_save_bytes(&split.hook_config, &hook_bytes);
+	struct sub_index_parameters parameters[2] = {
+		{ .address_bits = 0 },
+		{ .address_bits = 0 },
+	};
+
```
lorelei-sakai: Extra tab on this line.

src/c++/uds/src/uds/volume-index.c
```diff
@@ -1326,6 +1325,7 @@ int uds_make_volume_index(const struct uds_configuration *config, u64 volume_non
 		return UDS_SUCCESS;
 	}
 
+
```
lorelei-sakai: Extra blank line introduced here.

lorelei-sakai (changes requested):
Just two whitespace irregularities.

I can fix those up later if you don't get to it.

lorelei-sakai (approved):
Everything looks good now. Thans for your contribution!

PR #447, opened 2026-06-16
[VDO-6175] Add wipe of signatures before resize.

bjohnsto:
Backport commit 4898a8a5 to 8.3 branch to fix regression issue caused by backporting activation fixes.

PR #450, opened 2026-06-18
[VDO-6179] backport thin pool and lvmdevices changes to 8.3 to fix regression issues

bjohnsto:
RHEL10 has thin pool with VDO support so we should be supporting it. Additionally, backporting these changes allows us to properly backport lvmdevices changes in proper chronological order to make sure there are no more regressions in the 8.3 branch. 

The following commits are being backported: 

 1. b4ff8efb : dm vdo test: split thin device into thin and thin pool
 2. 2a825d92 :  dm vdo test: add thin pool using vdo device
 3. b3872222 :  dm vdo test: use -an and -kn flags for creation
 4. 995d9436 : dm vdo test: fix createThinPool in VolumeGroup.pm
 5. 24bbe848 : dm vdo test: Move lvmdevices cleanup to VDOTest.pm
 6. 5021e7b9 : dm vdo test: Fix LVMDevicesFileTest.pm
 7. 8dd930b1 :dm vdo test: fix auto activation issue with VDO


src/perl/Permabit/VolumeGroup.pm
```diff
@@ -203,18 +196,9 @@ sub createThinPool {
 
   # Create a thin pool in an existing volume group. The pool is not
   # immediately enabled.
-  $machine->runSystemCmd("sudo lvcreate --name $name $dataType --type=thin-pool"
+  $machine->runSystemCmd("sudo lvcreate --name $name --type=thin-pool"
```
lorelei-sakai: It looks like you have a patch ordering problem in creating the series (I'm not quite sure how that happened). But it looks like patch 4 ("dm vdo test: use -an and -kn flags for creation") was merged with main before patch 3 ("dm vdo test: fix createThinPool in VolumeGroup.pm". In this series, you've switched them, so the $dataType argument added in patch 3 goes away in patch 4. (It looks like you patched this by later adding in back in patch 9, but it's better to get the patch order correct in the first place.)

lorelei-sakai (changes requested):
Most of the patches  look fine, but there a weird patch ordering issue with the series.
Patch 4 ("dm vdo test: use -an and -kn flags for creation") from PR #182 should come before patch 3 ("dm vdo test: fix createThinPool in VolumeGroup.pm") from PR #302.
This has the effect that the $dataType argument added in patch 3 gets removed in patch 4. This is especially bad since that argument is basically the whole point of patch 3.  (This issue gets fixed ultimately in patch 9 when the argument comes back, but it shouldn't be there.)

Also, I think patch 9 (from PR #393) should come before patch 8 (from PR #447) but those don't overlap so it doesn't actually cause any issues.

Can I get you to redo this series and apply the patches in the actual merge order?

lorelei-sakai (approved):
This series generally looks much cleaner than the previous series, and it matches the 7 patches in the description, too. (The description is slightly out-of-date since you've repushed the patches so the hashes of the new patches are all different now. For the record, It's good to cite the order patches in the description but I think citing the newer ones is not necessary; keeping the titles and order the same, as you did, is sufficient to match them up.)

There are some small oddities in VolumeGroup.pm.
First, deleteThinVolume() should really be removed in the first patch (as per the original) instead of the last patch, where you have it.
Second, the first patch adds disableAutoActivation() (which the original did not do) only for it to be removed again in the final patch. (That removal appears to have been merged to the 8.3 branch previously.)

But the final state looks correct and these details don't seem that important, so I'm happy to approve this.

PR #457, opened 2026-06-26
Replace deprecated strncpy with secure strscpy

C2Redhat:
Modern mainline kernel trees have removed the implicit declarations and usage of the legacy strncpy function to avoid buffer overflows and ensure safe null-termination.

Fix the compilation error on modern toolchains by migrating to strscpy, satisfying newer compiler constraints (-Wimplicit-function-declaration).


lorelei-sakai (approved):
Lioks good.

PR #460, opened 2026-07-15
[VDO-5990][VDO-5649] Guard against use-after-free in repair loop

lorelei-sakai:
In nightly, this invalid read only seem to show up in RHEL 10, so this will get need to merged there as well.

Respell the vio launch loop to use the existing vio_count value. Using repair->vio_count in the loop termination condition is not safe because it is possible for the repair completion to get freed after launching the final metadata vio but before the final loop termination check.


lorelei-sakai:
Ken points out that this race could be fixed in test code only. However there may be value in avoiding this potentially invalid access anyway.

raeburn (approved):

PR #462, opened 2026-07-16
Remove rebuild and readonly kernlog checks in dmtest

lorelei-sakai:
Now that there are rebuild tests in dmtest, we cannot consider rebuild messages in the kernel logs to be an error.

This should fix the dmtest runs in the dm-linux CI actions.

### DIFF
--- a/src/perl/dmtest/DMTest.pm
+++ b/src/perl/dmtest/DMTest.pm
@@ -580,8 +580,12 @@ sub uninstallModules {
     }
   }
 
+  my $userMachine = $self->getUserMachine();
+  # Some dm tests put vdo into readonly mode in order to rebuild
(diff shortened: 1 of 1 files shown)


raeburn (approved):
Arguably, dmtest should be checking for those messages in the specific tests that care, and not in those that don't... but certainly at this level we shouldn't be making such blanket assumptions about the whole suite.

lorelei-sakai:
> Arguably, dmtest should be checking for those messages in the specific tests that care, and not in those that don't... but certainly at this level we shouldn't be making such blanket assumptions about the whole suite.

Yes individual tests can and should check for these messages. The rebuild tests do check that the rebuild messages appear, but it might be worth adding a general vdo test cleanup that checks for the lack of these messages.

PR #463, opened 2026-07-20
Embed uds geometry in config and volume

corwin:
This change embeds the index_geometry directly in uds_configuration and volume, instead of those structs having pointers. This eliminates the need to allocate (and free) geometries separately.


lorelei-sakai (approved):
This all looks basically fine, thanks. There are few place you made lines a little longer than I like but it's not worth fussing about.

However, github decided to break our CI this weekend so I'm going to hold off merging this until we can get a proper CI run just to make sure all the tests still work.

lorelei-sakai (approved):

